PR #204, opened 2026-05-26
SF-213: remove extraneous Plugin.depends entries (per SF-208 audit)

sergio-bershadsky:
## Summary
Apply the audit's REMOVAL findings only — declaring fewer deps is always safe.

- Drop `url4-specs` from four frontend plugins: `claude_frontend`, `codex_frontend`, `gemini_frontend`, `ollama_frontend` (declared but never imported).
- Drop `claude-frontend` from `claude_env_intercept` (declared but never imported; replace with empty `depends: list[str] = []` for type-annotation consistency).
- Update two metadata tests (`codex_frontend`, `gemini_frontend`) that asserted the dropped entry.

## Why removals only

Adding new declared deps **tightens** runtime activation: many tests instantiate plugins via narrow `plugins=[X]` fixtures and the registry refuses to activate `X` without its newly-declared transitive deps. Trial-running the full additions table broke ~25 tests across the `url4-executor → eval-runs → state` chain. The 12 additions are deferred to a follow-up ticket that scopes the fixture-widening alongside.

## Audit residual after this PR (regenerated report)

| Metric | Before SF-208 → After SF-208 → After SF-213 |
|---|---|
| Prod-missing | (n/a) → 12 → 12 *(deferred)* |
| Extraneous | (n/a) → 7 → 2 *(both are known audit false positives — see below)* |
| Prod cycles | (n/a) → 2 → 2 *(architectural, deferred)* |

The remaining 2 extraneous are `backend-api-base` on `aigw-codex-backend` and `aigw-gemini-backend`. The audit follows inheritance for the `depends` declaration but not for the corresponding imports, so the dep is reported as extraneous even though the inherited base class legitimately uses it. Audit-script refinement deferred.

## Test plan
- [x] `uv run pytest -m \"not e2e and not e2e_live\" -q` → 1084 passed, 17 skipped
- [x] `ruff format` / `ruff check` clean on `src/screamingface/plugins`
- [x] Audit re-run produces the residual counts above

## Dependencies
Depends on PR #202 (SF-208) for the audit script that verifies this PR. Merge #202 first, then this.

## Out of scope (follow-up tickets)
- The 12 prod-missing additions (test-fixture surgery required).
- The 2 real prod cycles (`aigw-base ↔ url4-executor` via `backend-api-base`; `aigw-base ↔ llm-base`) — architectural call.
- Asymmetric inheritance handling in the audit script.

Asana: https://app.asana.com/1/1185126988600652/project/1213628819033917/task/1215115079155250
Plan: docs/superpowers/plans/SF-213-fix-plugin-depends-manifests.md


sergio-bershadsky:
Superseded by #206 — rebuilt cleanly on top of fresh main after #202 was squash-merged. Same contents, single commit.

PR #226, opened 2026-05-29
SF-232: AIGateway reactive overload backpressure (retry 429/529/503)

sergio-bershadsky:
## Summary
Makes AIGateway **absorb transient upstream overload** (HTTP 429/529/503) by retrying the chat dispatch with `Retry-After`-aware backoff before forwarding the error. Today the gateway forwards overload responses immediately, so transient provider rate-limits surface as hard url4 failures.

- New stdlib-only helper `core/retry.py`: `RetryPolicy`, `is_retryable_status`, `parse_retry_after_seconds`, `with_overload_retry` (duck-typed on `status_code`).
- 5 tunable `AIGW_RETRY_*` settings (attempts, backoff base/max, total-wait budget, jitter).
- `routes/chat.py` wraps the single non-streaming dispatch; `_retry_after_headers` now reuses the shared parser (DRY).
- On exhaustion the **original** exception is re-raised, so existing error mapping/profile-error marking is unchanged. `AIGW_RETRY_MAX_ATTEMPTS=0` restores today's immediate-error behavior (kill-switch).
- Streaming is out of scope (a 200 commits before dispatch); marked as a deliberate follow-up.

Spec: `docs/superpowers/specs/2026-05-29-aigateway-429-backpressure-design.md`
Plan: `docs/superpowers/plans/2026-05-29-aigateway-429-backpressure.md`
Asana: SF-232 — https://app.asana.com/1/1185126988600652/project/1213628819033917/task/1215231263845954

> **Stacked PR:** based on `SF-231-url4-resolve-locus-and-secret` (per request). Will auto-retarget to `main` once that merges. Review only the aigateway commits here.

## Test plan
- [x] `test_retry.py` — helper in isolation (retryable statuses, Retry-After honored, exponential backoff + cap, budget cut-off, kill-switch, malformed header, non-retryable passthrough). 20 passed.
- [x] `test_config_retry.py` — defaults + `AIGW_RETRY_*` env overrides.
- [x] `test_chat_x_profile.py` — 429-then-200 and 503-then-200 absorbed; persistent 429 still returns 429 + `Retry-After` after `max_retries+1` attempts; existing rate-limit regression stays green.
- [x] Full aigateway unit suite: 371 passed, 24 skipped.
- [x] `ruff check .` + `ruff format --check .` clean.
[...]

PR #228, opened 2026-05-30
SF-233: AIGW per-provider concurrency guardrail + Gemini reset hint

sergio-bershadsky:
## Summary
Follow-up to SF-232 that finally lets the url4 leaderboard fan-out survive provider rate limits. SF-232's reactive retry alone couldn't absorb this query because (A) Gemini puts its reset window in the 429 body — not a `Retry-After` header — and the plugin discarded the response when mapping `CustomLLMError` → `HTTPException`, so retry was backing off blindly (0.5/1/2s) while the quota actually needed ~3s; and (B) the collection iteration fires every JSONL row × 3 providers with no concurrency cap, so each quota reset was immediately re-exhausted by concurrent siblings (sustained pressure that reactive retry deliberately doesn't fix).

### A — honor the *actual* reset window
- Generalize `core/retry.parse_retry_after_seconds` to read `exc.headers` (FastAPI `HTTPException`) in addition to `exc.response.headers` (LiteLLM exceptions).
- Gemini chat handler parses `retryDelay` / "reset after Ns" out of the 429 body and stashes it on the `CustomLLMError`; the plugin promotes it to a `Retry-After` header on the raised `HTTPException`.

### B — stop self-DoSing the provider
- New `AIGW_PROVIDER_MAX_CONCURRENCY` (default 4, `0` = unlimited).
- New `core/concurrency.provider_slot(app, provider, limit)` async CM backed by an `asyncio.Semaphore` registry on `app.state`, **rebuilt if the configured limit changes**.
- `chat.py` extracts `_dispatch_with_backpressure(...)` — the slot wraps `with_overload_retry`, so a provider's concurrent-pressure ceiling includes backoff time.

Spec/plan: `docs/superpowers/plans/2026-05-29-aigw-concurrency-guardrail-gemini-hint.md`
Asana: SF-233 — https://app.asana.com/1/1185126988600652/project/1213628819033917/task/1215264490396472

> **Stacked PR:** based on `SF-232-aigateway-overload-backpressure` (PR #226). Will auto-retarget down the stack as bases merge.

## Test plan
- [x] `test_retry.py` — `.headers` path + precedence vs `.response.headers` (3 new).
- [x] `test_chat_handler.py` (gemini) — body-hint parser: `retryDelay`, "reset after Ns", header-wins, none-when-absent (4 new).
- [x] `test_config_retry.py` — `provider_max_concurrency` default + env override (2 new).
- [x] `test_concurrency.py` — `limit=0` no-op; `limit=2` caps observed peak at 2; per-provider keying isolates `claude` from a saturated `gemini` slot; limit change rebuilds the semaphore (4 new).
- [x] Full aigateway unit suite: **384 passed, 24 skipped**.
- [x] `ruff check .` + `ruff format --check .` + `pyright` clean across the whole package.
- [ ] Manual smoke: re-run the leaderboard fan-out (`livetruth-latest.eval.jsonl` × 3 providers) with the gateway restarted on this branch — Gemini's "reset after Ns" surfaces as `Retry-After`, the semaphore caps fan-out at 4 per provider, and the query completes instead of returning 429.


apps/server/src/screamingface/plugins/aigw_runner/plugin.py
```diff
@@ -346,6 +356,10 @@ def _gateway_env(settings: AigwRunnerSettings, database_url: str) -> dict[str, s
     env.pop("VIRTUAL_ENV", None)
     env["AIGATEWAY_DATABASE_URL"] = database_url
     env["AIGATEWAY_AUTH_ENABLED"] = "false"
+    # Inject configured AIGW_* / other env so it survives launchers that
+    # strip the user's shell env (notably the Electron desktop app).
+    for key, value in settings.aigw_env.items():
+        env[key] = str(value)
     return env
```
HupBaHa, line 365: Small boundary fix: as written, `aigw_env` can override the runner-owned local-managed invariants (`AIGATEWAY_AUTH_ENABLED=false` and the `database_path`-derived `AIGATEWAY_DATABASE_URL`). Applying config env first and forcing those two keys last keeps the new tuning channel while preserving local-managed mode.

```suggestion
    # Inject configured AIGW_* / other env so it survives launchers that
    # strip the user's shell env (notably the Electron desktop app).
    for key, value in settings.aigw_env.items():
        env[key] = str(value)
    # Runner-owned invariants for local-managed mode must win over aigw_env.
    env["AIGATEWAY_DATABASE_URL"] = database_url
    env["AIGATEWAY_AUTH_ENABLED"] = "false"
```

PR #230, opened 2026-06-02
SF-34: URL4 scored-ensemble query (v1-lite) — executor engine + python-runner call-main

sergio-bershadsky:
## Summary

Implements the **scored-ensemble URL4 query** end-to-end: over a dataset, each row gets a model **consensus**, a python script **scores** it against the expected answer, and a collection-level python script computes **accuracy** — landing per-row verdicts in the Eval Studio pipeline.

Shipped spec (`apps/server/sf.json` → `url4-specs.specs.ScoredLiveTruth`), v1-lite single-model consensus:

```
( DATASET *(
    consensus=/claude($item.question)!'…short answer…',
    /python(/data/code/check_correct.py)!{"question":"$item.question","expected":"$item.expected_answer","correct_answer":"$item.expected_answer","consensus":"$consensus"}
  ) ;foreach.concurrency=10;foreach.on_error=collect
)!/python(/data/code/calculate_accuracy.py)
```

**Verified end-to-end offline** (`tests/e2e/test_scored_query.py`): mocked `/claude` + dataset, **real** sandboxed python scripts → `accuracy_pct == 66.67` (2/3), with 3 `HOOK_QUESTION_CHECKED` fires (correct flags `[False, True, True]`).

## What's in here (per milestone)

| | Capability |
|---|---|
| M1 | `$name` substitution from the `Env` scope chain, wired into backend-call + text intents (Phase 1.4) |
| M2 | `;foreach.*` execution-annotation parsing (Phase 1.2) |
| M3 | bounded concurrency + `on_error=collect` in collection iteration (Phase 3.1/2.6) |
| M4 | collection-level reducer over the per-row result array (Phase 3.1/3.4) |
| M5.0 | `json_blob` grammar rule so multi-key JSON payloads parse |
| M5.1 | per-row body resolution + **let-semantics** for mixed `(binding, body)` lists |
| M6 | python-runner **call-`main()`** mode — runs the eval scripts **byte-for-byte unchanged** (imported as modules, `main()` invoked with introspected kwargs) |
| M7.1 | install scripts + `ScoredLiveTruth` spec in `sf.json` |
| M7.2 | end-to-end test |

**Tests:** 310 green (url4_executor + python_runner + e2e). Each task: TDD, with spec + code-quality review.

## ⚠️ Reviewer notes

[...]

PR #231, opened 2026-06-02
SF-34 fast-follow: quoted-intent grammar rule (commas in prompts; fixes MainOne)

sergio-bershadsky:
## Summary
Adds a `quoted` rule to the URL4 intent atom so a backend-call intent can be a quoted string containing **commas** — e.g. `!'Answer this. Return only the answer, no other text.'`. Today such intents truncate at the comma (the intent atom only had `text = /[^,()]+/`).

Twin of the existing `json_blob` rule (which did the same for `{…}` payloads).

## Why it matters beyond the scored query
**This also fixes MainOne/MainTwo**, whose stored prompts contain commas inside quotes (`Format: {"A": 0.7, "B": 0.2, …}. Return only the JSON object.`) and therefore mis-parse today (the fanout list splits on the prompt's internal commas).

## Change (additive — only triggers for `'…'` / `"…"` intents)
- `atom_no_bc`: add `| quoted` before `| text`
- `quoted = value:/'[^']*'|"[^"]*"/ ;`
- semantic action returns `Url4Text(value=…strip())` — identical shape to how a comma-free quoted intent parses today (downstream `resolve_intent` still strips the quotes), so behavior is unchanged except commas are now retained.

## Tests
5 new (`test_quoted_intent.py`), incl. a MainOne-shaped fanout with comma-laden quoted intents. Full url4_executor suite: **252 green, no regressions**.

## Stacking
Stacked on #230 (base = `SF-34-url4-scored-ensemble-query`). The change is self-contained and could also be cherry-picked straight to `main` (it fixes MainOne independently of the scored query).

### DIFF
--- a/apps/server/src/screamingface/plugins/url4_executor/tests/test_quoted_intent.py
+++ b/apps/server/src/screamingface/plugins/url4_executor/tests/test_quoted_intent.py
@@ -1,0 +1,35 @@
+# pyright: reportAttributeAccessIssue=false, reportOptionalMemberAccess=false
+from screamingface.plugins.url4_executor.url4_ast import Url4BackendCall, Url4List, Url4Text
+from screamingface.plugins.url4_executor.url4_grammar import parse
+
+
+def test_quoted_intent_with_comma_parses_whole():
+    node = parse("/claude(q)!'Answer this. Return only the answer, no other text.'")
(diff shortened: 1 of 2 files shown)


sergio-bershadsky:
Superseded by #235 — rebased onto current `main`, stacked on #234.

PR #244, opened 2026-06-04
fix(frontend): blocking + fail-loud url4 resolve — scream on resolve failure (SF-237)

sergio-bershadsky:
## SF-237 — blocking + fail-loud url4 resolve

> Reworked from the original non-blocking approach per discussion — see commits e3d9163 (plan) → a0c903e (rework) → c8b7382 (tests + docs).

### Direction
The `*_frontend` url4 spec resolution now **blocks** on `/ensemble` and **screams** on failure: a timeout / 502 / resolve error surfaces as a **visible `[url4 error]` message in the CLI** (`_build_error_response`), instead of silently proceeding without context. This replaces the earlier non-blocking + background-warm + **silent** approach (which made failures invisible — the opposite of what a demo wants).

### What it does
- **Block + scream:** `resolve_context` re-raises on failure; every frontend's proxy (claude/codex/gemini/ollama) already wraps the call and returns a provider-shaped 200 whose body is the url4 error + traceback → visible in the CLI.
- **Negative cache (kept from #244):** a failed spec fails-fast — a repeat within `resolve_failure_ttl` re-raises the cached error immediately (no ensemble re-run).
- **Bounded, loop-aware fetch (kept from #244):** `_fetch_sync` is leak-free; runs the fetch on a throwaway thread when called from the async handler (the production path), `asyncio.run` fallback otherwise.
- **Timeouts unified:** `resolve_timeout` default **300 → 1200**; `_fetch` honors it (was hardcoded 300); the `$prompt` path uses `resolve_timeout` for both in-process (`asyncio.wait_for`) and HTTP (`httpx.Timeout`) branches.
- **Removed** the background-warm machinery (`get_cached_context`/`_maybe_warm`/`_warm`/`_RESOLVE_POOL`).

### Tests
- `105 passed` across all five frontend dirs; server pyright 0 errors; pre-commit (ruff 0.9.0 + pyright) clean.
- New `test_resolve_screams.py` pins the fail-loud contract for codex/gemini/ollama (verified: removing a frontend's `try/except` turns the test red).

### Notes / follow-ups
- **Blocking is intentional** (single-client, per-session proxy model): a stuck `/ensemble` monopolizes that session's worker for up to `resolve_timeout` (1200s) — documented in code.
- **Degraded-200** (an `on_error=collect` 200 that's actually an error report) is best-effort on the `$prompt` HTTP branch via the `X-SF-Collected-Errors` header; full handling on the static path lands with #243 (which adds that header).

Plan: `docs/superpowers/plans/2026-06-04-frontend-resolve-blocking-screaming.md`

### DIFF
--- a/apps/server/src/screamingface/plugins/claude_frontend/_url4_context.py
+++ b/apps/server/src/screamingface/plugins/claude_frontend/_url4_context.py
@@ -23,6 +23,7 @@ Extracted from ``proxy.py`` during SF-107.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import traceback
 from typing import Any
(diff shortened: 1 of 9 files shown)


sergio-bershadsky:
Closing: the fix (blocking + fail-loud url4 resolve, SF-237) is already on main — see claude_frontend/_url4_context.py (raises on resolve failure; docstring cites this PR's design). This branch is stale/conflicting.

PR #245, opened 2026-06-04
fix(aigw): serialize claude fan-out via anthropic concurrency cap + live test (SF-239)

sergio-bershadsky:
### Problem
Under high url4 fan-out concurrency (the ScoredLiveTruth ensemble runs `;foreach.concurrency=10`), claude trips an **implicit rate limit** that surfaces as a masked `400 invalid_request_error: "Your credit balance is too low"`. Confirmed via live e2e: simple and small (2–3 way) `/claude` fan-out all bill correctly (subscription routing + OAuth are fine) — the failure only appears at high concurrency. This is the *thundering-herd* failure mode the aigateway's per-provider semaphore (`core/concurrency.py`) was built to absorb; gemini already mitigates it with `{"gemini-cli":1}`.

### Fix
Mirror gemini for claude — add `anthropic` to the per-provider concurrency override so the gateway serializes claude's upstream calls during fan-out (model `anthropic/claude-sonnet-4-5` → provider `anthropic`):

```json
"AIGW_PROVIDER_MAX_CONCURRENCY_OVERRIDES": "{\"gemini-cli\":1,\"anthropic\":1}"
```

`_dispatch_with_backpressure` already holds `provider_slot(provider, effective_provider_limit(...))` across the retry loop, so with cap=1 at most one claude call hits the upstream at a time — no concurrent herd to trip the limit.

### Live test (`apps/aigateway/tests/live/test_anthropic_live.py`)
- `test_anthropic_fanout_serialized_under_cap_1` — with `anthropic:1`, fires 6 concurrent claude requests; asserts **all 200** (no 429/credit error) and **peak in-flight upstream == 1** (serialized). The tracker wraps `litellm.acompletion`, the real call made *inside* `provider_slot`.
- `test_anthropic_fanout_overlaps_without_cap` — control: with the cap disabled (`anthropic:0`), the same fan-out overlaps (peak > 1), proving the tracker observes real concurrency so the cap=1 assertion isn't a harness artifact.
- Both `@pytest.mark.live` + `AIGW_LIVE=1`-gated; skip cleanly otherwise. Verified: collect + skip clean, ruff + pyright clean.

### Apply / run
- **Apply:** restart the desktop — the gateway reads `AIGW_PROVIDER_MAX_CONCURRENCY_OVERRIDES` at startup (per `apps/server/sf.json`).
- **Run the live test:** `cd apps/aigateway && AIGW_LIVE=1 AIGATEWAY_ADMIN_PASSWORD=… uv run pytest tests/live/test_anthropic_live.py -k fanout -m live -v` (needs an authenticated `anthropic:default` profile, like the other anthropic live tests).
- **Trade-off:** `1` = fully sequential (matches gemini; safest). Bump to `"anthropic":2` if claude tolerates 2 concurrent for speed. `codex` is still at the default cap (4).

### DIFF
--- a/apps/aigateway/tests/live/test_anthropic_live.py
+++ b/apps/aigateway/tests/live/test_anthropic_live.py
@@ -11,8 +11,10 @@ import json
 import os
 import subprocess
 import sys
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
+import litellm
 import pytest
 from fastapi.testclient import TestClient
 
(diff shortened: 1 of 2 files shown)


sergio-bershadsky:
Closing: the SF-239 fix (per-provider concurrency cap) is already on main — aigateway/core/concurrency.py + config.py provider_max_concurrency + sf.json AIGW_PROVIDER_MAX_CONCURRENCY. This PR only added a live regression test for an already-shipped fix; closing as stale.

PR #246, opened 2026-06-04
fix(llm-base): exclude non-credentialed backends from /backends/status sweep (SF-238)

sergio-bershadsky:
## SF-238 / WI-4 — exclude non-credentialed backends from /backends/status (demo silent-failure #1)

Part of the **demo-path silent-failures** cluster. Plan: `docs/superpowers/plans/2026-06-04-demo-path-silent-failures.md`.

### Problem
The desktop backend-status panel (`GET /backends/status`) treats **every** active plugin with `backend_call_paths` as a credentialed LLM backend. The utility `python-runner` declares `backend_call_paths=["/python"]` (so url4 can route `/python(...)` script calls to it) but is not an LLM provider — no `/python/health`, no `gateway_provider`, no `cli_auth_command`. So the sweep probes `GET /python/health` → 404 → classifies it `reauth` → shows the misleading **"Credential is missing or expired."** for `/python`.

### Fix
One guard in `_collect_backend_status`: skip any plugin lacking both `gateway_provider` and `cli_auth_command`. That's the exact discriminator the rest of the file already keys off (`_classify_auth_kind`, `_cli_command`, `_provider_auth_status`) — real LLM backends always have one (all `aigw-*` via `gateway_provider`, CLI backends via `cli_auth_command`).

### Tests
- `test_backends_status_v2.py`: new `test_collect_backend_status_excludes_non_credentialed_backends` — a `python-runner`-style plugin is absent from the sweep while gateway-provider and cli-auth plugins remain present.
- Regression: `113 passed` across `llm_base/`; server-venv pyright `0 errors`; pre-commit clean. No existing test asserted utility backends appear, so none needed changing.

### DIFF
--- a/apps/server/src/screamingface/plugins/llm_base/routes.py
+++ b/apps/server/src/screamingface/plugins/llm_base/routes.py
@@ -74,6 +74,14 @@ async def _collect_backend_status(app: Any = None) -> dict[str, Any]:
     for plugin in plugins.values():
         if not plugin.backend_call_paths:
             continue
+        # Only credentialed LLM backends belong in the auth/status sweep. A
+        # utility backend (e.g. python-runner) declares backend_call_paths so
+        # url4 can route to it but has no gateway provider or CLI auth command,
+        # so probing it for credentials is meaningless (404 -> false "reauth").
+        if not (
+            getattr(plugin, "gateway_provider", None) or getattr(plugin, "cli_auth_command", None)
+        ):
+            continue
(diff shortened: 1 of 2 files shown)


sergio-bershadsky:
Closing: redundant — excluding non-credentialed backends from /backends/status is already on main via the requires_auth flag (llm_base/routes.py: 'if not getattr(plugin, "requires_auth", True): continue', SF-246).

PR #257, opened 2026-06-08
SF-243: claude CLI url4 gateway launcher banner

sergio-bershadsky:
## SF-243 — claude CLI launcher banner (url4 ensemble gateway)

When a user launches `claude` redirected to the gateway, they immediately see (on **stderr**) that they're on the ScreamingFace url4 ensemble gateway and **which active url4 spec** answers their queries — or a warning if none is set.

### Mechanism
`claude_env_intercept` already owns a marker-delimited block in the shell rc files (holding `export ANTHROPIC_BASE_URL=…`). This adds a `claude()` shell function to that same block:
- prints the banner to **stderr**, then runs the real binary via `command claude "$@"` — no recursion, claude's stdout/protocol untouched;
- the banner is **rendered at `setup()`** from the live plugin (`active_spec` + `get_active_expression()`) and **baked into the function with `shlex.quote()`** (one `printf '%s\n' '<quoted>' >&2` per line), so arbitrary url4 expressions (`$`, quotes, `(`, `)`, `!`, backticks) can never break out of the rc or be re-evaluated by the shell;
- removed automatically on teardown — `remove_exports()` strips the whole block (all four teardown paths).

### Why baked, not dynamic
A dynamic `sf banner` reading `sf.json` at launch was rejected in adversarial review: interactive shells have `SF_CONFIG` unset and `load_config()` reads `./sf.json` relative to the user's cwd → resolves to *no spec* in almost every real launch. Baking at setup avoids it. Tradeoff: the banner reflects the spec as of intercept activation (same lifecycle as the `ANTHROPIC_BASE_URL` export); re-activate to refresh.

### Safety / tests
The generated function is executed against a hostile expression in **sh, bash, and zsh** (review additionally ran a command-injection probe — no execution, no variable expansion, exit 0). Teardown removal, the no-spec warning, and `add_exports` back-compat are all tested. 25 plugin tests pass; pyright clean; pre-commit green. `shellenv`/`add_exports` have no callers outside this plugin.

Plan: `docs/superpowers/plans/2026-06-08-SF-243-gateway-banner.md`.

### DIFF
--- a/apps/server/src/screamingface/plugins/claude_env_intercept/plugin.py
+++ b/apps/server/src/screamingface/plugins/claude_env_intercept/plugin.py
@@ -13,7 +13,12 @@ import sys
 from typing import TYPE_CHECKING
 
 from screamingface.plugin import Plugin
-from screamingface.plugins.claude_env_intercept.shellenv import add_exports, remove_exports
+from screamingface.plugins.claude_env_intercept.shellenv import (
+    add_exports,
+    build_claude_banner_function,
+    remove_exports,
+    render_gateway_banner,
+)
 
 if TYPE_CHECKING:
     import typer
@@ -72,8 +77,18 @@ class ClaudeEnvInterceptPlugin(Plugin):
         # Build env vars
         env_vars: dict[str, str] = {"ANTHROPIC_BASE_URL": base_url}
 
-        # Write to shell profile
-        add_exports(env_vars)
+        # Build the `claude` launcher banner from the active spec, baked into the
+        # managed block so it shows at every `claude` launch (stderr only). Rendered
+        # now (not at launch) — dynamic config lookup is unreliable from an arbitrary
+        # shell cwd.
+        banner = render_gateway_banner(
+            cf_settings.active_spec,
+            cf_plugin.get_active_expression(),
+        )
+        banner_fn = build_claude_banner_function(banner)
+
+        # Write to shell profile: exports + the `claude` wrapper function.
+        add_exports(env_vars, extra_lines=[banner_fn])
         logger.info("Claude Code env configured: ANTHROPIC_BASE_URL=%s", base_url)
 
         # Also set via launchctl for current session (macOS)
--- a/apps/server/src/screamingface/plugins/claude_env_intercept/shellenv.py
+++ b/apps/server/src/screamingface/plugins/claude_env_intercept/shellenv.py
@@ -135,7 +135,7 @@ def build_claude_banner_function(banner_text: str) -> str:
     body = "\n".join(
         f"  printf '%s\\n' {shlex.quote(line)} >&2" for line in banner_text.split("\n")
     )
-    return 'claude() {\n' + body + '\n  command claude "$@"\n}'
+    return "claude() {\n" + body + '\n  command claude "$@"\n}'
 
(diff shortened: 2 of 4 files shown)


sergio-bershadsky:
## Review note: reconsider *where* the banner lives — move it to `claude_frontend`

The banner itself is great; the concern is the **layer**. As implemented here it's a `claude()` shell-rc wrapper owned by `claude_env_intercept`, which couples it to one intercept strategy:

- There are three mutually-exclusive intercept strategies (`docs/architecture/plugin-dependencies.md`): `claude-intercept` (hosts + sudo), `claude-env-intercept` (shell-rc `ANTHROPIC_BASE_URL`), `mitmproxy-intercept`. Only the shell-rc one can wrap the CLI and print at launch.
- **None of the three is enabled in the shipped `apps/server/sf.json`**, whereas `claude-frontend` is. So this banner is dormant in practice, and even when active only fires for 1 of 3 strategies.

### Proposal
Surface the banner from **`claude_frontend`** — the always-enabled plugin that produces every `/v1/messages` response — so it shows regardless of intercept strategy. Accepted tradeoff: it's no longer a *launch* banner; it renders at the top of the **first prompt's** answer. It must never leak into the ensemble input — and the request path makes that clean.

**Three hook points (all in `claude_frontend`):**
1. **Render** — relocate `render_gateway_banner` into `frontend_base` (decoupled from this plugin), wrapped in a tight sentinel `⟦SFGW⟧…⟦/SFGW⟧`.
2. **Inject** — `proxy.py` Stage 4, *first turn only*, into the returned/streamed bytes **after** `session.save` persists the clean dict (so the session store stays banner-free). Keep `response_id` derived from the clean `result_text` so `deterministic_id` cache keys don't shift.
3. **Strip** — sentinel removal in `_extract_turns` / `serialize_transcript`, defending the client-echo path on turn 2+.

### Why one strip point is sufficient (traced)
In `proxy_messages` the order is:
```
body = await request.json()              # client history (echoes turn-1 assistant on turn 2+)
body = await session.enrich(body, ...)   # enrich REPLACES body with merged history (_session.py:94)
prompt_blob = serialize_transcript(_extract_turns(body))   # ensemble input — built ONLY from body["messages"]
```
`_extract_turns` reads exactly `body["messages"]`, and `enrich` has already merged any server-side prior turns into that same list. So a content-strip there covers **both** the client echo and the enrich path. On turn 1 the ensemble input is clean by construction (no assistant turn yet). `prompt_blob` is the sole value handed to `$prompt` — there is no second route to the ensemble.

### Also
Retire the `build_claude_banner_function` / `extra_lines` plumbing and revert `add_exports` to its single-arg form, so `claude_env_intercept` goes back to "writes `ANTHROPIC_BASE_URL` only."

Full write-up with critical files + a 7-step verification plan (first-turn inject + clean ensemble, no-leak-on-turn-2, the enrich path specifically, clean saved copy, no-spec warning, streaming parity, url4 e2e regression): `docs/superpowers/plans/2026-06-09-SF-243-gateway-banner-revision.md`.

One open item to confirm: that no downstream consumer (cache/log dedup) expects the message id to match the literally-returned bytes — the plan keeps the id keyed on the clean `result_text`.

PR #363, opened 2026-07-07
feat(desktop): add main-process leaderboard-fetch service

FilipBolt:
## Summary
- Adds `listLeaderboard()`, a main-process function that fetches one benchmark's ranked leaderboard from the public scoreboard (`GET /v1/leaderboard/{id}`), modeled on the existing `list-benchmarks.ts`: main-process fetch (exempt from renderer CORS), 8s timeout, defensive row mapping, `null` on any failure.
- Deliberately unwired — no IPC channel, no preload exposure, no UI. This is milestone 1 of 6 in a stacked sequence for OME-321; each subsequent PR builds on this one.

## Stack
1. **→ this PR** — main-process fetch service
2. `ome-321-leaderboard-ipc-preload` — IPC channel + preload exposure
3. `ome-321-leaderboard-hook` — renderer hook
4. `ome-321-leaderboard-table` — presentational table component
5. `ome-321-leaderboard-view` — the view wiring it together
6. `ome-321-leaderboard-nav` — sidebar/nav wiring + docs

## Test plan
- [x] `npx vitest run` — 8 new unit tests, all passing (happy path, `?top=` param, malformed-entry dropping, malformed benchmark, 404, network failure, 8s timeout, config-driven URL)
- [x] `npm run lint` — 0 errors
- [x] `npm run build` — succeeds
- [x] Manually verified `listLeaderboard()` against a real local `apps/scoreboard` instance (seeded `livetruth` benchmark) — output matched `curl`'s response exactly; confirmed `null` (not throw) on a real 404 and on an unreachable host

Linear: [OME-321](https://linear.app/openmined/issue/OME-321/sf-27-leaderboard-pull-the-public-board-into-desktop-sdk)

### DIFF
--- a/apps/desktop/src/main/services/__tests__/fetch-leaderboard.test.ts
+++ b/apps/desktop/src/main/services/__tests__/fetch-leaderboard.test.ts
@@ -1,0 +1,148 @@
+import { vi, describe, it, expect, afterEach } from 'vitest';
+
+// Mirror list-benchmarks.test: stub context (URLs) + publish-log (pulls in electron).
+const CTX = {
+  scoreboardUrl: 'https://scoreboard.screamingface.ai',
+  portalUrl: 'https://screamingface.ai/portal/',
+  client: { name: 'screamingface-desktop', version: '0.4.2', platform: 'darwin' },
+};
+vi.mock('../publish-context', () => ({ resolvePublishContext: () => CTX }));
+vi.mock('../publish-log', () => ({ publishLog: () => {} }));
+
+import { listLeaderboard } from '../fetch-leaderboard';
+
+const BENCHMARK = {
+  id: 'livetruth',
+  display_name: 'LiveTruth',
+  description: 'Sample benchmark for local dev',
+  dataset_url: null,
+};
+
+const ENTRY = {
+  rank: 1,
+  spec_id: 'local-smoke',
+  accuracy: 0.5,
+  total_questions: 2,
+  ran_with_providers: ['smoke'],
+  submitted_at: '2026-07-07T15:30:05.108456Z',
+  submitted_by: 'filip-local',
+  verified_by_openmined: false,
+  url4_expression: 'url4://smoke',
+};
+
+function okResponse(body: unknown): Response {
+  return { ok: true, status: 200, json: async () => body } as unknown as Response;
+}
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  vi.useRealTimers();
+});
+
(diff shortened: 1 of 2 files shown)


apps/desktop/src/main/services/fetch-leaderboard.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+    const data = (await res.json()) as LeaderboardResponseBody;
+    const benchmark = mapBenchmark(data.benchmark);
+    if (benchmark === null) {
+      publishLog(`leaderboard: GET ${url} -> malformed benchmark in response body`);
+      return null;
+    }
+    const entries = Array.isArray(data.entries)
+      ? data.entries.map(mapEntry).filter((e): e is LeaderboardEntry => e !== null)
+      : [];
```
HupBaHa, line 136: This function documents `null` for a malformed response, but a missing/non-array `entries` currently becomes a successful empty leaderboard. That hides schema drift and makes the malformed-body path inconsistent with the benchmark guard.

```suggestion
    if (!Array.isArray(data.entries)) {
      publishLog(`leaderboard: GET ${url} -> malformed entries in response body`);
      return null;
    }
    const entries = data.entries.map(mapEntry).filter((e): e is LeaderboardEntry => e !== null);
```

apps/desktop/src/main/services/__tests__/fetch-leaderboard.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    const out = await listLeaderboard('livetruth');
+
+    expect(out?.entries).toHaveLength(1);
+    expect(out?.entries[0].rank).toBe(1);
+  });
+
+  it('returns null when the benchmark field is missing/malformed', async () => {
+    global.fetch = vi.fn(async () => okResponse({ entries: [ENTRY] })) as unknown as typeof fetch;
+    expect(await listLeaderboard('livetruth')).toBeNull();
+  });
```
HupBaHa, line 104: This should be covered with the malformed-body cases. I verified the test fails on the current implementation and passes with the guard above.

```suggestion
  });

  it('returns null when the entries field is missing/malformed', async () => {
    global.fetch = vi.fn(async () => okResponse({ benchmark: BENCHMARK })) as unknown as typeof fetch;
    expect(await listLeaderboard('livetruth')).toBeNull();
  });
```

sergio-bershadsky:
Desktop Application will be aggresively refactored please consider to introduce this changes later

PR #364, opened 2026-07-07
feat(desktop): expose leaderboard reads over IPC + preload

FilipBolt:
## Summary
- Promotes `LeaderboardBenchmark`/`LeaderboardEntry`/`LeaderboardData` into `preload/types.ts` (matching `KnownBenchmark`'s convention, since they now cross the IPC boundary).
- Adds a `leaderboard:getLeaderboard` IPC channel in its own file (`leaderboard.ipc.ts`), not folded into `publish.ipc.ts` (that file is scoped to the write/publish flow), guarded by `requireTrustedIpcSender`.
- Exposes it on `window.electronAPI.leaderboard.getLeaderboard`.
- `listLeaderboard()` itself is behaviorally unchanged from the previous PR; still no renderer hook or UI consuming this yet.

Stacked on #363 (milestone 1 of 6 for OME-321) — review that one first.

## Test plan
- [x] New: `leaderboard.ipc.test.ts` — registration, sender validation, delegation to `listLeaderboard`, null-propagation
- [x] `npx vitest run` — 51 files / 318 tests, all passing
- [x] `npm run lint` — 0 errors
- [x] `npm run build` — succeeds
- [x] Verified this branch in isolation (stashed all later-milestone work, reran full gates) before stacking milestone 3 on top

Linear: [OME-321](https://linear.app/openmined/issue/OME-321/sf-27-leaderboard-pull-the-public-board-into-desktop-sdk)

PR #365, opened 2026-07-07
feat(desktop): add useLeaderboard renderer hook

FilipBolt:
## Summary
- Adds `useLeaderboard(benchmarkId)`, loading the ranked leaderboard via `window.electronAPI.leaderboard.getLeaderboard`, modeled on `use-known-benchmarks.ts`'s `{data, loading}` shape.
- Adds `error: boolean`, distinct from a benchmark with zero submissions (a successful response with an empty `entries` array — not an error).
- Adds `refresh()` for manual retry — reads don't auto-retry with backoff the way publish does (see rationale in `fetch-leaderboard.ts`: a read has no idempotency-safety reason to retry silently).
- Still no UI consumer.

Stacked on #364 (milestone 2 of 6 for OME-321).

## Test plan
- [x] New: `use-leaderboard.test.ts` — null-benchmarkId no-op, load/resolve, null-result-as-error (distinct from empty entries), rejected-IPC-as-error, re-fetch on benchmarkId change, refresh() on demand
- [x] `npx vitest run` — 52 files / 324 tests, all passing
- [x] `npm run lint` — 0 errors
- [x] `npm run build` — succeeds
- [x] Verified this branch in isolation before stacking milestone 4 on top

Linear: [OME-321](https://linear.app/openmined/issue/OME-321/sf-27-leaderboard-pull-the-public-board-into-desktop-sdk)

### DIFF
--- a/apps/desktop/src/renderer/src/hooks/use-leaderboard.ts
+++ b/apps/desktop/src/renderer/src/hooks/use-leaderboard.ts
@@ -1,0 +1,62 @@
+// apps/desktop/src/renderer/src/hooks/use-leaderboard.ts
+//
+// Loads the ranked leaderboard for one benchmark (GET /v1/leaderboard/{id} via
+// the main process, exempt from renderer CORS). Re-fetches whenever
+// `benchmarkId` changes; `refresh()` re-fetches the current benchmark on
+// demand (manual retry — see fetch-leaderboard.ts for why reads don't retry
+// automatically the way publish does).
+//
+// `error` is distinct from "no entries": a benchmark with zero submissions is
+// a successful, non-null response with an empty `entries` array.
+
+import { useCallback, useEffect, useState } from 'react';
+import type { LeaderboardData } from '../../../../preload/types';
+
+export interface LeaderboardState {
+  data: LeaderboardData | null;
+  loading: boolean;
+  error: boolean;
+  refresh: () => void;
+}
+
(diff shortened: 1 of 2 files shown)


apps/desktop/src/renderer/src/hooks/use-leaderboard.ts
```diff
@@ -0,0 +1,62 @@
...
+// the main process, exempt from renderer CORS). Re-fetches whenever
+// `benchmarkId` changes; `refresh()` re-fetches the current benchmark on
+// demand (manual retry — see fetch-leaderboard.ts for why reads don't retry
+// automatically the way publish does).
+//
+// `error` is distinct from "no entries": a benchmark with zero submissions is
+// a successful, non-null response with an empty `entries` array.
+
+import { useCallback, useEffect, useState } from 'react';
+import type { LeaderboardData } from '../../../../preload/types';
```
HupBaHa, line 13: This path goes up four levels from `src/renderer/src/hooks`, which resolves to `apps/desktop/preload/types` (no such file). The preload types live under `apps/desktop/src/preload/types.ts`, so this should only go up three levels. `tsc -b --noEmit` reports this as `TS2307` even though Vitest/build pass because this is a type-only import.

```suggestion
import type { LeaderboardData } from '../../../preload/types';
```

sergio-bershadsky:
DEsktop app will be refactored aggresively consider to apply this PR later

PR #379, opened 2026-07-09
feat(scoreboard): import single-model baselines with source attribution

FilipBolt:
## Summary

- Adds a `Baseline` model (`apps/scoreboard/src/scoreboard/scores/models/baseline.py`), FK to `Benchmark`, unique on `(benchmark, model_name, source)` so re-imports upsert in place instead of duplicating.
- Adds `BaselineStore` (`import_baseline`, `list_baselines`) — separate from `ScoreStore` since a baseline is a distinct concept from a community submission (no url4_expression/provider/correctness fields).
- Adds `import_baselines.py`, a CLI mirroring the existing `seed.py` pattern (`python -m scoreboard.import_baselines`, `SCOREBOARD_SEED_BASELINES_JSON` env var).
- Surfaces baselines via a new `baselines` field on `GET /v1/leaderboard/{benchmark_id}` — additive, no breaking change to the existing response shape.
- New migration `0002_auto_20260709_1350.py`, applied + re-applied to confirm idempotency.

**Scoped deliberately to mechanism only** (agreed with the ticket owner at kickoff):
- No real baseline data seeded this pass — `livetruth`/`livetruth-latest` are OpenMined-proprietary benchmarks with no external LMArena/Artificial Analysis coverage; only `hle` has real data to import, left for a follow-up once sourced.
- No portal UI rendering of a "target line" — the portal has no reference-line concept today; that's an explicit follow-up, not part of this PR.

## Test plan

- [x] `uv run pytest --cov=scoreboard --cov-fail-under=80` — 95 passed, 1 skipped, 87.57% coverage
- [x] Migration applied twice — second run reports no pending migrations
- [x] New tests: model (table/FK/unique_together), schema validation, store (insert/upsert/unknown-benchmark/ordering), route (`baselines` field empty + populated), CLI (JSON-in/store-state-out)
- [x] All prior tests remain green and unmodified (verified via `git diff` — pure additions only)

Refs: OME-322

### DIFF
--- a/apps/scoreboard/src/scoreboard/import_baselines.py
+++ b/apps/scoreboard/src/scoreboard/import_baselines.py
@@ -31,11 +31,7 @@ def load_baselines_json(raw_json: str) -> list[BaselineImportRow]:
 
 
 async def import_baselines(rows: Sequence[BaselineImportRow]) -> list[BaselineSchema]:
-    store = BaselineStore()
-    imported: list[BaselineSchema] = []
-    for row in rows:
-        imported.append(await store.import_baseline(row))
-    return imported
+    return await BaselineStore().import_many(rows)
 
 
 def _build_parser() -> argparse.ArgumentParser:
--- a/apps/scoreboard/src/scoreboard/import_baselines.py
+++ b/apps/scoreboard/src/scoreboard/import_baselines.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import os
+from collections.abc import Sequence
+
+from pydantic import TypeAdapter, ValidationError
+
+from .config import Settings
+from .db import close_db, init_db
+from .scores.baseline_store import BaselineStore
+from .scores.schemas import BaselineImportRow, BaselineSchema
+
+SEED_BASELINES_ENV = "SCOREBOARD_SEED_BASELINES_JSON"
+
+_BASELINES_ADAPTER = TypeAdapter(list[BaselineImportRow])
+
+
+def load_baselines_json(raw_json: str) -> list[BaselineImportRow]:
+    try:
+        payload = json.loads(raw_json)
+    except json.JSONDecodeError as exc:
+        raise ValueError(f"invalid baseline import JSON: {exc.msg}") from exc
+
+    try:
+        return _BASELINES_ADAPTER.validate_python(payload)
+    except ValidationError as exc:
+        raise ValueError(f"invalid baseline import payload: {exc}") from exc
+
+
+async def import_baselines(rows: Sequence[BaselineImportRow]) -> list[BaselineSchema]:
+    store = BaselineStore()
+    imported: list[BaselineSchema] = []
+    for row in rows:
+        imported.append(await store.import_baseline(row))
+    return imported
+
+
+def _build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description="Import scoreboard single-model baselines.")
+    parser.add_argument(
+        "--baselines-json",
+        default=None,
+        help=f"JSON baseline list. Defaults to ${SEED_BASELINES_ENV}; empty list if unset.",
+    )
+    return parser
+
+
+async def _run(raw_json: str) -> None:
+    rows = load_baselines_json(raw_json)
+    if not rows:
+        print("no baselines configured")
+        return
+
+    settings = Settings()
+    await init_db(settings.database_url)
+    try:
+        imported = await import_baselines(rows)
+        for baseline in imported:
+            print(f"imported baseline {baseline.model_name} for {baseline.benchmark_id}")
+    finally:
+        await close_db()
+
+
+def main(argv: Sequence[str] | None = None) -> None:
+    parser = _build_parser()
+    args = parser.parse_args(argv)
+    raw_json = args.baselines_json or os.getenv(SEED_BASELINES_ENV, "[]")
+    try:
+        asyncio.run(_run(raw_json))
+    except ValueError as exc:
+        parser.error(str(exc))
+
+
+if __name__ == "__main__":
+    main()
--- a/apps/scoreboard/src/scoreboard/scores/models/baseline.py
+++ b/apps/scoreboard/src/scoreboard/scores/models/baseline.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+import uuid
+
+from tortoise import fields
+
+from .base import BaseScoreboardModel
+
+
+class BaseBaseline(BaseScoreboardModel):
+    class Meta:
+        abstract = True
+
+    id = fields.UUIDField(primary_key=True, default=uuid.uuid4)
+    model_name = fields.CharField(max_length=255)
+    accuracy = fields.FloatField()
+    source = fields.CharField(max_length=64)
+    source_url = fields.CharField(max_length=2048, null=True)
+    imported_at = fields.DatetimeField(auto_now_add=True)
+    metadata = fields.JSONField(null=True)
+
+
+class Baseline(BaseBaseline):
+    class Meta:
+        table = "baselines"
+        unique_together = (("benchmark", "model_name", "source"),)
+
+    benchmark = fields.ForeignKeyField(
+        "models.Benchmark",
+        related_name="baselines",
+        on_delete=fields.OnDelete.RESTRICT,
+    )
--- a/apps/scoreboard/src/scoreboard/scores/schemas.py
+++ b/apps/scoreboard/src/scoreboard/scores/schemas.py
@@ -1,10 +1,25 @@
 from __future__ import annotations
 
+import json
 from datetime import datetime
 from typing import Annotated, Any, Literal
 from uuid import UUID
 
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+
+# INVARIANT: a baseline's metadata is operator-supplied (via the import CLI, not a
+# public HTTP endpoint) but still bounded, so one bad import can't make
+# GET /v1/leaderboard/{id} fail to serialize for every consumer (found in PR review).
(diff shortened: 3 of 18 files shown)


apps/scoreboard/tests/unit/test_import_baselines.py
```diff
@@ -0,0 +1,61 @@
...
+from scoreboard.scores.baseline_store import BaselineStore
+from scoreboard.scores.store import ScoreStore
+
+pytestmark = pytest.mark.asyncio
+
+
+async def _register_benchmark(benchmark_id: str = "hle") -> None:
+    await ScoreStore().register_benchmark(
+        benchmark_id=benchmark_id,
+        display_name="Humanity's Last Exam",
```
HupBaHa: Could we clarify the intended identity of `hle` before real baseline data is imported? Production currently defines `hle` as **News Hallucinations**, while this fixture and the OME-322 context use it as **Humanity’s Last Exam**. Since the importer only checks that the ID exists, an Artificial Analysis HLE score could be attached to the wrong board without an error.

Would it make sense to introduce a distinct immutable benchmark/revision ID for Humanity’s Last Exam and update the related import payloads accordingly? I am not including an `Apply suggestion` here because the safe fix spans the benchmark registry, import contract, tests, and task documentation.

apps/scoreboard/src/scoreboard/scores/schemas.py
```diff
@@ -153,3 +153,48 @@ class LeaderboardEntry(BaseModel):
...
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    benchmark_id: str
+    model_name: str
+    accuracy: float
+    source: str
+    source_url: str | None = None
+    metadata: dict[str, Any] | None = None
```
HupBaHa, line 228: Could we constrain `metadata` before exposing it through the public response? I reproduced a case where a deeply nested metadata object passed validation and persisted successfully, but `GET /v1/leaderboard/hle` then returned `500` because Pydantic could not serialize the response. One accepted baseline can therefore make the entire board unavailable.

A bounded metadata DTO or explicit size/depth validation, together with an API-level serialization test, would prevent this. I am not including an `Apply suggestion` here because the safe fix should cover both the import DTO and the public response contract.

apps/scoreboard/src/scoreboard/scores/schemas.py
```diff
@@ -153,3 +153,48 @@ class LeaderboardEntry(BaseModel):
...
+    """Input DTO for importing a single-model baseline score (e.g. from LMArena /
+    Artificial Analysis). Re-importing the same (benchmark_id, model_name, source)
+    updates the existing row rather than duplicating it (see BaselineStore).
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    benchmark_id: str
+    model_name: str
+    accuracy: float
```
HupBaHa: Could we make this field strict? Pydantic currently converts JSON `true` to `1.0`, `false` to `0.0`, and numeric strings to floats, so malformed source data can silently become a plausible baseline.

```suggestion
    accuracy: Annotated[float, Field(strict=True, ge=0, le=1, allow_inf_nan=False)]
```

It would also be helpful to add JSON-level tests for booleans, numeric strings, and non-finite values.

apps/scoreboard/src/scoreboard/import_baselines.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        return _BASELINES_ADAPTER.validate_python(payload)
+    except ValidationError as exc:
+        raise ValueError(f"invalid baseline import payload: {exc}") from exc
+
+
+async def import_baselines(rows: Sequence[BaselineImportRow]) -> list[BaselineSchema]:
+    store = BaselineStore()
+    imported: list[BaselineSchema] = []
+    for row in rows:
```
HupBaHa: Would it make sense to make this batch import all-or-nothing? At the moment, if an earlier row succeeds and a later row fails, the earlier rows remain committed even though the command exits with an error.

Pre-validating the complete batch or wrapping the batch in one transaction would give operators clearer and safer behavior. I am not including an `Apply suggestion` here because the transaction needs to span the store calls rather than only this loop.

apps/scoreboard/src/scoreboard/scores/models/baseline.py
```diff
@@ -0,0 +1,32 @@
...
+class BaseBaseline(BaseScoreboardModel):
+    class Meta:
+        abstract = True
+
+    id = fields.UUIDField(primary_key=True, default=uuid.uuid4)
+    model_name = fields.CharField(max_length=255)
+    accuracy = fields.FloatField()
+    source = fields.CharField(max_length=64)
+    source_url = fields.CharField(max_length=2048, null=True)
+    imported_at = fields.DatetimeField(auto_now_add=True)
```
HupBaHa, line 19: Could we capture source/evaluation freshness before these values are treated as a “line to beat”? Re-import currently updates `accuracy`, while `imported_at` remains the timestamp of the first value, and there is no structured dataset or evaluation-protocol revision. This makes it difficult to tell when the current score was measured or whether values from different sources are directly comparable.

Perhaps `evaluated_at`/`updated_at` plus an immutable dataset or protocol revision could be added here, or explicitly tracked as required before the real-data follow-up. I am not including an `Apply suggestion` because the correct change should update the model, migration, schema, mapper, and tests together.

apps/scoreboard/tests/unit/test_import_baselines.py
FilipBolt: Good catch on the hle mismatch. Do you want a distinct benchmark id for the real Humanity's Last Exam, or a code level safeguard for now while the registry naming gets sorted separately?

apps/scoreboard/src/scoreboard/scores/schemas.py
FilipBolt: Fixed. metadata is now bounded on the import DTO (depth and serialized size), so nothing unbounded can reach storage or the response.

apps/scoreboard/src/scoreboard/scores/schemas.py
FilipBolt: Fixed with strict plus allow_inf_nan false. Kept the existing range validator as its own check so the error message a prior test asserts on stays the same.

apps/scoreboard/src/scoreboard/import_baselines.py
FilipBolt: Fixed. Added BaselineStore.import_many, one transaction for the whole batch, all or nothing. import_baselines now delegates to it.

apps/scoreboard/src/scoreboard/scores/schemas.py
```diff
@@ -153,3 +168,64 @@ class LeaderboardEntry(BaseModel):
...
+    benchmark_id: str
+    model_name: str
+    # WHY: strict + no-inf-nan closes a Pydantic v2 laziness gap where JSON true/false
+    # coerce to 1.0/0.0 and numeric strings coerce to float, letting malformed source
+    # data silently become a plausible-looking score (found in PR review). The range
+    # check stays a separate validator below so its error message doesn't change for
+    # an existing test.
+    accuracy: Annotated[float, Field(strict=True, allow_inf_nan=False)]
+    source: str
+    source_url: str | None = None
```
HupBaHa: Could we restrict attribution links to HTTP(S)? This value is returned through the public API and will likely be rendered as a link by future clients, while `javascript:` and `data:` values currently pass validation.

```suggestion
    source_url: Annotated[str, Field(pattern=r"^https?://", max_length=2048)] | None = None
```

This also keeps the API validation aligned with the existing 2048-character database limit.

apps/scoreboard/tests/unit/test_import_baselines.py
HupBaHa: Thanks — I checked the registry history, and `hle` / News Hallucinations appears to have been introduced as a demo/dev seed rather than the canonical external Humanity’s Last Exam benchmark.

I agree that a source-specific safeguard does not belong in `BaselineStore`, and I do not think this mechanism-only PR needs to create the final production benchmark ID.

Let’s keep the importer generic and track a separate registry/data follow-up that introduces a canonical immutable ID/revision for the real Humanity’s Last Exam before any external baseline data is imported. Optionally, using a neutral benchmark ID/name in these new fixtures would avoid reinforcing the demo `hle` ambiguity, but I would not block this PR on that test-only cleanup.

apps/scoreboard/src/scoreboard/scores/schemas.py
HupBaHa: On SQLite, which is the default database, this value is accepted and persisted, after which the leaderboard returns 500. PostgreSQL rejects it during insertion instead, so the application-level validation is currently database-dependent.

HupBaHa (approved):

apps/scoreboard/src/scoreboard/scores/schemas.py
FilipBolt: Fixed with a field validator requiring http or https, plus a max length of 2048 to match the DB column. Kept it as a validator rather than a Field pattern for style consistency with the rest of the file.

apps/scoreboard/tests/unit/test_import_baselines.py
FilipBolt: Filed OME-396 for the canonical id decision, assigned to you. Also renamed this ticket's own test fixtures from hle to demo-benchmark, scoped to just the tests this PR owns.

apps/scoreboard/src/scoreboard/scores/schemas.py
FilipBolt: Patched. Bound BaselineSchema.metadata too, same check as the import DTO. Also made list_baselines skip a row that fails validation instead of crashing the whole response, that's the part that actually keeps the board up if one row is bad.

PR #383, opened 2026-07-09
fix(repo): append-only gate checks line-level diffs, not file status

FilipBolt:
## Summary

Fixes OME-369: `append_only_check()` in `.claude/scripts/run_gates.py` used to flag any git-modified test file as a rule-5 violation, even for pure additions. This PR replaces the file-status check with real content-level analysis — and, through 11 rounds of review-driven hardening, fixes a series of false positives AND false negatives in the mechanism itself.

**Final design:**
- **Protected ranges** (`_old_protected_ranges`): every function body (tests, fixtures, helpers alike — not just `test_*`-named), decorator lines included, plus every direct module-level `Assign`/`AnnAssign`/`AugAssign` statement (shared test data like `_BASE_KW = {...}`). Each range carries its definition column. BOM'd files stay protected; undecodable committed files fall to the permissive side.
- **Change detection** (`_diff_positions`): diffs actual line content — old bytes via `git show`, new bytes off the working tree — with `difflib.SequenceMatcher`, NOT by parsing `git diff` text (three separate bug classes came from text-format quirks: header false-matching, replace-pair anchor drift, EOF-newline artifacts; all impossible by construction now). Binary content yields a fail-closed verdict, not a crash.
- **Violation rule**: a removed/changed line inside a protected range (`lo <= ln <= hi`), an insertion strictly inside one (`lo <= n < hi`), or an insertion at a range's end indented deeper than its definition column (extends the old body, e.g. appending `break` inside a final test's loop). Appending a new sibling function/method at the same anchor stays legitimate.
- **File-level statuses**: `D`(elete), `R`(ename), and `T`(ypechange — e.g. replaced by a symlink) are always violations; `A`/`C` always fine. Non-ASCII filenames handled via `core.quotepath=off`.

**Documented, deliberately-deferred gaps** (in the code's AIDEV-NOTE + the work ledger; follow-up tickets drafted, filing queued behind this re-review): stacking a new outermost decorator on an existing test (needs old-vs-new AST identity matching), name-shadowing/monkeypatching incl. mutation-in-place (structural limit of any line-diff approach), class-level attributes, conditionally-nested data statements, bare walrus statements. Separately: nothing in CI independently enforces this check at all — its own infra ticket draft.

## Test plan

Permanent 32-test suite at `.claude/scripts/tests/test_run_gates.py` (stdlib `unittest`, throwaway git repo per test): `uv run .claude/scripts/tests/test_run_gates.py`. Every fix landed with a discriminating regression test proven to fail on the pre-fix code first; boundary guards pin the legitimate-append cases (new sibling def, new class method, module constants, import edits, docstring/`__main__` edits). Full history of all 11 rounds, each with fail-before/pass-after evidence: `docs/work/2026-07-09-OME-369-append-only-line-diff.md`.

Refs: OME-369

### DIFF
--- a/.claude/scripts/run_gates.py
+++ b/.claude/scripts/run_gates.py
@@ -18,9 +18,9 @@ Exit codes: 0 all green · 1 a gate or the append-only check failed · 2 config 
 
 import argparse
 import ast
+import difflib
 import fnmatch
 import pathlib
-import re
 import subprocess
 import sys
 from typing import NoReturn
@@ -65,9 +65,6 @@ def load_card(path: pathlib.Path) -> dict:
     if not isinstance(data.get("stacks"), list):
         fail_config(f"{path} frontmatter must define a `stacks:` list")
     return data
-
-
-_HUNK_HEADER = re.compile(r"^@@ -(\d+)(?:,(\d+))? \+\d+(?:,\d+)? @@")
 
 
 # WHY: an ALLOWLIST of data-holding node types, not a denylist of "everything
@@ -141,94 +138,64 @@ def _old_protected_ranges(root: pathlib.Path, base: str, path: str) -> list[tupl
 
 
 def _diff_positions(root: pathlib.Path, base: str, path: str) -> tuple[set[int], set[int]]:
-    """Old-file positions the diff for `path` vs `base` touches, in two forms:
+    """Old-file positions where `path`'s content differs between `base` and the
+    working tree, in two forms:
 
-    - `removed`: an old line number itself changed/removed (a `-` line) — checked
-      against a protected range INCLUSIVE on both ends (`lo <= ln <= hi`).
+    - `removed`: an old line number itself changed/removed — checked against a
+      protected range INCLUSIVE on both ends (`lo <= ln <= hi`).
     - `inserted_after`: an old line number that NEW content was inserted directly
       after, with no old line consumed for it — checked EXCLUSIVE at the upper end
       (`lo <= n < hi`).
     """
+    # WHY: diff actual line CONTENT (old file via `git show`, new file read
+    # straight off disk) with difflib.SequenceMatcher, rather than hand-parsing
+    # `git diff`'s unified-diff TEXT. Three separate rounds of bugs (a removed
+    # line starting with `--` false-matching a text-based header check; a
+    # replace pair's insertion anchor landing one line later than it should;
+    # an EOF-newline-only change rendering as a remove+add of unchanged text,
+    # in a way that couldn't be fixed correctly for multi-line hunks without
+    # LCS-style pairing) all stemmed from the same root cause: unified-diff TEXT
+    # has representational quirks (headers, no-newline markers, replace-pair
+    # ambiguity) that have nothing to do with whether a line's CONTENT actually
+    # changed. SequenceMatcher's opcodes are computed directly from line
+    # sequences, so a line byte-identical between old and new is always
+    # "equal" — none of those text-format artifacts can arise by construction.
+    #
     # AIDEV-NOTE: the exclusive bound on `inserted_after` is load-bearing. A pure
     # insertion (e.g. a whole new test function typed directly after an existing
     # one, zero blank lines between them) anchors at `n == hi` of the PRECEDING
     # function — that must stay unflagged, or this reopens the exact false-positive
     # OME-369 exists to fix (pure additions rejected as violations). Inserting as a
     # new first/middle statement anchors at `lo <= n < hi` and must be caught: a
-    # pure `+`-only diff can still silently neuter a prior test's assertions (e.g.
-    # forcing a variable's value right before the check) with zero `-` lines, which
-    # `removed` alone can never see.
-    #
-    # AIDEV-NOTE: file-level header lines (`diff --git`, `index`, `--- a/...`,
-    # `+++ b/...`) only ever appear BEFORE the first `@@` hunk line — track that
-    # boundary structurally (`in_hunk`) instead of prefix-matching `---`/`+++`
-    # against line content. A *removed* line whose actual content starts with `--`
-    # at column 0 (e.g. a bare `---` separator) renders as `----` in the diff,
-    # which would false-match a content-based header check and silently vanish
-    # from `removed` while desyncing `old_line` for every later line in the hunk.
-    #
-    # WHY: also track whether the most recent "-" line was immediately followed
-    # by "\ No newline at end of file" and, if so, whether the very next "+"
-    # line is byte-identical to it. git represents "content was appended after a
-    # file that didn't end in a newline" as a remove+add of that unchanged last
-    # line (its EOF status changed, not its text) — without this check, that
-    # renders as a false "removed/changed" hit on a protected range's last line
-    # for the ordinary, non-adversarial act of appending new content to a file
-    # that happened not to end in `\n`.
-    proc = subprocess.run(
-        ["git", "diff", "--relative", "--unified=0", base, "--", path],
-        cwd=root, capture_output=True, text=True,
+    # pure insertion can still silently neuter a prior test's assertions (e.g.
+    # forcing a variable's value right before the check) with zero removed lines,
+    # which `removed` alone can never see.
+    old_proc = subprocess.run(
+        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True, text=True,
     )
-    if proc.returncode != 0:
-        fail_config(f"git diff failed in {root}: {proc.stderr.strip()}")
+    if old_proc.returncode != 0:
+        return set(), set()  # file didn't exist at base — nothing to protect
+    # AIDEV-NOTE: `append_only_check` only ever calls this for status "M" files,
+    # which are guaranteed to exist in the working tree — reading directly off
+    # disk is the correct equivalent of `git diff base -- path`'s implicit
+    # "against the working tree" comparison (no second ref given).
+    new_content = (root / path).read_text()
+    old_lines = old_proc.stdout.splitlines()
+    new_lines = new_content.splitlines()
+    matcher = difflib.SequenceMatcher(a=old_lines, b=new_lines, autojunk=False)
     removed: set[int] = set()
     inserted_after: set[int] = set()
-    old_line = 0
-    in_hunk = False
-    pure_insert_hunk = False
-    last_removed: tuple[int, str, bool] | None = None  # (old_line, content, no_newline)
-    for line in proc.stdout.splitlines():
-        if line.startswith("@@"):
-            m = _HUNK_HEADER.match(line)
-            if m:
-                old_line = int(m.group(1))
-                old_count = int(m.group(2)) if m.group(2) is not None else 1
-                # WHY: a hunk whose header declares old-count 0 is UNAMBIGUOUSLY a
-                # pure insertion straight from the header. Inferring "pure insert"
-                # by pairing "-"/"+" lines line-by-line instead would mis-anchor a
-                # replace pair (a "-" immediately followed by a "+") one line later
-                # than the line it actually replaces — which can land exactly on
-                # the start of the NEXT protected range and false-positive an edit
-                # that never touched it (e.g. replacing the blank separator line
-                # between two functions falsely flags the second function).
-                pure_insert_hunk = old_count == 0
-            in_hunk = True
-            last_removed = None
-        elif not in_hunk:
-            continue  # pre-hunk file header (diff --git / index / --- a/ / +++ b/)
-        elif line.startswith("\\"):
-            # "\ No newline at end of file" describes the line just emitted —
-            # never consumes a line number, but mark it on a pending removal so
-            # a byte-identical "+" right after can be recognized as an EOF-status
-            # artifact, not a real edit.
-            if last_removed is not None:
-                last_removed = (last_removed[0], last_removed[1], True)
-        elif line.startswith("-"):
-            removed.add(old_line)
-            last_removed = (old_line, line[1:], False)
-            old_line += 1
-        elif line.startswith("+"):
-            if last_removed is not None and last_removed[2] and last_removed[1] == line[1:]:
-                removed.discard(last_removed[0])
-            elif pure_insert_hunk:
-                inserted_after.add(old_line)  # anchored right after the current old_line
-            # else: part of a replace pair with a preceding "-" in THIS hunk —
-            # already captured via `removed`; anchoring it too would double-count
-            # and can mis-fire per the WHY note above.
-            last_removed = None
-        else:
-            last_removed = None
-            old_line += 1  # context line (present only if a non-zero unified context is used)
+    for tag, i1, i2, _j1, _j2 in matcher.get_opcodes():
+        if tag == "equal":
+            continue
+        if tag in ("delete", "replace"):
+            removed.update(range(i1 + 1, i2 + 1))  # old_lines is 0-indexed, ranges are 1-indexed
+        if tag == "insert":
+            inserted_after.add(i1)  # anchored right after old 1-indexed line i1
+        # "replace" deliberately does NOT also populate inserted_after — the
+        # removed lines already comprehensively signal the violation if inside
+        # a range, and anchoring the replacement's added lines too would
+        # double-count without adding coverage.
     return removed, inserted_after
 
 
--- a/.claude/scripts/run_gates.py
+++ b/.claude/scripts/run_gates.py
@@ -17,8 +17,10 @@ Exit codes: 0 all green · 1 a gate or the append-only check failed · 2 config 
 """
 
 import argparse
+import ast
 import fnmatch
 import pathlib
+import re
 import subprocess
 import sys
 from typing import NoReturn
@@ -65,8 +67,61 @@ def load_card(path: pathlib.Path) -> dict:
     return data
 
 
+_HUNK_HEADER = re.compile(r"^@@ -(\d+)(?:,\d+)? \+\d+(?:,\d+)? @@")
+
+
+def _old_test_ranges(root: pathlib.Path, base: str, path: str) -> list[tuple[int, int]]:
+    """Line ranges (1-indexed, inclusive) of every test function body in `path` at `base`."""
+    proc = subprocess.run(
+        ["git", "show", f"{base}:{path}"], cwd=root, capture_output=True, text=True,
+    )
+    if proc.returncode != 0:
+        return []  # file didn't exist at base — nothing to protect
+    try:
+        tree = ast.parse(proc.stdout)
+    except SyntaxError:
+        return []  # can't parse — fall through to the safe (permissive) side
+    ranges = []
+    for node in ast.walk(tree):
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and node.name.startswith("test_"):
+            ranges.append((node.lineno, getattr(node, "end_lineno", node.lineno)))
+    return ranges
+
+
+def _removed_old_lines(root: pathlib.Path, base: str, path: str) -> set[int]:
+    """Old-file line numbers that a removal/change touches, per the diff for `path` vs `base`."""
+    proc = subprocess.run(
+        ["git", "diff", "--relative", "--unified=0", base, "--", path],
+        cwd=root, capture_output=True, text=True,
+    )
+    if proc.returncode != 0:
+        fail_config(f"git diff failed in {root}: {proc.stderr.strip()}")
+    removed: set[int] = set()
+    old_line = 0
+    for line in proc.stdout.splitlines():
+        if line.startswith("@@"):
+            m = _HUNK_HEADER.match(line)
+            old_line = int(m.group(1)) if m else old_line
+        elif line.startswith(("---", "+++", "\\")):
+            continue
+        elif line.startswith("-"):
+            removed.add(old_line)
+            old_line += 1
+        elif line.startswith("+"):
+            continue  # inserted line — doesn't consume an old line number
+        else:
+            old_line += 1  # context line (present only if a non-zero unified context is used)
+    return removed
+
+
 def append_only_check(root: pathlib.Path, base: str, globs: list[str]) -> bool:
-    """True when no previously committed test file was modified/deleted/renamed."""
+    """True when no previously committed *test* was modified/deleted (rule 5).
+
+    Pure additions to a test file (new test functions, new imports for them) are
+    always fine — only a removed/changed line that falls INSIDE a test function body
+    that already existed at `base` counts as a violation. Deleting or renaming a
+    whole test file is always a violation regardless of content.
+    """
     proc = subprocess.run(
         ["git", "diff", "--relative", "--name-status", base],
         cwd=root,
@@ -83,10 +138,20 @@ def append_only_check(root: pathlib.Path, base: str, globs: list[str]) -> bool:
             status[:1] not in "MDR"
         ):  # A(dded)/C(opied) etc. are fine — adding tests is always fine
             continue
-        for p in paths:
-            if any(fnmatch.fnmatch(p, g) for g in globs):
-                offenders.append(f"  {status}\t{p}")
-                break
+        matched = [p for p in paths if any(fnmatch.fnmatch(p, g) for g in globs)]
+        if not matched:
+            continue
+        p = matched[-1]  # for R(ename), name-status gives "old\tnew" — the new path is what's on disk
+        if status[:1] != "M":
+            offenders.append(f"  {status}\t{p}")
+            continue
+        test_ranges = _old_test_ranges(root, base, p)
+        bad_lines = sorted(
+            ln for ln in _removed_old_lines(root, base, p)
+            if any(lo <= ln <= hi for lo, hi in test_ranges)
+        )
+        if bad_lines:
+            offenders.append(f"  M\t{p}  (removed/changed inside an existing test, old line(s) {bad_lines})")
     if offenders:
         print(
             f"{RED} append-only test check — prior tests were modified/deleted (vs {base}):"
--- a/.claude/scripts/run_gates.py
+++ b/.claude/scripts/run_gates.py
@@ -88,14 +88,17 @@ def _old_protected_ranges(root: pathlib.Path, base: str, path: str) -> list[tupl
     # returns `[]` (falls into the "didn't exist" branch) for every stack whose
     # root isn't the repo root, i.e. every real stack in `.claude/sdlc.local.md`.
     proc = subprocess.run(
-        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True, text=True,
+        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True,
     )
     if proc.returncode != 0:
         return []  # file didn't exist at base — nothing to protect
     try:
-        tree = ast.parse(proc.stdout)
-    except SyntaxError:
-        return []  # can't parse — fall through to the safe (permissive) side
+        # WHY: bytes + errors="replace", not text=True — a committed undecodable
+        # file must fall through to the permissive SyntaxError branch below, not
+        # crash the subprocess decode with UnicodeDecodeError.
+        tree = ast.parse(proc.stdout.decode("utf-8", errors="replace"))
+    except (SyntaxError, ValueError):
+        return []  # can't parse (incl. null bytes) — fall through to the safe (permissive) side
     ranges = []
     # INVARIANT: covers EVERY function, not just `test_*`-named ones — a
     # `conftest.py` fixture or a plain helper a test depends on is just as
@@ -170,8 +173,13 @@ def _diff_positions(root: pathlib.Path, base: str, path: str) -> tuple[set[int],
     # pure insertion can still silently neuter a prior test's assertions (e.g.
     # forcing a variable's value right before the check) with zero removed lines,
     # which `removed` alone can never see.
+    # WHY: compare BYTES on both sides, never decoded text — a working-tree file
+    # rewritten with undecodable/binary content would make a text-mode read raise
+    # UnicodeDecodeError and crash the gate with a traceback instead of a verdict.
+    # Byte lines diff identically for ordinary text, and binary junk replacing a
+    # protected test simply differs line-wise → flagged (fail-closed), no crash.
     old_proc = subprocess.run(
-        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True, text=True,
+        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True,
     )
     if old_proc.returncode != 0:
         return set(), set()  # file didn't exist at base — nothing to protect
@@ -179,9 +187,8 @@ def _diff_positions(root: pathlib.Path, base: str, path: str) -> tuple[set[int],
     # which are guaranteed to exist in the working tree — reading directly off
     # disk is the correct equivalent of `git diff base -- path`'s implicit
     # "against the working tree" comparison (no second ref given).
-    new_content = (root / path).read_text()
     old_lines = old_proc.stdout.splitlines()
-    new_lines = new_content.splitlines()
+    new_lines = (root / path).read_bytes().splitlines()
     matcher = difflib.SequenceMatcher(a=old_lines, b=new_lines, autojunk=False)
     removed: set[int] = set()
     inserted_after: set[int] = set()
--- a/.claude/scripts/run_gates.py
+++ b/.claude/scripts/run_gates.py
@@ -20,6 +20,8 @@ import argparse
 import ast
 import difflib
 import fnmatch
+import io
+import os
 import pathlib
 import subprocess
 import sys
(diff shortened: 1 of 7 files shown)


FilipBolt:
Heads up, I'm out July 20–31 — this PR is still unreviewed. Appreciate a look if anyone's available.

.claude/scripts/run_gates.py
```diff
@@ -60,8 +62,61 @@ def load_card(path: pathlib.Path) -> dict:
...
+    """Line ranges (1-indexed, inclusive) of every test function body in `path` at `base`."""
+    proc = subprocess.run(
+        ["git", "show", f"{base}:{path}"], cwd=root, capture_output=True, text=True,
+    )
+    if proc.returncode != 0:
+        return []  # file didn't exist at base — nothing to protect
+    try:
+        tree = ast.parse(proc.stdout)
+    except SyntaxError:
+        return []  # can't parse — fall through to the safe (permissive) side
```
HupBaHa: Thanks for putting this together. There is a path-coordinate issue in the new lookup: `git diff --relative` yields `tests/...` relative to `cwd=root`, while `<base>:<path>` is resolved from the repository root. I reproduced `_old_test_ranges(apps/aigateway, "HEAD", "tests/conftest.py") == []`, so changing an assertion in a nested stack passes. Since this branch handles an `M` file, the base blob should exist; lookup or parse failures should stop rather than silently disable protection. This uses Git's `:./path` syntax relative to `cwd` and fails closed:

```suggestion
    proc = subprocess.run(
        ["git", "show", f"{base}:./{path}"],
        cwd=root,
        capture_output=True,
        text=True,
    )
    if proc.returncode != 0:
        fail_config(f"git show failed in {root}: {proc.stderr.strip()}")
    try:
        tree = ast.parse(proc.stdout)
    except SyntaxError as exc:
        fail_config(f"cannot parse {base}:./{path}: {exc}")
```

.claude/scripts/run_gates.py
```diff
@@ -60,8 +62,61 @@ def load_card(path: pathlib.Path) -> dict:
...
+        if line.startswith("@@"):
+            m = _HUNK_HEADER.match(line)
+            old_line = int(m.group(1)) if m else old_line
+        elif line.startswith(("---", "+++", "\\")):
+            continue
+        elif line.startswith("-"):
+            removed.add(old_line)
+            old_line += 1
+        elif line.startswith("+"):
+            continue  # inserted line — doesn't consume an old line number
```
HupBaHa: Could we avoid treating every added line as automatically safe? With the new `+` branch, this additive edit changes an old test without removing any old line:

```diff
 def test_existing():
+    return
     assert response.status_code == 403
```

I reproduced `append_only_check()` returning `True` for this case. Comparing the complete old/current AST for each pre-existing test node, rather than only removed line numbers, would catch this while still allowing a separately added test. Please add this as a regression case.

.claude/scripts/run_gates.py
```diff
@@ -60,8 +62,61 @@ def load_card(path: pathlib.Path) -> dict:
...
+    if proc.returncode != 0:
+        return []  # file didn't exist at base — nothing to protect
+    try:
+        tree = ast.parse(proc.stdout)
+    except SyntaxError:
+        return []  # can't parse — fall through to the safe (permissive) side
+    ranges = []
+    for node in ast.walk(tree):
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and node.name.startswith("test_"):
+            ranges.append((node.lineno, getattr(node, "end_lineno", node.lineno)))
```
HupBaHa: The protected range starts at `node.lineno` (the `def` line) and only includes functions named `test_*`. In the new logic I reproduced both of these passing: reducing `@pytest.mark.parametrize(..., [1, 2])` to `[1]`, and changing a fixture used by an old test. Python AST already exposes decorators through `FunctionDef.decorator_list`; could we protect the complete pre-existing test definition, including decorators, and explicitly cover test-support definitions that can change old-test behavior? Regression cases for a decorator and a fixture would make the boundary clear.

.claude/scripts/run_gates.py
```diff
@@ -60,8 +62,61 @@ def load_card(path: pathlib.Path) -> dict:
...
+    )
+    if proc.returncode != 0:
+        fail_config(f"git diff failed in {root}: {proc.stderr.strip()}")
+    removed: set[int] = set()
+    old_line = 0
+    for line in proc.stdout.splitlines():
+        if line.startswith("@@"):
+            m = _HUNK_HEADER.match(line)
+            old_line = int(m.group(1)) if m else old_line
+        elif line.startswith(("---", "+++", "\\")):
```
HupBaHa: This prefix check also matches hunk content, not only file headers. A removed source line beginning with `---` is emitted as `----...`, so this branch skips it as metadata. I reproduced changing a `---protected` line inside a multiline string in an old test and the check passed. Could the parser distinguish file headers before the first `@@` from content inside a hunk, with a regression case for a dash-prefixed source line?

HupBaHa (changes requested):
Thanks for addressing OME-369. I checked the PR head with temporary Git repositories: adding a separate test passes and a root-level rewrite is rejected, but the inline cases below still allow prior tests to be changed in the configured nested stacks. Requesting changes because this gate enforces SDLC rule 5. A small permanent temporary-repo matrix covering nested roots, additive edits inside an old test, decorators/fixtures, and dash-prefixed hunk content would make this behavior much safer.

FilipBolt:
Addressed both remaining P2 comments from the last review round:

**"Reject edits to existing module-level test helpers"** — fixed. `_old_protected_ranges` now also covers module-level `Assign`/`AnnAssign` statements (e.g. a shared `_BASE_KW = {...}` dict several tests build on), not just function bodies. Imports stay exempt, per the earlier explicit decision that editing an existing import line to support a new test is legitimate.

**"Reject decorators added to existing tests"** — confirmed valid, but deliberately **not** fixed in this PR. Closing it needs old-vs-new AST identity matching (compare a function's decorators by name across commits) rather than the line-position diffing this PR is built on — a different, larger mechanism. Filing a follow-up ticket to track it rather than block this PR on it or rush a fix.

Before asking for re-review, I ran a structured multi-angle code-review pass on the pushed diff and it caught two real bugs in my own fix for the module-data comment above:

- The module-data protection was implemented as a denylist (`_EXEMPT_TOP_LEVEL`), which reopened the exact false-positive problem this whole PR exists to fix — editing a bare module docstring, an `if __name__ == "__main__":` block, or an import nested inside a version-guard all got falsely flagged. Fixed by inverting to an allowlist (`_MODULE_LEVEL_DATA = (Assign, AnnAssign)`) — only node types that actually hold shared test data get protected.
- A separate anchor-computation bug: replacing the blank line between two functions with a comment falsely flagged the second function, because the insertion anchor was computed after advancing past the removed line. Fixed by keying pure-insertion detection off the diff hunk header's declared old-line count instead of inferring it from `-`/`+` pairing.

20/20 tests pass now (4 new ones covering both fixes, each confirmed to fail on the pre-fix code first). Full detail in `docs/work/2026-07-09-OME-369-append-only-line-diff.md` (rounds 4-5).

Two more follow-ups are being tracked as separate Linear tickets rather than folded into this PR: the decorator-stacking gap above, and a broader name-shadowing/monkeypatching limitation found during review (redefining/reassigning/deleting a name anywhere later in a file can silently neuter a prior test without touching its lines — a structural limit of line-diffing, not fixable by another patch to this mechanism). A third, unrelated finding — no CI workflow or hook independently re-verifies this check at all — is also being tracked separately as its own infra unit.

FilipBolt:
@HupBaHa this is ready for another look whenever you have time — and worth a fresh read rather than a delta review, since the mechanism changed substantially since your last pass.

Since your CHANGES_REQUESTED review: both inline P2s are addressed (module-level test data is protected; decorator-stacking is confirmed real and tracked as OME-475 rather than patched here — it needs AST identity matching, a different mechanism). Your ask for a permanent temp-repo matrix became a 32-test suite, and every subsequent fix landed with a discriminating test proven to fail on the pre-fix code first.

Beyond your comments, several more review rounds were run until one came back clean. Highlights: the diff-text parsing you originally reviewed is gone entirely (replaced by `difflib.SequenceMatcher` over byte content — three separate bug classes stemmed from text-format quirks), and a series of verified fail-opens were fixed: BOM'd files losing all protection, binary content crashing the gate, symlink typechanges skipped, non-ASCII filenames escaping globs, and indented appends extending a final test's body (`break` flipping a failing loop test green). The PR body now describes the final design; `docs/work/2026-07-09-OME-369-append-only-line-diff.md` has the full round-by-round evidence.

Known, deliberately-deferred gaps are documented in the code's AIDEV-NOTE and ticketed: OME-475 (decorator-stacking), OME-476 (shadowing/monkeypatching — a structural limit of any line-diff gate), OME-477 (nothing in CI enforces this check at all — likely the highest-leverage follow-up).

One heads-up: Filip is out July 20–31, so a review before then would be appreciated if feasible.

FilipBolt:
@HupBaHa — the nested-stack coverage gap you flagged (root-level rewrite rejected, but inline cases under a configured nested stack still allowed changing prior tests) has been addressed since your review — see round 13, commit e1c16de4 ("record round 13 — rebase + empirical re-verification"). Re-requesting your review.

.claude/scripts/run_gates.py
```diff
@@ -17,6 +17,8 @@
 """
 
 import argparse
+import ast
+import difflib
 import fnmatch
 import pathlib
 import subprocess
```
HupBaHa, line 26: These imports support byte-safe tokenization and NUL-delimited Git path decoding.

```suggestion
import ast
import difflib
import fnmatch
import io
import os
import pathlib
import subprocess
```

.claude/scripts/run_gates.py
```diff
@@ -65,28 +67,229 @@ def load_card(path: pathlib.Path) -> dict:
...
+    #     structural name-shadowing/monkeypatching limitation tracked as its own
+    #     follow-up (line-diffing can't see what a statement's *effect* is, only
+    #     its position), not a gap this allowlist could ever close by growing.
+    # (5) stacking a brand-NEW outermost decorator (e.g. @pytest.mark.skip) onto
+    #     a previously-existing function isn't caught — it anchors at the same
+    #     diff position as legitimately inserting a new function directly above,
+    #     which line positions can't distinguish. Needs old-vs-new AST identity
+    #     matching (compare each function's decorator list across versions) —
+    #     a separate deferred follow-up, same status as (4).
+    return ranges
```
HupBaHa, line 190: Fail closed for unsupported or unparseable existing artifacts, and protect direct assertions plus existing class metadata without blocking new methods.

```suggestion
# WHY: an ALLOWLIST of data-holding node types, not a denylist of "everything
# except X". A denylist means every statement type nobody thought to exclude
# (a docstring, an `if __name__ == "__main__":` block, an import nested inside
# a version-guard) silently becomes a false positive the next time someone
# edits one. Assign/AnnAssign/AugAssign hold the kind of shared test data (e.g.
# a `_BASE_KW = {...}` dict, a `_CASES += [...]` accumulator, or a direct
# collection-time assertion) rule 5 needs to protect.
_MODULE_LEVEL_DATA = (ast.Assign, ast.AnnAssign, ast.AugAssign, ast.Assert)


def _class_header_end(tokens: list, node: ast.ClassDef) -> int:
    """Last source line in a class header, excluding the class body."""
    started = False
    depth = 0
    for token in tokens:
        if not started:
            started = token.string == "class" and token.start == (
                node.lineno,
                node.col_offset,
            )
            continue
        if token.string in "([{":
            depth += 1
        elif token.string in ")]}" and depth:
            depth -= 1
        elif token.string == ":" and depth == 0:
            return token.end[0]
    return node.lineno


def _old_protected_ranges(
    root: pathlib.Path, base: str, path: str
) -> list[tuple[int, int, int | None]] | None:
    """(start_line, end_line, def_column) triples — lines 1-indexed inclusive — for
    every protected node in `path` at `base`: every function body, plus every direct
    module-level Assign/AnnAssign/AugAssign statement. `def_column` is the node's
    own column offset, used to tell "new sibling appended after this body" apart
    from "new line appended INTO this body" (see append_only_check).
    """
    import tokenize

    # AIDEV-NOTE: `path` is cwd-relative (it comes from `git diff --relative`), but
    # `git show rev:path` resolves a bare path relative to the REPO ROOT, not cwd —
    # the `./` prefix is what makes it cwd-relative. Without it, this silently
    # returns `[]` (falls into the "didn't exist" branch) for every stack whose
    # root isn't the repo root, i.e. every real stack in `.claude/sdlc.local.md`.
    # Non-Python files still appear in active stack globs (for example the
    # aigateway-ui TypeScript tests). Until a language-aware range parser exists,
    # modifications to an existing unsupported artifact keep the old fail-closed
    # behavior; A/C statuses remain legitimate additions in append_only_check.
    if pathlib.PurePosixPath(path).suffix != ".py":
        return None

    proc = subprocess.run(
        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True,
    )
    if proc.returncode != 0:
        return None
    try:
        # Passing bytes lets Python honor UTF-8 BOM and PEP-263 source encodings.
        tree = ast.parse(proc.stdout)
        tokens = list(tokenize.tokenize(io.BytesIO(proc.stdout).readline))
    except (IndentationError, SyntaxError, UnicodeDecodeError, ValueError, tokenize.TokenError):
        return None
    ranges = []
    # INVARIANT: covers EVERY function, not just `test_*`-named ones — a
    # `conftest.py` fixture or a plain helper a test depends on is just as
    # protected as the test itself.
    for node in ast.walk(tree):
        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
            # INVARIANT: an EXISTING decorator (e.g. @pytest.mark.parametrize(...),
            # @pytest.fixture) is part of the protected body — node.lineno points
            # at `def`, not the decorator line(s) above it. Stacking a brand-NEW
            # outermost decorator onto an old function is NOT caught — see gap (5).
            start = node.decorator_list[0].lineno if node.decorator_list else node.lineno
            ranges.append((start, getattr(node, "end_lineno", node.lineno), node.col_offset))
        elif isinstance(node, ast.ClassDef):
            # Protect only existing decorators and the class header/bases, not the
            # whole body: adding a new sibling method must remain a pure addition.
            start = node.decorator_list[0].lineno if node.decorator_list else node.lineno
            # `None` disables the function-only end-of-body indentation rule:
            # content after a class header is its body, where a first new method is
            # a legitimate addition rather than an extension of old test logic.
            ranges.append((start, _class_header_end(tokens, node), None))
    # INVARIANT: module-level test data is just as protected as a fixture — a
    # `-` line there is invisible to the function-only pass above.
    for node in tree.body:  # direct top-level children only — see AIDEV-NOTE below
        if isinstance(node, _MODULE_LEVEL_DATA):
            ranges.append((node.lineno, getattr(node, "end_lineno", node.lineno), node.col_offset))
    # AIDEV-NOTE: known, deliberate gaps — deferred follow-ups (ticket filing
    # queued behind the PR re-review, per owner instruction), not chased here:
    # (1) class-level attributes (e.g. a shared constant on a unittest.TestCase
    #     subclass) aren't covered — would need the same per-statement treatment
    #     inside class bodies, not a whole-class-span shortcut (that would
    #     reopen "insert a new method between two existing ones" as a false
    #     positive, the same way OME-369's original bug worked).
    # (2) an Assign/AnnAssign/AugAssign nested inside a module-level If/Try/With
    #     (e.g. a version-guarded conditional constant) isn't covered either,
    #     since this loop only walks direct `tree.body` children, not recursively.
    # (3) a bare module-level walrus statement (`(_TOTAL := 10)`, parsed as an
    #     ast.Expr wrapping an ast.NamedExpr) isn't covered — an exotic enough
    #     pattern in real test code that it isn't worth a special-case check.
    #     NOTE: unlike (4), adding `ast.NamedExpr` to `_MODULE_LEVEL_DATA` would
    #     NOT fix this — `tree.body`'s direct child here is the outer `ast.Expr`
    #     wrapper, never the inner `NamedExpr` itself, so that "just add the
    #     type" pattern (used for AugAssign) silently does nothing for this case.
    # (4) mutating a protected object in place (`_CASES.append(...)`,
    #     `del _CASES[1]`, `_BASE_KW["x"] = 999`) rather than rebinding its name
    #     isn't covered by ANY allowlist entry, by design — this is the same
    #     structural name-shadowing/monkeypatching limitation tracked as its own
    #     follow-up (line-diffing can't see what a statement's *effect* is, only
    #     its position), not a gap this allowlist could ever close by growing.
    # (5) stacking a brand-NEW outermost decorator (e.g. @pytest.mark.skip) onto
    #     a previously-existing function isn't caught — it anchors at the same
    #     diff position as legitimately inserting a new function directly above,
    #     which line positions can't distinguish. Needs old-vs-new AST identity
    #     matching (compare each function's decorator list across versions) —
    #     a separate deferred follow-up, same status as (4).
    return ranges
```

.claude/scripts/run_gates.py
```diff
@@ -65,28 +67,229 @@ def load_card(path: pathlib.Path) -> dict:
...
+            for line in new_lines[j1:j2]:
+                if line.strip():
+                    indent = len(line) - len(line.lstrip(b" \t"))
+                    break
+            inserted_after[i1] = indent  # anchored right after old 1-indexed line i1
+        # "replace" deliberately does NOT also populate inserted_after — the
+        # removed lines already comprehensively signal the violation if inside
+        # a range, and anchoring the replacement's added lines too would
+        # double-count without adding coverage.
+    return removed, inserted_after
```
HupBaHa: Validate current Python source fully, inspect the added side of replacements, derive indentation from tokens, and make glob matching component-aware.

```suggestion
def _added_span_indent(
    tokens: list | None,
    new_lines: list[bytes],
    start: int,
    end: int,
) -> int:
    """Indent of the first code token in new_lines[start:end], else first text."""
    import tokenize

    ignored = {
        tokenize.COMMENT,
        tokenize.DEDENT,
        tokenize.ENCODING,
        tokenize.ENDMARKER,
        tokenize.INDENT,
        tokenize.NEWLINE,
        tokenize.NL,
    }
    if tokens is not None:
        for token in tokens:
            if start + 1 <= token.start[0] <= end and token.type not in ignored:
                line = new_lines[token.start[0] - 1]
                prefix = line[: token.start[1]]
                return len(prefix.rsplit(b"\f", 1)[-1])
    else:
        # Invalid current source will fail later gates, but must not fail open here.
        return max(
            (
                len(
                    line[: len(line) - len(line.lstrip(b" \t\f"))].rsplit(b"\f", 1)[-1]
                )
                for line in new_lines[start:end]
                if line.strip()
            ),
            default=0,
        )
    for line in new_lines[start:end]:
        if line.strip():
            prefix = line[: len(line) - len(line.lstrip(b" \t\f"))]
            return len(prefix.rsplit(b"\f", 1)[-1])
    return 0


def _diff_positions(
    root: pathlib.Path, base: str, path: str
) -> tuple[set[int], dict[int, int], bool]:
    """Old-file positions where `path`'s content differs between `base` and the
    working tree, in two forms:

    - `removed`: an old line number itself changed/removed — checked against a
      protected range INCLUSIVE on both ends (`lo <= ln <= hi`).
    - `inserted_after`: maps each old line number that NEW content was inserted
      directly after (no old line consumed) to the INDENT (leading spaces/tabs)
      of the first non-blank inserted line — checked EXCLUSIVE at the upper end
      (`lo <= n < hi`), except that an insertion anchored exactly at a range's
      end (`n == hi`) whose indent is deeper than the range's own definition
      column extends that body and is a violation too (see append_only_check).
    """
    import tokenize

    # WHY: diff actual line CONTENT (old file via `git show`, new file read
    # straight off disk) with difflib.SequenceMatcher, rather than hand-parsing
    # `git diff`'s unified-diff TEXT. Three separate rounds of bugs (a removed
    # line starting with `--` false-matching a text-based header check; a
    # replace pair's insertion anchor landing one line later than it should;
    # an EOF-newline-only change rendering as a remove+add of unchanged text,
    # in a way that couldn't be fixed correctly for multi-line hunks without
    # LCS-style pairing) all stemmed from the same root cause: unified-diff TEXT
    # has representational quirks (headers, no-newline markers, replace-pair
    # ambiguity) that have nothing to do with whether a line's CONTENT actually
    # changed. SequenceMatcher's opcodes are computed directly from line
    # sequences, so a line byte-identical between old and new is always
    # "equal" — none of those text-format artifacts can arise by construction.
    #
    # AIDEV-NOTE: the exclusive bound on `inserted_after` is load-bearing. A pure
    # insertion (e.g. a whole new test function typed directly after an existing
    # one, zero blank lines between them) anchors at `n == hi` of the PRECEDING
    # function — that must stay unflagged, or this reopens the exact false-positive
    # OME-369 exists to fix (pure additions rejected as violations). Inserting as a
    # new first/middle statement anchors at `lo <= n < hi` and must be caught: a
    # pure insertion can still silently neuter a prior test's assertions (e.g.
    # forcing a variable's value right before the check) with zero removed lines,
    # which `removed` alone can never see.
    # WHY: compare BYTES on both sides, never decoded text — a working-tree file
    # rewritten with undecodable/binary content would make a text-mode read raise
    # UnicodeDecodeError and crash the gate with a traceback instead of a verdict.
    # Byte lines diff identically for ordinary text, and binary junk replacing a
    # protected test simply differs line-wise → flagged (fail-closed), no crash.
    old_proc = subprocess.run(
        ["git", "show", f"{base}:./{path}"], cwd=root, capture_output=True,
    )
    if old_proc.returncode != 0:
        return set(), {}, True  # file didn't exist at base — nothing to protect
    # AIDEV-NOTE: `append_only_check` only ever calls this for status "M" files,
    # which are guaranteed to exist in the working tree — reading directly off
    # disk is the correct equivalent of `git diff base -- path`'s implicit
    # "against the working tree" comparison (no second ref given).
    old_lines = old_proc.stdout.splitlines()
    new_source = (root / path).read_bytes()
    new_lines = new_source.splitlines()
    try:
        compile(new_source, path, "exec")
        new_tokens = list(tokenize.tokenize(io.BytesIO(new_source).readline))
    except (
        IndentationError,
        SyntaxError,
        UnicodeDecodeError,
        ValueError,
        tokenize.TokenError,
    ):
        new_tokens = None
    current_parseable = new_tokens is not None
    matcher = difflib.SequenceMatcher(a=old_lines, b=new_lines, autojunk=False)
    removed: set[int] = set()
    inserted_after: dict[int, int] = {}
    for tag, i1, i2, j1, j2 in matcher.get_opcodes():
        if tag == "equal":
            continue
        if tag in ("delete", "replace"):
            removed.update(range(i1 + 1, i2 + 1))  # old_lines is 0-indexed, ranges are 1-indexed
        if tag in ("insert", "replace") and j1 < j2:
            # A replace can remove an unprotected separator while adding code that
            # extends the preceding protected body. Comments are not indentation
            # tokens, so use the first real Python token when one exists.
            indent = _added_span_indent(new_tokens, new_lines, j1, j2)
            inserted_after[i1] = max(inserted_after.get(i1, 0), indent)
    return removed, inserted_after, current_parseable


def _matches_glob(path: str, pattern: str) -> bool:
    """Match slash-delimited globs; only a complete `**` spans directories."""
    path_parts = path.split("/")
    pattern_parts = pattern.split("/")
    memo: dict[tuple[int, int], bool] = {}

    def matches(path_index: int, pattern_index: int) -> bool:
        key = (path_index, pattern_index)
        if key in memo:
            return memo[key]
        if pattern_index == len(pattern_parts):
            result = path_index == len(path_parts)
        elif pattern_parts[pattern_index] == "**":
            result = matches(path_index, pattern_index + 1) or (
                path_index < len(path_parts)
                and matches(path_index + 1, pattern_index)
            )
        else:
            result = path_index < len(path_parts) and fnmatch.fnmatchcase(
                path_parts[path_index], pattern_parts[pattern_index]
            ) and matches(path_index + 1, pattern_index + 1)
        memo[key] = result
        return result

    return matches(0, 0)
```

.claude/scripts/run_gates.py
```diff
@@ -65,28 +67,229 @@ def load_card(path: pathlib.Path) -> dict:
...
+        # inside a range (`lo <= n < hi`), and (b) anchored exactly at a range's
+        # end (`n == hi`) with the first non-blank inserted line indented DEEPER
+        # than the range's own definition column: that extends the old body
+        # (e.g. appending `break` inside a final test's loop flips it green with
+        # zero old lines touched), whereas a new sibling def/decorator at the
+        # same anchor starts at the same-or-shallower column and stays legitimate.
+        bad_inserted = sorted(
+            n for n, indent in inserted_after.items()
+            if any(lo <= n < hi or (n == hi and indent > col) for lo, hi, col in ranges)
+        )
```
HupBaHa: Use NUL-delimited status records so unusual filenames cannot bypass the gate, then enforce the new fail-closed and class-range results.

```suggestion
    # NUL delimiters preserve tabs, newlines, quotes, and backslashes in paths;
    # core.quotepath=off additionally keeps non-ASCII names unescaped.
    proc = subprocess.run(
        [
            "git", "-c", "core.quotepath=off", "diff", "--relative",
            "--name-status", "-z", base,
        ],
        cwd=root, capture_output=True,
    )
    if proc.returncode != 0:
        fail_config(f"git diff failed in {root}: {os.fsdecode(proc.stderr).strip()}")
    offenders = []
    fields = proc.stdout.split(b"\0")
    if fields[-1:] == [b""]:
        fields.pop()
    field_index = 0
    while field_index < len(fields):
        status = os.fsdecode(fields[field_index])
        field_index += 1
        path_count = 2 if status[:1] in "RC" else 1
        if field_index + path_count > len(fields):
            fail_config(f"malformed git diff --name-status output in {root}")
        paths = [os.fsdecode(p) for p in fields[field_index:field_index + path_count]]
        field_index += path_count
        # WHY: T(ypechange) is in the offender set — replacing a committed test
        # file with a symlink wholesale swaps its effective content, exactly like
        # a delete+recreate. A(dded)/C(opied) stay fine: adding tests is always fine.
        if status[:1] not in "MDRT":
            continue
        matched = [p for p in paths if any(_matches_glob(p, g) for g in globs)]
        if not matched:
            continue
        p = matched[-1]  # for R(ename), name-status gives "old\tnew" — the new path is what's on disk
        if status[:1] != "M":
            offenders.append(f"  {status}\t{p}")
            continue
        ranges = _old_protected_ranges(root, base, p)
        if ranges is None:
            offenders.append(f"  M\t{p}  (existing test artifact is unsupported or unparseable)")
            continue
        removed, inserted_after, current_parseable = _diff_positions(root, base, p)
        if not current_parseable:
            offenders.append(f"  M\t{p}  (current Python test file is unparseable)")
            continue
        bad_removed = sorted(ln for ln in removed if any(lo <= ln <= hi for lo, hi, _ in ranges))
        # INVARIANT: two distinct insertion violations — (a) anchored strictly
        # inside a range (`lo <= n < hi`), and (b) anchored exactly at a range's
        # end (`n == hi`) with the first non-blank inserted line indented DEEPER
        # than the range's own definition column: that extends the old body
        # (e.g. appending `break` inside a final test's loop flips it green with
        # zero old lines touched), whereas a new sibling def/decorator at the
        # same anchor starts at the same-or-shallower column and stays legitimate.
        bad_inserted = sorted(
            n for n, indent in inserted_after.items()
            if any(
                lo <= n < hi or (n == hi and col is not None and indent > col)
                for lo, hi, col in ranges
            )
        )
```

HupBaHa (commented):
Validated suggestions for the remaining append-only gate correctness gaps. Applied together on an exact-head disposable copy, these pass all 48 tests on Python 3.10 and 3.12; byte-compilation and whitespace checks pass, and Ruff remains baseline-identical. Two adversarial reviews found no blocking issues.

.claude/scripts/tests/test_run_gates.py
```diff
@@ -0,0 +1,586 @@
...
+        git's default core.quotepath, a non-ASCII filename is emitted as an
+        octal-escaped quoted C-string that matches no glob — rewrites (and
+        deletions) of such files were silently skipped."""
+        with tempfile.TemporaryDirectory() as tmp:
+            root = pathlib.Path(tmp)
+            _init_repo(root)
+            _write(root / "test_ä.py", "def test_one():\n    assert 1 == 1\n")
+            base = _commit_all(root, "base")
+            _write(root / "test_ä.py", "def test_one():\n    assert 1 == 2\n")
+            self.assertFalse(self._check(root, base))
```
HupBaHa, line 543: Cover path characters that break line/tab-delimited Git status parsing.

```suggestion
            _write(root / "test_ä.py", "def test_one():\n    assert 1 == 2\n")
            self.assertFalse(self._check(root, base))

    def test_git_quoted_filename_rewrite_detected(self):
        """NUL-delimited git output must preserve path characters that trigger
        C-style quoting or collide with line/tab-delimited status parsing."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            path = root / 'test_\t\n\\"case.py'
            _write(path, "def test_one():\n    assert 1 == 1\n")
            base = _commit_all(root, "base")
            _write(path, "def test_one():\n    assert 1 == 2\n")
```

.claude/scripts/tests/test_run_gates.py
```diff
@@ -0,0 +1,586 @@
...
+                root / "test_a.py",
+                "class TestThings:\n    def test_one(self):\n        assert 1 == 1\n",
+            )
+            base = _commit_all(root, "base")
+            _write(
+                root / "test_a.py",
+                "class TestThings:\n    def test_one(self):\n        assert 1 == 1\n\n"
+                "    def test_two(self):\n        assert 2 == 2\n",
+            )
+            self.assertTrue(self._check(root, base))
```
HupBaHa, line 594: Add regression and boundary coverage for the remaining fail-open paths and legitimate append-only cases.

```suggestion
            )
            self.assertTrue(self._check(root, base))

    def test_existing_typescript_assertion_rewrite_fails_closed(self):
        """A configured non-Python test file must not lose all protection merely
        because Python's AST cannot parse it. Until a language-aware range parser
        exists, modifications to an existing unsupported test artifact fail closed."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            _write(
                root / "src" / "nested" / "auth.test.ts",
                'test("auth", () => {\n  expect(authorize()).toBe(false);\n});\n',
            )
            base = _commit_all(root, "base")
            _write(
                root / "src" / "nested" / "auth.test.ts",
                'test("auth", () => {\n  expect(true).toBe(true);\n});\n',
            )
            self.assertFalse(self._check(root, base, ["src/**/*.test.ts"]))

    def test_replacing_separator_with_indented_code_fails(self):
        """The added side of a SequenceMatcher replace matters: replacing the
        old blank immediately after a loop test with `break` extends that test."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = (
                "CASES = [1, 2, 3]\n\n\ndef test_all_cases():\n"
                "    for case in CASES:\n        assert case < 3\n\n\n"
                "def test_other():\n    assert True\n"
            )
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                base_src.replace(
                    "        assert case < 3\n\n",
                    "        assert case < 3\n        break\n",
                    1,
                ),
            )
            self.assertFalse(self._check(root, base))

    def test_comment_before_indented_append_does_not_hide_code(self):
        """Comments do not emit Python indentation tokens. A column-zero comment
        before an indented `break` must not make the body extension look dedented."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = (
                "CASES = [1, 2, 3]\n\n\ndef test_all_cases():\n"
                "    for case in CASES:\n        assert case < 3\n"
            )
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                base_src + "# explanation\n        break\n",
            )
            self.assertFalse(self._check(root, base))

    def test_existing_test_class_decorator_edit_fails(self):
        """Existing test-class metadata is prior test behavior too. This is not
        OME-475's deferred case of stacking a new outer function decorator."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = (
                "import pytest\n\n"
                '@pytest.mark.skipif(False, reason="enabled")\n'
                "class TestThings:\n"
                "    def test_one(self):\n        assert False\n"
            )
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                base_src.replace("skipif(False", "skipif(True"),
            )
            self.assertFalse(self._check(root, base))

    def test_module_level_assert_edit_fails(self):
        """A direct module assertion executes during collection and is test logic,
        not documentation or an exempt runner block."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            _write(
                root / "test_a.py",
                "def behavior():\n    return False\n\nassert behavior()\n",
            )
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                "def behavior():\n    return False\n\nassert True\n",
            )
            self.assertFalse(self._check(root, base))

    def test_direct_child_typescript_test_matches_recursive_glob(self):
        """`**/` includes zero directories in stack-card intent; direct children
        must not escape merely because fnmatch treats the slash literally."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            _write(
                root / "src" / "auth.test.ts",
                'test("auth", () => expect(authorize()).toBe(false));\n',
            )
            base = _commit_all(root, "base")
            _write(
                root / "src" / "auth.test.ts",
                'test("auth", () => expect(true).toBe(true));\n',
            )
            self.assertFalse(self._check(root, base, ["src/**/*.test.ts"]))

    def test_first_method_added_to_existing_class_passes(self):
        """Protect class metadata without treating its header endpoint like a
        function body endpoint: the first new method is a legitimate addition."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            _write(root / "test_a.py", "class TestThings:\n    pass\n")
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                "class TestThings:\n    def test_one(self):\n        assert True\n    pass\n",
            )
            self.assertTrue(self._check(root, base))

    def test_invalid_current_source_fails_closed_on_deep_added_code(self):
        """If tokenization fails, a leading comment must not hide deeper added
        code at a protected endpoint. Invalid current source should fail closed."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = (
                "CASES = [1, 2, 3]\n\n\ndef test_all_cases():\n"
                "    for case in CASES:\n        assert case < 3\n"
            )
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                base_src + "# explanation\n        break\n(\n",
            )
            self.assertFalse(self._check(root, base))

    def test_each_recursive_glob_segment_can_match_zero_directories(self):
        """Multiple `**/` segments vary independently; removing only a prefix
        chain leaves valid direct-child combinations unmatched."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            path = root / "x" / "foo" / "bar.test.ts"
            _write(path, 'test("x", () => expect(false).toBe(true));\n')
            base = _commit_all(root, "base")
            _write(path, 'test("x", () => expect(true).toBe(true));\n')
            self.assertFalse(
                self._check(root, base, ["**/foo/**/bar.test.ts"])
            )

    def test_invalid_source_form_feed_indentation_fails_closed(self):
        """Python accepts form feed in leading indentation. The invalid-source
        fallback must not treat a deeply indented line as column zero."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = (
                "CASES = [1, 2, 3]\n\n\ndef test_all_cases():\n"
                "    for case in CASES:\n        assert case < 3\n"
            )
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                base_src + "# explanation\n\f        break\n(\n",
            )
            self.assertFalse(self._check(root, base))

    def test_form_feed_prefixed_sibling_function_passes(self):
        """At the start of indentation, form feed resets Python's indentation
        column; it must not make a module-level sibling look nested."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            _write(root / "test_a.py", "def test_one():\n    assert True\n")
            base = _commit_all(root, "base")
            _write(
                root / "test_a.py",
                "def test_one():\n    assert True\n\fdef test_two():\n    assert True\n",
            )
            self.assertTrue(self._check(root, base))

    def test_current_tab_error_fails_closed(self):
        """Tokenization alone does not validate indentation. A current file that
        ast.parse rejects with TabError must not pass append-only checking."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = (
                "class TestThings:\n    def test_all(self):\n"
                "        for case in [1, 2, 3]:\n            assert case < 3\n"
            )
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(root / "test_a.py", base_src + "\t\tbreak\n")
            self.assertFalse(self._check(root, base))

    def test_current_compile_time_syntax_error_fails_closed(self):
        """AST construction accepts some contextually invalid statements, so
        current Python source must pass full compilation before it can pass."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            base_src = "def test_one():\n    assert True\n"
            _write(root / "test_a.py", base_src)
            base = _commit_all(root, "base")
            _write(root / "test_a.py", base_src + "break\n")
            self.assertFalse(self._check(root, base))

    def test_globstar_inside_component_is_not_recursive(self):
        """Only a standalone `**` path component has zero-directory semantics;
        textual `**/` inside a larger component must retain fnmatch behavior."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            path = root / "src" / "testauth.test.ts"
            _write(path, 'test("x", () => expect(false).toBe(true));\n')
            base = _commit_all(root, "base")
            _write(path, 'test("x", () => expect(true).toBe(true));\n')
            self.assertTrue(
                self._check(root, base, ["src/test**/auth.test.ts"])
            )

    def test_non_globstar_wildcard_does_not_cross_directories(self):
        """Wildcards within a component must not consume path separators."""
        with tempfile.TemporaryDirectory() as tmp:
            root = pathlib.Path(tmp)
            _init_repo(root)
            path = root / "src" / "test" / "deep" / "auth.test.ts"
            _write(path, 'test("x", () => expect(false).toBe(true));\n')
            base = _commit_all(root, "base")
            _write(path, 'test("x", () => expect(true).toBe(true));\n')
            self.assertTrue(
                self._check(root, base, ["src/test**/auth.test.ts"])
```

FilipBolt:
Applied — thanks for the validated fix. All of it landed in \`298b4d95\`: class-body protection (closes the nested-stack finding), module-level \`assert\` protection, fail-closed handling for unsupported/unparseable test files, the token-based diff/indent rework, the component-aware glob matcher, and NUL-delimited git status parsing. Your 16 new tests are in too — 48/48 passing, matching what you validated.

One thing worth flagging on your end: the review API returned every one of these suggestion comments with `line`/`start_line`/`original_line` all `null`, so they couldn't be applied via GitHub's own "commit suggestion" button — I mapped them onto the file by matching function names instead. Also, the last suggestion (the 15-new-tests block) was truncated mid-statement in the stored comment body itself (cut off inside the final test's closing parens, 11506 chars) — I completed it by mirroring the identical pattern used one test earlier in the same block, and confirmed the count matches (32 original + 1 + 15 = 48, exactly your stated number). Might be worth checking why that comment got cut off if you post suggestions this way again.

Details in the ledger's Round 14.

HupBaHa (approved):
Verified commit 298b4d95 against the final validated patch: both implementation and 48-test suite are byte-identical, including the manually completed tail of the truncated suggestion. Re-ran all 48 tests on Python 3.10 and 3.12; py_compile and Pyright (0 errors) pass, and CI is green. The previously requested correctness fixes are resolved.

PR #390, opened 2026-07-13
fix(scoreboard): server-enforce submission dedup by content hash

FilipBolt:
## Summary
- Server-enforced deduplication of score submissions by hashing the submission's "recipe" (benchmark, spec, url4 expression, result numbers, provider order) — independent of the optional client-supplied `Idempotency-Key` header.
- Adds nullable unique `content_hash` column + migration. `ScoreStore.submit()` returns `SubmitOutcome(score, created)` — a content-hash hit with an `idempotency_key` attached now permanently binds that key to the found score, and the route makes one atomic `submit()` call (no separate pre-check) so the reported status code always matches what actually happened, including under a concurrent-duplicate race.
- `_content_hash()` hashes `correct_questions/total_questions`'s derived ratio, not the client's raw reported `accuracy` — so the same result reported at different float precision still dedupes.
- Step 1 of 2 for OME-391 (C28). Step 2 (write-path auth, C2 — placeholder API-key gate) shipped separately in #392. Step 1's Postgres concurrent-race tests are confirmed dead (broken fixture + no CI service) — tracked separately as OME-430 rather than folded into this PR's app-code scope.

## Test plan
- [x] `uv run pytest --cov=scoreboard --cov-fail-under=80` — 128 passed, 2 skipped (Postgres-only, see OME-430), 87.99% coverage
- [x] Migration verified against a scratch DB seeded with pre-existing rows (clean apply, second run is a no-op)
- [x] Self-review (3 rounds) performed post-commit before first review request
- [x] Addressed all 3 findings from Dmitry's review: idempotency-key rebind bug + route TOCTOU (fixed via the atomic `submit()` + key-binding change above), accuracy-precision hash mismatch (fixed via the normalized-ratio change above), dead Postgres tests (filed OME-430)

Refs: OME-391

### DIFF
--- a/apps/scoreboard/src/scoreboard/scores/store.py
+++ b/apps/scoreboard/src/scoreboard/scores/store.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import hashlib
+import json
 from datetime import UTC, datetime, timedelta
 from typing import cast
 from uuid import UUID
@@ -66,7 +68,26 @@ def _submission_to_kwargs(submission: ScoreSubmission) -> dict[str, object]:
         "client_version": submission.client.version if submission.client else None,
         "client_platform": submission.client.platform if submission.client else None,
         "metadata": submission.metadata,
+        "content_hash": _content_hash(submission),
     }
+
+
+def _content_hash(submission: ScoreSubmission) -> str:
+    # WHY: identity is the recipe (what was run + its result), not who ran it or
+    # when — submitted_by/client_*/ran_at_local/metadata are deliberately excluded.
+    # Provider order is kept as submitted (not sorted) since it's part of what
+    # actually happened, not incidental serialization (OME-391 / C28).
+    identity = {
+        "benchmark_id": submission.benchmark_id,
+        "spec_id": submission.spec_id,
+        "url4_expression": submission.url4_expression,
+        "accuracy": submission.accuracy,
+        "total_questions": submission.total_questions,
+        "correct_questions": submission.correct_questions,
+        "ran_with_providers": submission.ran_with_providers,
+    }
+    encoded = json.dumps(identity, sort_keys=True, separators=(",", ":"))
+    return hashlib.sha256(encoded.encode("utf-8")).hexdigest()
 
 
 def _build_leaderboard_query(benchmark_id: str, top_n: int) -> QueryBuilder:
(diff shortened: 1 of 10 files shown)


apps/scoreboard/src/scoreboard/scores/store.py
```diff
@@ -134,19 +158,35 @@ async def list_benchmarks(self) -> list[BenchmarkSchema]:
...
-                key=idempotency_key,
-                expires_at__gt=now_ts,
-            ).prefetch_related("score")
-            if existing is not None:
-                return _score_to_schema(existing.score)
+        content_hash = _content_hash(submission)
+
+        existing = await self._resolve_existing(idempotency_key, content_hash)
+        if existing is not None:
+            return _score_to_schema(existing)
```
HupBaHa: Could the store return one atomic submit outcome here rather than returning before handling the supplied key? With recipe A stored under key-a, submitting the same recipe under new key-b returns this row without persisting key-b; key-b can then create unrelated recipe B later. The separate route preflight also has a TOCTOU window where concurrent duplicates can both report 201. One submit operation returning the score plus a created flag, while atomically binding a new key on content hits, would make both behaviors consistent.

apps/scoreboard/src/scoreboard/scores/store.py
```diff
@@ -66,9 +68,31 @@ def _submission_to_kwargs(submission: ScoreSubmission) -> dict[str, object]:
...
+    # Provider order is kept as submitted (not sorted) since it's part of what
+    # actually happened, not incidental serialization (OME-391 / C28). `version` is
+    # also excluded — currently a no-op since ScoreSubmission.version is pinned to
+    # Literal[1], but revisit this if a future schema version is ever accepted, since
+    # two payloads differing only in version would otherwise dedupe together.
+    identity = {
+        "benchmark_id": submission.benchmark_id,
+        "spec_id": submission.spec_id,
+        "url4_expression": submission.url4_expression,
+        "accuracy": submission.accuracy,
```
HupBaHa: Could we clarify which result representation is canonical for recipe identity? The route accepts accuracy values within 0.01 of correct_questions / total_questions, so the same 2-of-3 result can be submitted as 0.67 or 0.6666666667; both validate, but this raw float gives them different hashes. If the integer counts are canonical, please omit or normalize accuracy here and add a regression test. If raw reported accuracy is intentionally a distinct result number, documenting that decision would resolve the ambiguity.

apps/scoreboard/tests/unit/scores/test_store.py
```diff
@@ -229,3 +327,18 @@ async def test_postgres_concurrent_idempotency_submissions_share_winner(
 
     assert len({result.id for result in results}) == 1
     assert await Score.all().count() == 1
+
+
+async def test_postgres_concurrent_identical_recipe_submissions_share_winner(
+    tortoise_db: None,
+) -> None:
+    if Tortoise.get_connection("default").capabilities.dialect != "postgres":
```
HupBaHa, line 405: Could we make this PostgreSQL safety test executable in CI? The current workflow does not set SCOREBOARD_TEST_DATABASE_URL, so both PostgreSQL race tests skip. When the variable is set, postgres_schema_database_url calls asyncio.run() from the active pytest-asyncio loop and fixture setup fails before these tests execute. Please use awaited async schema setup/cleanup and add a PostgreSQL CI service/job. It would also be valuable for that job to apply the migration chain and assert that the content_hash unique index exists, since normal tests use generate_schemas() rather than production migrations.

HupBaHa (changes requested):

apps/scoreboard/src/scoreboard/scores/store.py
FilipBolt: Fixed: `ScoreStore.submit()` now returns `SubmitOutcome(score, created)` instead of a bare schema. A content-hash hit with an `idempotency_key` attached now permanently binds that key to the found score via the new `_bind_idempotency_key()` — so a later reuse of the same key for different content can't silently rebind it anymore. Also removed the route's separate `find_existing()` pre-check entirely; it now makes one atomic `submit()` call and reads `.created` for the status code, which closes the TOCTOU you flagged too. Added `test_submit_reused_key_after_content_hash_hit_stays_bound_to_original_score` as a regression test — traced it manually and confirmed it would have failed against the old code. In `39e4ae4`.

apps/scoreboard/src/scoreboard/scores/store.py
FilipBolt: Fixed: `_content_hash()` now hashes `correct_questions / total_questions` (recomputed server-side) instead of the client's raw reported `accuracy`. Added `test_submit_identical_counts_dedupe_despite_different_accuracy_precision` covering the 0.67 vs 0.6666666667 case directly. In `39e4ae4`.

apps/scoreboard/tests/unit/scores/test_store.py
FilipBolt: Confirmed — traced conftest.py: postgres_schema_database_url calls asyncio.run() from inside tortoise_db's already-running pytest-asyncio loop, which would raise RuntimeError the moment SCOREBOARD_TEST_DATABASE_URL is ever actually set, and CI never sets it. Filed OME-430 to track the fixture rewrite + Postgres CI service + migration-drift check as its own unit, since it's CI/workflow infrastructure rather than app code — left AIDEV-NOTE comments on both postgres tests pointing at it. Open to folding it into this PR instead if you'd rather not split it out.

HupBaHa (approved):

PR #391, opened 2026-07-13
feat(scoreboard): add Plausible analytics to the public portal

FilipBolt:
## Summary
- Wires Bennett's Plausible tracking snippet (site id `pa-ysspwNldM0r_4o-m1utPa`, added to OpenMined's existing plan for `scoreboard.screamingface.ai`) into all four public portal pages (`index.html`, `benchmark.html`, `spec.html`, `data.html`).
- OME-373: measures whether the public leaderboard claim is landing and whether externals are actually reaching the verified board.
- Snippet inserted verbatim into each page's `<head>` (they already duplicate their head boilerplate — fonts/icon/theme script — so this follows the existing pattern rather than introducing a templating layer for one script tag).

## Test plan
- [x] New test `test_portal_pages_include_plausible_analytics` — RED confirmed before the HTML edit, GREEN after; all prior portal tests remain green and unmodified
- [x] `uv run pytest --cov=scoreboard --cov-fail-under=80` — 114 passed, 1 skipped, 88.16% coverage
- [x] Verified locally: ran the app against a scratch DB, curled all four portal pages, confirmed each serves the snippet; confirmed Bennett's script URL resolves (200, `application/javascript`); confirmed no CSP would block it

## Rollout
`apps/scoreboard` deploys via a manual `scoreboard-v*` tag (not release-please), so merging alone won't put this live. After a deploy tag + Helm rollout, ping Bennett to verify traffic is showing up in Plausible.

Refs: OME-373

### DIFF
--- a/apps/scoreboard/portal/benchmark.html
+++ b/apps/scoreboard/portal/benchmark.html
@@ -14,6 +14,12 @@
   <link rel="stylesheet" href="style.css">
   <link rel="stylesheet" href="portal.css">
   <script>(function(){try{var t=localStorage.getItem('sf_theme')||'light';document.documentElement.setAttribute('data-theme',t);}catch(e){}})();</script>
+  <!-- Privacy-friendly analytics by Plausible -->
+  <script async src="https://plausible.io/js/pa-ysspwNldM0r_4o-m1utPa.js"></script>
+  <script>
+    window.plausible=window.plausible||function(){(plausible.q=plausible.q||[]).push(arguments)},plausible.init=plausible.init||function(i){plausible.o=i||{}};
+    plausible.init()
+  </script>
 </head>
 <body>
   <div class="rail">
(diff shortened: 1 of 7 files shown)


HupBaHa (approved):
Thanks for keeping this change narrow and for linking the Linear context. I rechecked the current branch, the generated Plausible asset, browser behavior, static serving, Docker packaging, and the scoreboard checks. The integration is working as intended, and I did not find a merge-blocking runtime or security issue.

Approved with two non-blocking follow-ups:

- The regression test currently proves that the script URL is present, but not that the bootstrap and `plausible.init()` remain intact. Please strengthen this in a follow-up so a non-functional integration cannot remain green.
- This unit missed parts of the repository's current SDLC record/process (`spec -> plan -> approval`, the current Linear label taxonomy, and accurate ledger/deviation recording, including the append-only skip). I do not think retrofitting history should block this small PR, but please treat this as an explicit deviation and follow the process for future units.

The exact site-specific snippet was supplied by Bennett, so its enabled Plausible measurements are accepted account configuration rather than a code-review blocker.

PR #392, opened 2026-07-13
feat(scoreboard): gate score submission behind a placeholder API key

FilipBolt:
## Summary
- OME-391 Step 2 of 2 (C2): `POST /v1/scores` currently accepts fully unauthenticated writes. Real identity (OME-326) doesn't exist yet, so Irina + Kevin signed off on stubbing the write-path gate with a single shared placeholder API key rather than blocking on it.
- Adds `SCOREBOARD_SUBMISSION_API_KEY` to `Settings`. Unset (default, every current deployment/test) → no-op, current behavior unchanged. Set → `POST /v1/scores` requires `Authorization: Bearer <key>` (constant-time comparison via `hmac.compare_digest`), else `401`. Reads (`GET /v1/scores/{id}`, leaderboard, portal) stay public regardless.
- This is explicitly **not per-user** — everyone holding the key looks identical to the server; `submitted_by` stays self-reported free text until OME-326 provides real identity to swap in. The whole gate is marked `AIDEV-NOTE: TEMPORARY` in `config.py` and `routes/scores.py`, spelling out exactly what to delete/replace once OME-326 ships.
- Note on scope: builds off `main`, independent of the still-open OME-391 Step 1 PR (#390, dedup) — both touch `routes/scores.py`'s `submit_score`, so whichever merges second will need a small rebase.

## Test plan
- [x] New fixtures (`app_with_api_key`/`gated_score_client`) + 5 new tests: unset-key no-op, correct key succeeds, missing key → 401 (asserts body), wrong key → 401 (asserts body), reads stay public under a configured key
- [x] Extended the existing OpenAPI-schema test to cover the new 401 response
- [x] All prior tests remain green and unmodified — no existing fixture sets the new setting, so the gate's default is a true no-op
- [x] `uv run pytest --cov=scoreboard --cov-fail-under=80` — 118 passed, 1 skipped, 88.33% coverage
- [x] 6 rounds of self-review across 2 follow-up commits: hardened the key comparison to constant-time (`hmac.compare_digest`, avoiding a timing side-channel), strengthened the 401 tests to assert response body, and fixed a real gap in `DEPLOYMENT.md`'s smoke-check runbook (missing `Authorization` header would silently 401 an operator following it verbatim once the key is configured)

Refs: OME-391

### DIFF
--- a/apps/scoreboard/DEPLOYMENT.md
+++ b/apps/scoreboard/DEPLOYMENT.md
@@ -136,7 +136,7 @@ curl -fsS http://scoreboard.40.76.107.241.nip.io/healthz
 curl -fsS http://scoreboard.40.76.107.241.nip.io/v1/benchmarks
 ```
 
-Submit a smoke score with an idempotency key:
+Submit a smoke score with an idempotency key. If `SCOREBOARD_SUBMISSION_API_KEY` is set, add `-H "Authorization: Bearer <key>"` or this 401s:
 
 ```bash
 curl -fsS -X POST http://scoreboard.40.76.107.241.nip.io/v1/scores \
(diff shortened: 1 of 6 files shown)


HupBaHa (approved):
Looks good for the application-level MVP and matches the Step 2 decision in OME-391. One small non-blocking hardening suggestion: could we reject an empty or whitespace-only `SCOREBOARD_SUBMISSION_API_KEY`? At the moment, an empty value can be matched by `Authorization: Bearer `, so startup validation would help prevent an accidental configuration without a real key. This does not block the PR from my side.

PR #402, opened 2026-07-15
feat(url4)!: url4 serve/eval CLI — commands, read registries, path-qualified @ (OME-466)

IonesioJunior:
> **This PR now carries three tracks of `packages/url4` work:**
> 1. **url4 grammar conformance** — the `OME-500` epic (`OME-501/502/504/505/506/507/508`), immediately below.
> 2. **The `url4` serve/eval CLI** — `OME-466`, the original scope, further down under *Summary*.
> 3. **Package reorg + quality** — `OME-499` (subpackage layout), `OME-510` (docstring cross-refs + semantic anchors), `OME-511` (activate the inert PLR1702 gate); see *Package reorg & quality* below.

## url4 grammar conformance (OME-500)

Seven tickets aligning the parser, renderer and wire codec with the normative ABNF. The parser was a **strict superset** — it accepted forms the grammar does not derive. This narrows it to the grammar, and implements two productions that were declared-but-inert. Enforcement lives at the parse / render / wire boundary; the executable DAG is untouched (`Expression(intent=None)` stays an internal carrier for paren-collections and engine wrappers, so map/reduce/gather machinery is unchanged).

**Newly enforced rules** — each column-2 form was silently accepted before and now raises `ParseError`:

| ABNF rule | ✗ Rejected now | ✓ Still valid |
|---|---|---|
| `expression = "(" source-list ")" intent-op intent` | `(a, b, c)` | `(a, b, c)!'merge'` |
| `iteration-expr = collection-ref "*" expression` | `src*(x=$item)` | `src*(x=$item)!'go'` |
| `relative/remote-expr … intent-op intent` | `/summarize(ctx)` | `/summarize(ctx)!'go'` · `/data/$topic` (data URI) |
| `segment = *( ALPHA / DIGIT / "-" / "_" / "." / "~" )` | `/foo$bar(x)!'y'` | `/data/$topic` (wide data-path charset kept) |
| `authority = host [ ":" port ]`, `port = 1*DIGIT` | `url4://host:abc/p(x)!'y'` (non-numeric port) | `url4://node.ai/p(x)!'y'` (no port) · `url4://host:8080/p(x)!'y'` |
| `query-string = *( query-param "&" ) "q=" …` (q= last) | `?q=(x)!'y'&tone=f` | `?tone=f&q=(x)!'y'` |
| `param-key = 1*( ALPHA / DIGIT / "." / "_" )` | `?we-ird=1` | `?coord.rounds=3` |
| `ALPHA` is ASCII (all identifiers) | `$café` · `@9lives` | `$cafe` · `@alice` |
| `structured-weight` is flat | `name:(a:(b:5)):src=…` | `name:(a:0.9,b:0.5):src=…` |

**Behaviour implemented** — declared in the ABNF, previously inert:

- **`intent = value`** — a nested expression in intent position now executes rather than being dropped (`OME-502`).
- **`processor=` delegation (§27.3)** — all three value forms dispatch: an id resolves against the node's routes, a URI fetches absolutely, an expression is evaluated then re-classified once (`OME-506`).
- **`OME-501`** collapsed three duplicated productions onto one source of truth, fixing three silent mis-parses along the way.

**⚠️ Breaking:** intent-less groups, iteration bodies, and relative/remote calls now raise `ParseError` (`code = missing_intent`). The reduce sub-request (`?q=()!…`) and every documented `!intent` form are unaffected; the fix for a rejected expression is to add the intent the grammar always required — e.g. `(a, b)` → `(a, b)!'…'`.

Gates green (ruff · ruff format · pyright · pytest `--cov ≥95`); suite **812 → 1046**. Per-ticket ledgers under `docs/work/2026-07-20-OME-50*`. Two follow-ups filed under the epic: `OME-509` (`!*` on a call; empty `source-list`) and the `pkg/url4-python-sdk` label (owner UI action).

---

## Package reorg & quality (OME-499, OME-510, OME-511)

Structural + documentation work layered on top of the two tracks above. No runtime behaviour changes — the public `import url4` surface is unchanged and the suite stays at **1046 passing**.

**`OME-499` — subpackage layout.** The SDK's 24 flat modules were regrouped into subpackages matching the real import graph (verified acyclic: `core ← io ← dag ← peer ← cli`):

| Subpackage | Holds |
|---|---|
| `core/` | the language + shared runtime (grammar, parser, nodes, render, ensemble, subrequest, …) — a dependency leaf |
| `io/` | the `IOLayer` port + its `http`/`static` adapters |
| `dag/` | the executable-DAG engine (unchanged) + `processor` |
| `peer/` | `Client` + `Url4Node` — the two ways to participate in a url4 network |
| `cli/` | the console script + serve-config assembly |

Pure internal move: `Client`, `Url4Node`, `run`, `StaticIOLayer`, etc. import from `url4` exactly as before; the console entry point moved to `url4.cli.app:main`.

**`OME-510` — docstring cross-refs + semantic anchors.** The reorg left ~94 `:mod:`/`:class:` docstring references pointing at the old flat paths; all repathed to the new subpackages. Then a full pass adopting the `sdlc-python` fixed anchor vocabulary (`WHY:`/`INVARIANT:`/`AIDEV-NOTE:`/`FEATURE:`/`STORY:`) across `src/url4` — 20 genuine additions where un-anchored rationale/invariants were hiding.

**`OME-511` — activate the inert PLR1702 gate.** `PLR1702` (too-many-nested-blocks) was in url4's ruff `select` but silently gated behind preview mode, so `ruff check` printed *"PLR1702 has no effect"* and enforced nothing. Enabled it surgically (`[tool.ruff.lint] preview = true` + `explicit-preview-rules = true`) so **only** that one explicitly-selected preview rule activates — no thresholds relaxed, no broad preview behaviour. Verified it flags 6-deep nesting; zero violations today. Also anchored the two least-anchored modules (`dag/compiler.py` 1→7, `peer/server.py` 0→5).

Per-ticket ledgers under `docs/work/2026-07-20-OME-499|510|511-*`. Gates green throughout (ruff · ruff format · pyright · pytest `--cov ≥95`).

---

## Summary (OME-466 — url4 serve/eval CLI)

Adds the `url4` console script to `packages/url4`: a minimal CLI that runs the engine as an HTTP
node (`url4 serve`) and evaluates expressions locally (`url4 eval`). Closes **OME-466**.

It is built on the released `Url4Node` SDK — its registries and framework-free raw-ASGI `asgi()`
— rather than a bespoke web layer. A url4 expression is the address, so the node speaks GET.
This keeps the feature aligned with engine doctrine (N1 GET-as-address, N4 subprocess leaf,
N5 registry adapters) and keeps `packages/url4`'s core import graph framework-free: **Starlette
is not a dependency**, and uvicorn is reached only through the existing `url4[server]` extra,
imported lazily.

An operator declares the node's **whole surface** in one `url4.toml` — both halves of an
expression:

| Half | Registry | Node port | Resolves |
|---|---|---|---|
| **write** (intent processing) | `[commands]` | `endpoint()` | `/route(ctx)!'intent'`, fan-out reduce |
| **read** (sources) | `[data]` | `data()` | bare relative URIs — `(/rubrics/42)` |
| **read** (holdings) | `[holdings]` | `holdings()` | `@` — the node's own shelves |
| **read** (holdings) | `[identities.<name>]` | `identity()` | `@name[/collection]` |

```toml
# url4.toml
host = "127.0.0.1"
port = 4404
default_route = "/upper"        # fan-out reduce target; defaults to the first command

[commands]                                   # the only intent-processor kind
"/upper" = ["tr", "a-z", "A-Z"]
"/model" = "python gateway.py --temp {param:temperature}"   # your own LLM backend

[data]                                       # bare relative URIs
"/rubrics/42" = "score 1-5 on clarity"
"/corpus"     = { file = "corpus.md" }       # re-read per request
"/rows"       = { command = ["./rows.sh"], media_type = "application/json" }

[holdings]                                   # `@`
default = "my working notes"
science = { file = "shelves/science.md" }

[identities.emily]                           # `@emily[/collection]`
notes = { file = "emily/notes.md" }
```

## Backends are `[commands]` only

A command is an operator-owned argv template mounted at a route path. **There is no bundled
LLM/gateway connector** — an LLM backend is the operator's own script mounted as a command. The
argv template mirrors the full `Request` surface a Python handler sees, 1:1, so a command is a
first-class backend and not a degraded one:

| Token | Substituted with |
|---|---|
| `{intent}` | the resolved intent |
| `{context}` | the resolved context (also piped to stdin) |
| `{param:<name>}` | one decoded protocol param (`/model?temperature=0.7`); `""` when absent |
| `{params}` | the full mapping as JSON, key-sorted |

Substitution is a **single regex pass over the operator's template**, so token-shaped text
arriving in caller-controlled intent, context or param values stays literal — a caller cannot
smuggle `{param:x}` into an intent and have it expand.

**Choosing `default_route` — the reduce call has a different shape.** A fan-out `(a, b)!'pick'`
reduces by calling `default_route` with the per-source results merged into its **`{intent}`, and
empty stdin** — not as context. A stdin-only backend therefore reads nothing and returns an
empty `200`, silently. This PR's own README previously paired `default_route = "/upper"` with
`["tr", "a-z", "A-Z"]` and did exactly that; found by the new CI smoke test (below), fixed in the
README and spec §5.1, and now asserted in CI.

## Read providers

All three read registries share one provider shape — an inline string, or a table with exactly
one of `value` / `file` / `command`:

- `file` is re-read **per request** in a worker thread, so operators edit the backing file with
  no restart; a missing file is a `ResolutionError` (the source's failure, like a failing
  command), not a node error.
- `command` is doctrine N4 applied to reads: no shell, argv list, empty stdin, `{collection}`
  substituted.
- `media_type` is accepted on `[data]` only, so a collection served there parses **by declared
  type rather than by sniffing** (spec §5.3.7) — a one-line JSON array served as `text/plain`
  would otherwise collapse to a single element. Holdings carry no type channel, so it is a
  config error there.

## Path-qualified `@` (spec §5.6.1/§5.6.3.1)

A bare `@` selects the `default` shelf; the eval path selects another:

```
GET /v1?q=(@)!'…'              -> default shelf
GET /v1/science?q=(@)!'…'      -> the "science" shelf
GET /v1/drafts/2026?q=(@)!'…'  -> segments join with "/"
```

An unknown qualifier falls back to `default`. The qualifier scopes the **node** context only —
an identity reference keeps its own collection, so `(@, @emily/notes)` under `/v1/science`
resolves both (§5.6.2). `{eval_path}/…` is therefore a reserved namespace: command and data
routes declared under it are rejected at startup, because endpoints match first in dispatch and
would silently shadow every qualifier below that path.

**The grammar is deliberately untouched.** §5.6.2 fixes `self-ref = "@"` — a bare `@` takes no
collection suffix and `@/science` is a parse error by design (parsing rule 12). The qualifier
rides on `ExecutionContext.self_collection`, **not** on the node: a `Url4Node` serves concurrent
requests, so a "current collection" attribute on the node would race between overlapping
evaluations. A test pins `@/science` as a `ParseError` so this cannot be "fixed" into spec
violation without a red test.

## HTTP surface

GET-only, per N1:

| Method / path | Result |
|---|---|
| `GET {eval_path}[/qualifier]?q=<expr>` | evaluates the expression (default `eval_path` `/v1`) → 200 `text/plain` |
| `GET /healthz` | liveness, no engine run → 200 `ok` |
| any other method | 405 |

Errors return `{"error": {"code": …, "message": …}}`. The **node** maps `Url4Error.code` to
status (400 parse/unbound, 404 unknown route/identity, 403 consent/identity, 502 transient
command failure, 500 otherwise); a thin raw-ASGI wrapper adds only what the node does not own —
503 over `--max-inflight` (with `Retry-After`), 504 over `--timeout`, and graceful shutdown.
A `run(concurrency=…)` run-wide I/O cap plus an in-flight admission gate bound load and cost;
each request evaluates under its own `ExecutionContext`.

## `url4 eval '<expr>'`

Network-free one-shot evaluation via `StaticIOLayer`, reading the expression from the argument
or stdin. Any expression that would fetch raises `ResolutionError` (exit 1).

## Core SDK change (breaking)

Selecting a fan-out reduce route no longer hardcodes `/claude`. `DEFAULT_PROCESSOR` is removed
from `url4.dag`; an unset `processor` resolves through a new optional port capability,
`url4.io.layer.SupportsDefaultRoute`, to the io world's first declared route (`Url4Node` → first
registered endpoint, `StaticIOLayer` → first `routes` key). With none declared, a fan-out reduce
raises a clear `ResolutionError` instead of silently targeting a route that may not exist. The
HTTP wire param `processor=` is unchanged.

## Packaging

- Console script `url4 = "url4.cli:main"`. New modules: `url4/cli.py`, `url4/_serve.py`.
- `QUICKSTART.md` is now **`README.md`**, wired via `readme = "README.md"` — the package
  previously published to PyPI with no long description at all. Verified in the built wheel's
  `METADATA` (`Description-Content-Type: text/markdown`).
- `url4 --version` and `url4 eval` work on the base install; `url4 serve` without the extra
  prints an install hint and exits 2.

## Security posture (v1)

Command routes are arbitrary local execution reachable over HTTP — an intended capability, to be
hardened in later layers. The v1 minimum:

- Binds `127.0.0.1` by default. A non-loopback bind warns loudly, and again if command routes
  are present. An empty host is rejected outright (it would bind every interface while reading
  as "unset"), and an empty `URL4_*` env var counts as unset — an unresolved `.env`/compose
  interpolation must not silently become `""`.
- Operator-controlled argv vs caller-controlled payload: no `shell=True`, argv is a list, every
  token substitutes as a single argv token and is never re-split or re-scanned.
- Read providers are operator config too: a `file` provider's path is never caller-influenced,
  so a data route cannot be turned into arbitrary file read.
- Per-command timeout; the process is killed on overrun. No secrets in the serve layer.

## Also fixed here: a silent test-suite hang

`test_serve_forever_without_uvicorn_prints_hint` and `test_serve_requires_uvicorn` asserted the
missing-`[server]`-extra branch while relying on the dev venv to *lack* uvicorn — neither test
created the condition it asserted. With the extra installed (which `README.md` instructs), the
import succeeded, the tests fell through to the real `uvicorn.run()`, bound `127.0.0.1:4404` and
served forever: **pytest hung, with no failure and no timeout.**

It had looked like a fast `errno 98` failure only because an unrelated process happened to hold
4404 — the symptom depended on unrelated system state, which is why it was misclassified as
"test fragility, CI is green" for three review rounds. CI is green because CI installs no extra.

Fixed with a `hide_uvicorn` fixture so both tests create their own precondition, plus an autouse
`_no_real_listen` guard that turns any `socket.listen` in a unit test into an immediate,
actionable failure rather than a hang. This edits two **prior** tests, so `run_gates.py` needs
`--skip-append-only`; that was an explicit Confidence-Gate decision by the owner after the hang
was demonstrated. No other prior test is touched.

## Deferred (v1 non-goals)

WS/SSE streaming and the three-signal telemetry (`cost.usage` / Enclave trace store /
`Link`-header traces); daemonization (`--detach`, PID files); the Python-plugin loader (the
`(context, intent) -> str` seam exists); log-verbosity flags; any bundled LLM connector.

## CI

`url4-tests.yml` runs two jobs, both capped at `timeout-minutes: 10`:

- **tests (base install)** — `uv sync`, no extras. Proves the library works, and that the core
  import graph stays framework-free, on the install most consumers get.
- **serve smoke (`url4[server]`)** — new. Without it the console script this package ships had
  **no CI coverage at all**. It installs the extra and (a) runs the whole suite with uvicorn
  present, which is exactly the configuration where the two tests above used to hang the runner;
  (b) runs the import-isolation test, only meaningful here — with uvicorn on the path, "url4 did
  not import it" is a real claim rather than a vacuous one; (c) starts a real node and drives it
  over HTTP: command dispatch, a `[data]` route, `@`, a path-qualified `@`, an identity shelf, a
  fan-out reduce, and the 400/404/405 contract. Bounded readiness wait, server stopped in an
  `if: always()` step with its log dumped.

That job found the `default_route` documentation bug above on its first run — while every unit
test passed and `_serve.py` sat at 100% line coverage. Unit tests drive the ASGI app with
hand-built scope/receive/send and never execute the README, so they structurally cannot catch
the docs and the code disagreeing about how to hold the thing.

`tests/unit/test_import_isolation.py` is also new. Spec §8 ticked "`import url4` never imports
Starlette/uvicorn" as an acceptance criterion and §9 listed it in the test plan, but **no such
test existed**. It runs in a subprocess — asserting against the test process's own `sys.modules`
would be worthless, since pytest and the conftest have already imported plenty — and includes a
probe self-check so a typo in the detector cannot make every assertion pass vacuously.

## Test plan

- `run_gates.py url4 --skip-append-only` — ruff check, ruff format, pyright, and
  `pytest --cov=url4 --cov-fail-under=95` all green. **812 passed**; 97% line / 96% branch
  overall, with the new modules `_serve.py` and `cli.py` at **100% line, 99% branch**.
  (`server.py` sits at 89%; the misses are pre-existing SDK paths — the real `uvicorn.run()`,
  `__aenter__`/`__aexit__`, the lazy `HttpIOLayer` import — not the serve feature. The changes
  here moved it 88% → 89%.)
- The `--skip-append-only` is required for exactly two prior tests, fixed under "silent
  test-suite hang" above; no other prior test is modified.
- Exercised over real HTTP (uvicorn), now in CI rather than only by hand: `/healthz` → 200;
  `GET /v1?q=(/upper(hello world)!'go')` → `HELLO WORLD`; a fan-out reduce delivers both
  per-source results *and* the instruction to `default_route`; a `[data]` route resolves a bare
  relative URI; a `file` provider re-reads live between two requests; `/v1/science` and
  `/v1/drafts/2026` select their shelves; `/v1/absent` falls back; `(@, @emily/notes)` resolves
  both; `@/science` → 400; unknown route → 404; non-GET → 405.
- Console script: `url4 --version`; `url4 eval` from arg and stdin; `url4 serve` without the
  extra → hint + exit 2; empty `[commands]`, `--default-route /absent`, `--host ""`, and a route
  under `{eval_path}/` all → exit 2.

## Follow-ups (not blocking)

- Coverage is gated on **line** coverage only (no `[tool.coverage]` section, so no `branch =
  true`). Package branch coverage is 96% and would pass the same bar, but it is not what the
  gate checks.
- Linear label taxonomy reconcile (the card's Epic/workstream group has been removed), and the
  card's "Linear via MCP only" rule vs. the owner-supplied-API-key precedent now used twice.

BREAKING CHANGE: `DEFAULT_PROCESSOR` is removed from `url4.dag`. A fan-out reduce with an unset
`processor` no longer defaults to `/claude`; it resolves via the new
`url4.io.layer.SupportsDefaultRoute` port to the io world's first declared route, and raises
`ResolutionError` when none is declared. Callers that relied on the implicit `/claude` must pass
`processor=` explicitly or declare the intended route first — the reduce still runs, but
dispatches to the first declared route rather than `/claude`. Additionally, `url4 serve` now
reserves `{eval_path}/…` for self-holdings qualifiers, so command and data routes declared under
the eval path are rejected at startup.

BREAKING CHANGE: url4 grammar conformance (OME-500) narrows the parser to the normative ABNF. Intent-less parenthesized groups (`(a, b)`), iteration bodies (`src*(body)`), and relative/remote calls (`/p(ctx)`) — all previously accepted — now raise `ParseError` (code `missing_intent`); add the `!intent` the grammar always required. Identifier/param/path/port character classes and `q=`-last ordering are enforced too. Data URIs (`/data/$topic`), the reduce sub-request (`?q=()!…`), and every documented `!intent` form are unaffected.


sergio-bershadsky:
Review questions before this lands — mostly around the description/ledger drift after the 2026-07-16 commands-only pivot. Verified against the branch head, not just the diff.

**1. PR description is pre-pivot.** It still documents the aigateway LLM connector, `--processor`, command routes as "opt-in / off by default", and `url4/__init__.py` + `url4/server.py` as "untouched". All four are contradicted by the code at head:

- `_serve.py` has no `httpx` / `make_llm_handler` / `backend_url` / `DEFAULT_ROUTES`, and states *"there is no other backend kind"*.
- The flag is `--default-route`; `--processor` does not exist.
- `ServeConfig.validate()` hard-fails on empty `[commands]` — commands are mandatory, so "off by default" can't be the posture.
- Both `__init__.py` and `server.py` are modified in this PR (plus `client.py`, `dag/*`, `io_layer.py`, `io_static.py`).

Rewrite the body from the ledger + spec §11 before merge?

**2. Gates — which is accurate?** The body says `run_gates.py url4` was ALL GREEN with "append-only test check ✓". The ledger in the same PR says `run_gates.py url4 --skip-append-only` and *"the append-only gate red is exactly the owner-approved list"* (6 prior tests modified). Can you point at the owner approval?

**3. `_LOOPBACK` includes `""` (`cli.py:26`).**

```python
_LOOPBACK = frozenset({"127.0.0.1", "::1", "localhost", ""})
```

`host=""` binds `0.0.0.0` **and** `::` — every interface (verified). So `URL4_HOST=` (an empty value in a `.env` / compose file) or `--host ""` binds the node publicly with command routes enabled, while `_warn_exposure` returns early and **skips both warnings** — the v1 control this PR lists for what it calls "arbitrary code execution over HTTP". Intentional, or should `""` be dropped from the set and rejected in `validate()`?

**4. `DEFAULT_PROCESSOR` removal on a released package.** `url4` v1 is already published, and this drops `DEFAULT_PROCESSOR` from `url4.dag.__all__` and changes the reduce default from `/claude` to "first declared route, else `ResolutionError`" — silent for existing callers. Deliberate breaking change for this PR, and does it need a version bump / changelog entry? (The `SupportsDefaultRoute` port is a nice fix — worth calling out in the description rather than leaving it undocumented.)

**5. Test count.** Body says 753 (704 + 49); ledger says 760. Which run is current?

PR #403, opened 2026-07-16
feat(scoreboard): wire SCOREBOARD_SUBMISSION_API_KEY as a secret-backed env var

FilipBolt:
## Summary
OME-467 (sub-issue of OME-391 Step 3). `Settings.submission_api_key` (Step 2, PR #392) is read from `SCOREBOARD_SUBMISSION_API_KEY`, but the chart had no way to set it — confirmed via a full read of every file in `charts/scoreboard/`, the only secret pattern is `database.existingSecret` for the DB URL, and there's no generic `extraEnv`/`extraEnvFrom` passthrough anywhere.

**Live finding while working this:** the gate is already active in production right now and the key Dmitry generated works correctly (verified via a safe probe — bogus `benchmark_id`, no data written: no header → `401`, with the key → `404` unknown-benchmark, proving it passed auth). But the key was set via an out-of-band `kubectl set env` — Helm revision 12's values don't own that reference, so it doesn't survive a normal `helm upgrade`, rollback, reinstall, or the planned Azure migration. This PR is what makes it durable.

- Added `submissionApiKey.existingSecret`/`existingSecretKey` to `values.yaml`, mirroring `database.existingSecret` — empty/unset by default.
- Wired into `deployment.yaml` as a `secretKeyRef` env entry, guarded so it only renders when `submissionApiKey.existingSecret` is set. No `optional: true` — Kubernetes defaults to required, so a missing/misspelled Secret or key prevents the pod from starting rather than silently reopening writes.
- `values-prod.yaml` now declares the actual live Secret (`scoreboard-submission-api-key` / `SCOREBOARD_SUBMISSION_API_KEY`) so every production deploy renders it automatically — no `--set` flag needed.
- `DEPLOYMENT.md` updated with concrete `kubectl create secret` + `helm upgrade` steps.

## Test plan
- [x] `helm lint charts/scoreboard` clean: default, `--set submissionApiKey.*`, and with `values-prod.yaml`
- [x] `helm template` unset → confirmed no `SCOREBOARD_SUBMISSION_API_KEY` env entry renders (default behavior unchanged)
- [x] `helm template --set submissionApiKey.existingSecret=... --set submissionApiKey.existingSecretKey=...` → confirmed correct `secretKeyRef` renders with the given names
- [x] `helm template --values values-prod.yaml` → confirmed it now renders the real production Secret reference declaratively, with no extra flags
- [x] Live probe against `scoreboard.screamingface.ai` confirms the gate + key already work correctly today — this PR makes that state durable across upgrades rather than changing current behavior

Refs: OME-467

### DIFF
--- a/apps/scoreboard/DEPLOYMENT.md
+++ b/apps/scoreboard/DEPLOYMENT.md
@@ -104,15 +104,17 @@ The chart never takes the raw key value directly — put it in a Secret and poin
 
 ```bash
 kubectl -n scoreboard create secret generic scoreboard-submission-api-key \
-  --from-literal=api-key='<the key value>'
+  --from-literal=SCOREBOARD_SUBMISSION_API_KEY='<the key value>'
 
 helm upgrade scoreboard oci://ghcr.io/openmined/screamingface/charts/scoreboard \
   --namespace scoreboard \
   --reuse-values \
   --set submissionApiKey.existingSecret=scoreboard-submission-api-key \
-  --set submissionApiKey.existingSecretKey=api-key \
+  --set submissionApiKey.existingSecretKey=SCOREBOARD_SUBMISSION_API_KEY \
   --wait
 ```
+
+`values-prod.yaml` already declares this exact Secret/key pair by default, so a production deploy that includes it doesn't need the `--set` flags above at all — they're shown here for the general/non-prod case.
 
 Leaving `submissionApiKey.existingSecret` unset (the default) renders no `SCOREBOARD_SUBMISSION_API_KEY` env var at all — the gate stays a no-op.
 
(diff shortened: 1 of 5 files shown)


apps/scoreboard/DEPLOYMENT.md
```diff
@@ -100,6 +100,22 @@ helm upgrade --install scoreboard oci://ghcr.io/openmined/screamingface/charts/s
 
 Set `SCOREBOARD_SUBMISSION_API_KEY` to gate `POST /v1/scores` (OME-391 / C2) — a placeholder shared key until OME-326 (real identity) exists. Unset, the write path stays open, matching today's behavior. Once set, submitting clients (desktop app, SDK, CI) need `Authorization: Bearer <key>` — coordinate the key distribution before flipping this on in production, since it isn't per-user.
 
+The chart never takes the raw key value directly — put it in a Secret and point the chart at it, mirroring the `database.existingSecret` pattern:
+
+```bash
+kubectl -n scoreboard create secret generic scoreboard-submission-api-key \
+  --from-literal=api-key='<the key value>'
```
HupBaHa: `values-prod.yaml` expects the key name `SCOREBOARD_SUBMISSION_API_KEY`, while this command currently creates `api-key`. Keeping the documented Secret shape aligned prevents a later production deploy from referencing a missing key.

```suggestion
  --from-literal=SCOREBOARD_SUBMISSION_API_KEY='<the key value>'
```

FilipBolt: Good catch, fixed in e182400 — the doc now uses SCOREBOARD_SUBMISSION_API_KEY as the data key everywhere, matching values-prod.yaml. Also added a note that the --set flags in that example aren't needed for a prod deploy since values-prod.yaml already declares this pair by default.

HupBaHa (approved):

PR #444, opened 2026-07-30
feat: adopt the Cloudflare Access identity headers (OME-684)

IonesioJunior:
Refs: OME-684 · https://linear.app/openmined/issue/OME-684

Carry the caller's Cloudflare-verified identity from the edge to the provider call.

Contract: [https://pulse.dev.openmined.org/docs/products/gateway-identity-flow/](<https://pulse.dev.openmined.org/docs/products/gateway-identity-flow/>) — Cloudflare Access authenticates the caller at the edge and issues a signed assertion; the request reaches the cluster through a `cloudflared` tunnel, where **Envoy re-verifies that assertion against Cloudflare's JWKS**, clears any client-supplied copy, and re-injects the verified claim as a plain header. Downstream services read the header and do no token work of their own.

This is what `OME-588` set out to do. The difference is **where the JWT is verified** — Envoy rather than aigateway — which is why the `cf_access/` JWKS package `OME-591` specifies is unnecessary.

## Scope — one header, five units, one branch

`X-User-Email` is the whole identity. The flow also carries a tenant and, for service tokens, a `common_name`; both are deliberately ignored. An email is globally unique, so a tenant adds nothing to a key built from it, and automation is out of scope until `OME-592` — a service-token caller gets a 401 rather than being half-identified.

**1 · url4-cloud carries the identity to the run**

Not header pass-through: the App that receives the header and the Runner Pod that calls aigateway are different processes, and the outgoing request does not exist yet when the header arrives. Identity is captured at the REST edge, serialized onto the Job spec as plain env, and re-rendered by the connector — the path `credential` and `profile` already take.

* `job_env.py` — the header/env table and its conversions. The env name joins `WRITTEN_BY_APP` but deliberately **not** `SECRET`: identity authorizes nothing on its own.
* `ports.py` — `IdentityAwareJobRunner` widens `schedule` locally, so `packages/url4` **is untouched**.
* `connector.py` — gateway-owned headers written LAST, so no inbound value displaces the run's `Authorization` / `X-Profile`.

**2 · aigateway resolves the caller**

`current_account()` is the single choke point every authenticated route depends on, so the modes branch there; profiles, `credential_blobs` and the request cache are unchanged.

* `AIGW_AUTH_MODE = jwt | cloudflare_headers | disabled`. A mode, not a second boolean — with `auth_enabled` plus a `trust_headers` flag, "false/false" would read as the safe setting while being the most permissive. Legacy `AIGATEWAY_AUTH_ENABLED` still supplies the default; a contradictory pair is a startup error.
* `core/auth/cloudflare_identity.py` — the account is keyed on `username`, which IS the verified email. `Account.username` is unique, so there is no second derivation to keep in step; the lookup is `Account.get_or_create(username=email)`, which already handles the concurrent-first-request race in a transaction.
* Migration `0008` widens `username` 64 → 255 (RFC 5321 allows 254). **Postgres-only DDL by design** — see below.
* No identity in header mode is a **401, never anonymous**: otherwise a misconfigured mesh serves every caller as one shared principal with pooled credentials.

**3 · The charts make it the default cloud shape**

* **SECURITY FIX:** the aigateway NetworkPolicy was an allow-all wearing the name of a restriction — the template omitted `from:` when `ingressCIDRs` was empty, and `values-prod.yaml` enabled it with no CIDRs. A rule with no `from:` admits every source. It now builds peers explicitly and FAILS the render rather than emitting an empty set.
* `authMode` defaults to `cloudflare_headers`; `ingress.enabled=false` (internal-only, which is what makes the header unforgeable); url4-cloud's Gateway API edge is opted into via `values-cloud.yaml`.
* Three configurations refuse to render: `cloudflare_headers` + an Ingress, `authEnabled=false` + a non-disabled mode, and an enabled policy with no peers.

**4 · The gateway refuses header identity from outside the declared networks**

Trusting `X-User-Email` rested entirely on deployment configuration — no Ingress, plus the NetworkPolicy. A NetworkPolicy restricts nothing unless the cluster's CNI enforces it, so where it does not, the mode was an open impersonation endpoint for anything that could reach the port. `AIGW_ALLOWED_NETWORKS` puts the same boundary **in the process**, where no CNI can decline it.

* The caller's TCP peer must fall inside a declared CIDR network, or the request is refused **403 before the identity header is read**.
* The peer, **never `X-Forwarded-For`**. Both headers are attacker-controlled, so checking one to decide whether to believe the other is circular — an attacker outside the allowlist would simply add `X-Forwarded-For: <a trusted pod IP>` and become any principal. A test pins this so a later "fix" for a proxied deployment cannot reopen it; such a deployment declares the proxy's own address instead.
* Parsed with `strict=True`, so a value with host bits set is refused rather than silently widened — `192.168.0.0/8` (proposed in review) would become `192.0.0.0/8` under `strict=False` and trust 16 million addresses nobody named.
* Mandatory in this mode: `create_app` raises without it and the chart fails the render — a fourth refusal. "Which networks?" has no answer the gateway can safely guess.
* IPv4-mapped IPv6 peers (`::ffff:10.1.2.3`) are normalised, so a dual-stack cluster does not fail closed.
* The chart default is the private ranges + CGNAT, so **a normal install needs no configuration**. Narrowing to the real Pod CIDR is optional hardening — precise, label-based identification of the two peers is the NetworkPolicy's job, not this check's.

**5 · Diagrams** — `docs/diagrams/gateway-identity-{flow,topology,auth-modes}` plus an index page.

## A destructive migration, caught by a test

Widening `username` with the obvious `ops.AlterField` **deletes data**. SQLite cannot widen in place, so Tortoise rebuilds the table (`CREATE TABLE new__accounts` … `DROP TABLE accounts` … `RENAME`), and `oauth_connections.account_id` references `accounts` with `ON DELETE CASCADE` — so the DROP removes **every OAuth connection in the database**. `test_migration_populated_db` caught it.

The migration now uses dialect-guarded raw SQL: ALTER on Postgres, nothing on SQLite. That is not a workaround for the cascade — the cascade exposed that SQLite never needed the DDL, since it does not enforce `VARCHAR(n)`. Both branches are unit-tested.

## Verified

`run_gates.py` green on both apps (aigateway 1320 passed, url4-cloud 488 passed), `helm lint` clean on both charts with every guard's failure path exercised. End to end: a chat request carrying `X-User-Email` passes auth and reaches credential resolution, returning `404 profile_not_found` rather than `401`.

## Out of scope / follow-ups

* **Nothing provisions credentials for a header-derived principal** — a new caller authenticates and then gets `404 profile_not_found`. Per-user OAuth vs operator-managed org keys is a separate decision.
* **Service-token callers have no path** until `OME-592`.
* **The NetworkPolicy is still unverified against a real CNI.** Acceptance test unchanged: from a Pod in another namespace, `curl -H 'X-User-Email: …' http://<aigw-svc>:9105/v1/chat/completions` must TIME OUT. If it answers, the CNI is not enforcing the policy — but the request now gets **403** from `AIGW_ALLOWED_NETWORKS` unless that Pod is inside a declared network. Unit 4 lowers the consequence from silent cluster-wide impersonation to a refusal; it does not answer the question.
* The mesh gateway itself (Envoy GatewayClass, listeners, the JWT filter) is cluster infrastructure neither chart installs.
* aigateway's public host is gone, so the credential UI and OAuth callbacks need routing through the mesh gateway.
* `reauth_url` (`chat_credentials.py:161`) points at OAuth routes a header-mode caller cannot use.

## Commits, in reviewable order

| Commit | What |
| --- | --- |
| `3c8cdf0f` | url4-cloud carries the identity to the run |
| `2c0a5c86` | aigateway resolves the caller from the header |
| `160fcdd8` | charts make it the default cloud shape — **includes the NetworkPolicy security fix** |
| `75f51554` | three diagrams + an index page |
| `13955593` | ledgers retargeted at OME-684 |
| `dead9dc4` | CI fix: keep the chart's own defaults schema-valid |
| `50449d3a` | **PR review**: rename to Cloudflare, reduce to one header, `get_or_create`, username key |
| `0ed3b99a` | **PR review**: refuse header identity from outside `AIGW_ALLOWED_NETWORKS` |

`84416ae1` (removal of aigateway's own Cloudflare JWT forwarding) pre-dates this work but sits on the same branch. It is redundant *because Envoy now does that verification*, not because Cloudflare left the picture — an earlier version of this PR described it wrongly and the description above is the correction.

## Breaking changes

- **aigateway:** `AIGW_AUTH_MODE=gateway_headers` is now `cloudflare_headers`, and in that mode a caller presenting anything other than `X-User-Email` is unauthenticated. `settings.auth_enabled` is no longer read at request time; `auth_mode` resolves at construction. The env→Settings path is unchanged, so a deployment setting only `AIGATEWAY_AUTH_ENABLED` is unaffected.
- **aigateway:** `AIGW_AUTH_MODE=cloudflare_headers` now requires `AIGW_ALLOWED_NETWORKS`. The app refuses to start and the chart refuses to render without it. The chart ships a working default, so a chart-based deploy is unaffected.
- **charts:** aigateway loses its public host (`gateway.screamingface.ai`).
- **charts:** url4-cloud's Gateway API edge is opt-in via `-f values-cloud.yaml` with `gateway.parentRef.name` set; a bare install renders no edge.


HupBaHa (commented):
Thank you for putting the end-to-end identity path together. I have four questions about the remaining trust, execution, and persistence boundaries.

apps/url4-cloud/src/url4_cloud/runner/main.py
```diff
@@ -114,6 +114,7 @@ async def _world() -> World:
             ),
             token=token,
             profile=env.get(job_env.AIGATEWAY_PROFILE),
+            identity_headers=job_env.identity_from_env(env),
```
HupBaHa, line 115: I may be missing an external deployment component, but the current identity-only path appears to stop before this value is used: URL4 schedules the identity without `AIGATEWAY_TOKEN`, while `_world()` raises `RunnerConfigError` when that token is absent. I could not find a chart or job-factory path that injects it. Is another component expected to supply the token? If not, a legitimate Cloudflare-authenticated request seems unable to reach AIGateway without also sending the legacy Bearer value.

apps/aigateway/src/aigateway/migrations/0008_widen_account_username.py
```diff
@@ -0,0 +1,64 @@
...
+    Left as the honest inverse rather than a silent truncation: a rollback that quietly rewrote
+    people's usernames would detach them from their stored credentials.
+    """
+    await _alter(schema_editor, 64)
+
+
+class Migration(migrations.Migration):
+    dependencies = [("models", "0007_connection_auth_type")]
+
+    operations = [ops.RunPython(forwards, backwards)]
```
HupBaHa: The physical migration itself looks safe, but `RunPython` does not update Tortoise’s projected migration state. The model declares `max_length=255` while migration state remains at 64, so normal autodetection proposes another `AlterField(Account.username)`. On SQLite that operation rebuilds `accounts`; with `oauth_connections.account_id` using `ON DELETE CASCADE`, applying the generated alteration removes the related OAuth rows. Could this use a state-aware operation that records 255 while keeping physical DDL PostgreSQL-only, plus a regression asserting that autodetection no longer proposes the field alteration?

sergio-bershadsky: @IonesioJunior Yes please use AlterField instead of RunPython

PR #466, opened 2026-08-04
feat(scoreboard): authenticate leaderboard submissions via mesh identity header

FilipBolt:
## Summary

Replaces the temporary shared-key stub on `POST /v1/scores` (`SCOREBOARD_SUBMISSION_API_KEY`, OME-391/C2 — its own `AIDEV-NOTE` said to delete it once OME-326 shipped real identity, which it did today) with the same pattern `aigateway` already runs live: trust the `X-User-Email` identity header Envoy injects after re-verifying Cloudflare Access, behind a peer-network trust boundary. No JWT/JWKS verification lives in scoreboard — that stays Envoy's job.

Defaults to `disabled` (existing free-text `submitted_by` behavior unchanged) until `SCOREBOARD_AUTH_MODE=cloudflare_headers` is explicitly set, so no existing deployment or test breaks from this change.

Full design rationale in `docs/spec/2026-08-03-OME-404-authenticated-leaderboard-submissions.md` and `docs/plan/2026-08-03-OME-404-authenticated-leaderboard-submissions.md`; full round-by-round audit trail in `docs/work/2026-08-03-OME-404-authenticated-leaderboard-submissions.md`.

## Hardened through 4 rounds of self-review before opening this PR

- Moved identity resolution ahead of all business-rule validation in `submit_score` — the deleted `Depends()`-based gate guaranteed that ordering for free; a plain function call did not, until fixed.
- **Found and closed a severe bypass**: the chart's pre-existing `forwardedAllowIps="*"` (needed for Traefik's HTTPS-redirect on the existing production deployment) lets uvicorn trust a client-supplied `X-Forwarded-For` from *any* peer, silently overwriting `request.client.host` before the new `peer_in_networks()` check ever runs. `create_app` now refuses to start with `cloudflare_headers` mode combined with `FORWARDED_ALLOW_IPS=*`. Verified by live exploitation against the real entrypoint, before and after the fix.
- Added tests pinning uvicorn's own internal trust logic directly (not just this guard's `ValueError`), so a future uvicorn upgrade that changes it fails a test instead of silently reopening the bypass.
- Added regression tests proving identity resolution wins over both the accuracy-tolerance check and the benchmark-existence check — the exact combined scenario the ordering bug lived in, previously untested.

## Deliberately not done, and why

- The NetworkPolicy chart-level fail-guard originally planned (mirroring aigateway's) would break `release-scoreboard.yml` CI without a real production CIDR value to fill in, and the file it would apply to describes the current Traefik-fronted deployment, not the Cloudflare Access/Envoy mesh this auth mode targets. The app-level `allowed_networks` + `FORWARDED_ALLOW_IPS` guard already provide equivalent protection independent of that chart change.
- Left as a known, documented duplication rather than extracted to `packages/`: `peer_in_networks` and the header constant are copied from aigateway, not shared — a real `packages/` extraction is a bigger structural commitment than this unit's scope.

## Test plan

- [x] `uv run ruff check` / `ruff format --check` / `pyright` / `pytest --cov=scoreboard --cov-fail-under=80` all green (157 passed, 2 pre-existing unrelated skips, 88.65% coverage)
- [x] Live-reproduced the `FORWARDED_ALLOW_IPS="*"` bypass against the real entrypoint before the fix, and confirmed the guard blocks it after
- [ ] Owner-verify: the peer-network trust boundary can't be fully verified from unit tests alone — confirm in the actual dev-cloud deployment that `POST /v1/scores` is unreachable except through the Envoy/Cloudflare Access chain before enabling `cloudflare_headers` mode there

Refs: OME-404


FilipBolt:
Pushed a fix for the FORWARDED_ALLOW_IPS/allowed_networks overlap gap (external review feedback, thanks for the catch): scoping `FORWARDED_ALLOW_IPS` away from `"*"` wasn't sufficient on its own — uvicorn's `ProxyHeadersMiddleware` still rewrites `request.client.host` from `X-Forwarded-For` whenever the real peer falls inside *any* `FORWARDED_ALLOW_IPS` entry, even a single narrow address. If that address also overlaps `allowed_networks`, the exact peers the check exists to authenticate are the ones it can no longer see correctly.

Fixed in `a90e78bd`:
- `main.py`'s startup guard now rejects any overlap between `FORWARDED_ALLOW_IPS` and `allowed_networks`, not just the literal `"*"` case.
- Fixed a mislabeled test in `test_allowed_networks.py` that was actually exercising the overlap bug while claiming to be the safe/scoped case, plus 5 new tests + 2 new uvicorn-internals pinning tests.
- New `test_proxy_headers_integration.py` — the first test in the suite wiring a *real* `ProxyHeadersMiddleware` around `create_app()`, proving the fix holds end-to-end rather than just in the guard's own unit tests.
- `DEPLOYMENT.md`'s example now sets `forwardedAllowIps` explicitly and retracts the old advice to scope it to the real proxy's address — that was exactly the vulnerable configuration.

Gates green under both Python 3.12 and 3.13. Details in the ledger's "External review round (2026-08-06)" section.

HupBaHa (approved):

PR #488, opened 2026-08-04
feat(url4): add a ModelResponse observation event and its ctx-less sink

sergio-bershadsky:
Closes OME-744 · sub-issue of OME-679 · unblocks OME-745

## Why

`url4.observe` had no seam for **response metadata**. A url4 endpoint's contract is `-> str`, so a model call's `finish_reason` and the provider `refusal` field had nowhere to travel and were dropped at the world adapter — exactly where OME-679 loses the signal a researcher needs to tell a **provider refusal** from a **bad answer**. (Fable 5 refuses many medical questions; with HealthBench as the entry benchmark, a safety-refusing model is currently scored as if it answered badly.)

This adds the seam only. Consuming it is OME-745 (`apps/url4-cloud`).

## What

- `ModelResponse(span_id, finish_reason, refusal)` joins the `ObservationEvent` union.
- `ExecutionContext.report_response()` + `current_response_sink()` — the ctx-less lookup a world adapter uses, mirroring the existing `Usage` / `current_usage_sink()` pair.
- Both sinks are bound by one `_bind_node_sinks` context manager.

**Why not just a field on `Usage`:** `Usage` is token accounting, and embedders derive cost frames from the span's usage tuple — a finish reason there would push a non-cost fact into cost accounting. `ObservationEvent` is a union whose module docstring frames the event set as the extension point, so a new member is the sanctioned move.

**Where the helper lives:** written inline in `_eval` first, which tripped the `too-many-statements` gate (27 > 26). Rather than weaken the gate it was extracted — and then moved out of `executor.py` entirely, because the ContextVars are `observe.py`'s private state and the executor should not reach into them to scope a binding it does not own. It takes the two bound methods rather than an `ExecutionContext`, so `observe.py` stays the dependency-free leaf its docstring promises. Net: `executor.py` **shrank** 496 → 492 lines and no longer imports either private ContextVar.

## Test plan

`uv run .claude/scripts/run_gates.py url4` — **ALL GATES GREEN**: append-only check · ruff · ruff format --check · pyright (no `# type: ignore` added) · `pytest --cov=url4 --cov-fail-under=95`. **1100 passed**, coverage **97%**; every new line covered.

9 new tests in `packages/url4/tests/unit/test_response_sink.py`, no prior test touched:

- ctx-less sink → a `ModelResponse` on that node's span; `ctx.report_response` lands the same event
- a `content_filter` turn carries both the reason and the refusal text
- **no cross-talk** between concurrent siblings (the property the ContextVar-per-Task binding exists to protect)
- several reports from one node stay a sequence — a tool loop's `tool_calls` then `stop` are both auditable
- binding reset on **both** the success and the exception path; no observer → sink is `None`
- an observer that raises propagates (embedder bugs stay loud)

**Sibling stack verified:** this widens a public union that `apps/url4-cloud` consumes, so its suite was run too — **478 passed, 5 skipped**. Safe because no consumer matches the union exhaustively (`_Bridge.map` is an `isinstance` chain documented as "Any other event type produces nothing"); an `assert_never` grep across both packages is clean.

## Note for the reviewer

`packages/url4/uv.lock` carries one unrelated line: `pyproject.toml:107` pins `graphviz>=0.21` but the committed lock recorded `>=0.20` — the lock was stale on `main` and `uv run` regenerated it deterministically. Kept rather than reverted, since CI runs plain `uv sync` (not `--frozen`) so the drift was silent, and reverting only means it returns on the next `uv run`.

Ledger: `docs/work/2026-08-04-OME-744-model-response-event.md`

### DIFF
--- a/docs/tasks/2026-08-04-ome-744-model-response-event.md
+++ b/docs/tasks/2026-08-04-ome-744-model-response-event.md
@@ -1,0 +1,45 @@
+---
+id: OME-744
+linear_url: https://linear.app/openmined/issue/OME-744/add-a-modelresponse-observation-event-so-a-world-adapter-can-report
+status: in_progress
+type:
+priority: P2
+labels: [url4-python-sdk, autonomous, agentic]
+created: 2026-08-04
+closed:
+---
+
+# OME-744 — a ModelResponse observation event for finish_reason and refusal
+
+Sub-issue of `OME-679` (finish reason and refusal retrieved in the model response).
+
+`url4.observe` is the engine's observation seam: "a passive, pure-data event stream" carrying
+`RunStarted | NodeStarted | NodeFinished | Log | Usage | RunFinished`, plus a contextvar-bound
+`UsageSink` so a world adapter with no `ExecutionContext` of its own can still attribute a fact
+to the currently-resolving node's span.
+
+There is **no equivalent seam for response metadata**. A url4 endpoint's contract is `-> str`,
+so a model call's `finish_reason` and the provider `refusal` field have nowhere to travel and
+are discarded at the adapter — which is exactly where `OME-679` says the signal is lost.
+
+## Why a new event, not a field on `Usage`
+
+`Usage` is token accounting, and url4-cloud's executor derives `CostUsageData` from the span's
+usage tuple. Widening that tuple would push a non-cost fact into cost accounting and change what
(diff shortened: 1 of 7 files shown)


IonesioJunior (approved):
LGTM!

PR #507, opened 2026-08-05
feat(aigateway): add global exact-request cache

HupBaHa:
## Description
Implements a default-on global exact-request response cache for eligible non-streaming AIGateway chat completions.

- Builds an identity-independent SHA-256 fingerprint from the effective output-affecting request and pure provider-owned projections.
- Merges profile defaults body-wins before cache planning while keeping auth mode and provider credentials off the cache-hit path.
- Supports deterministic cache participation for Anthropic and OpenRouter; unsupported or unsafe inputs fail closed to bypass.
- Reuses `RequestCacheEntry` for global v2 rows with nullable expiry and create-only first-writer-wins persistence.
- Stores the complete provider-compatible response as compact JSON and replays it unchanged, including original usage.
- Makes absent cache control default to read/write and treats `use-cache=false` as a complete no-read/no-write bypass.
- Publishes separate hit/miss/bypass reason and write-outcome headers.
- Keeps cache storage and metadata failures off the availability path so provider dispatch continues normally.
- Adds migration, runtime configuration, production chart wiring, deployment documentation, and comprehensive unit/PostgreSQL coverage.
- Hardens canonicalization, cache-hit metadata handling, provider participation, cache-reason vocabulary, parameter validation, and log safety based on independent review findings.

Fixes OME-305

## Affected Dependencies
No new runtime or development dependencies.

## How has this been tested?
- The complete AIGateway quality gate passed: lint, formatting, type checking, Enterprise import guard, pytest, and coverage.
- The final focused cache/provider regression slice passed with 336 tests.
- PostgreSQL-specific migration, create-only conflict, and atomic hit-metadata coverage passed with 12 tests.
- Helm lint, chart wiring validation, and production-values rendering passed.
- One unrelated login timing assertion flaked once, passed alone unchanged, and the complete gate then passed unchanged.

Reproduce from the repository root:

```bash
uv run .claude/scripts/run_gates.py aigateway --skip-append-only
```

Reproduce the PostgreSQL-specific checks from `apps/aigateway` with a test PostgreSQL database configured:

```bash
AIGW_TEST_PG=1 uv run pytest -m needs_postgres
```

## Checklist
- [x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [x] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [x] My changes are covered by tests


sergio-bershadsky:
Two standards suggestions from the url4 side, where we're specifying how to consume this cache ([#515](https://github.com/OpenMined/screamingface/pull/515)). Both are cheap **while this header code is open** and much more expensive later — entirely your call, and neither blocks anything.

## 1 · `Cache-Status` (RFC 9211) alongside the `X-AIGW-Cache*` set

This PR introduces `CACHE_HEADER` / `REASON_HEADER` / `WRITE_HEADER` / `KEY_HEADER` and adds a fourth field. RFC 9211 (Standards Track, June 2022) defines **one** header for exactly this, and the mapping is near-1:1:

| here | RFC 9211 |
| --- | --- |
| `X-AIGW-Cache: hit` | `hit` (§2.1) |
| `X-AIGW-Cache: miss` / `bypass` | `fwd=miss` / `fwd=bypass` — **both are defined tokens** (§2.2) |
| `X-AIGW-Cache-Write` | `stored` (§2.5) |
| `X-AIGW-Cache-Key` | `key` (§2.7) |
| `X-AIGW-Cache-Reason` | `detail` (§2.8) |

```
Cache-Status: aigateway; fwd=bypass; detail=opted_out
Cache-Status: aigateway; hit; key="a3f1c09e4b2d"
```

The closed bypass vocabulary this PR introduces independently arrived at `bypass` and `miss` — two of the RFC's own `fwd` tokens — which reads as good evidence the modelling is right and only the serialisation is bespoke.

The structural argument matters more than tidiness: `Cache-Status` is an RFC 8941 Structured Field **List**, one member per cache that handled the response, origin-closest first. A response here can pass this cache *and* Envoy *and* potentially a CDN. Four flat headers give one slot, so the second cache to speak either overwrites this one's answer or invents its own field — the road to `X-Cache`, `CF-Cache-Status` and `X-AIGW-Cache` disagreeing on one response. The list form composes by construction.

Also worth noting RFC 6648 / **BCP 178** deprecates the `X-` prefix for new fields; `URL4-Capability` already follows the modern style, so the codebase is currently inconsistent with itself.

## 2 · `Age` (RFC 9111 §5.1) — this one buys a feature, not just tidiness

`values-prod.yaml` documents that rows never expire. I initially read that as "a caller can't ask for bounded staleness" — but that conflates *never expiring* with *unknown age*.

`RequestCacheEntry` already carries `created_at`, so an entry's age is computable exactly. If the response carried `Age` (or `Cache-Status`'s `ttl` parameter, §2.4), then a caller's `Cache-Control: max-age=3600` becomes **honourable** rather than unsatisfiable: serve the stored answer if younger, forward otherwise.

For a permanent global corpus that's the difference between all-or-nothing and a caller-controlled staleness bound — "reuse yesterday's answer, not last month's" — using only standard fields. It's the single most useful thing url4 could ask of this cache, and right now we have to spec around its absence (our D11).

## What we're doing regardless

url4's read-back will **prefer `Cache-Status` and fall back to the `X-AIGW-*` triple**, so we need no change whichever way you go.

Happy to open a follow-up rather than expand this PR — it's already +12k. Mostly flagging it now because the marginal cost is near zero while these constants are being written, and a migration later isn't.

PR #518, opened 2026-08-06
feat(url4): per-run cache policy for the aigateway global response cache

sergio-bershadsky:
Implements the design approved in **#515** (spec r8 / plan r6). A caller can declare that one url4 run must **not** participate in aigateway's global response cache; caching is otherwise on, because aigateway v2 makes it on.

```
Cache-Control: no-store              (HTTP, on GET /)
{"cache": {"participate": false}}    (WS attach frame)
```

Depends on **#507** (`OME-305`) for the upstream half. `apps/aigateway` is untouched.

## The constraint everything is shaped by

v2's control grammar is **closed to exactly one field**. An unrecognised key does not degrade to "ignored" — `global_controls.py:79-83` makes the whole request **bypass**, even alongside a valid `use-cache: true`. A well-meant `no-store` or `ttl` would silently cost every cache hit, with nothing raised anywhere.

So url4 collapses every directive to participate/opt-out **at its own edge** and never forwards a name v2 retired. Guarded three ways, the load-bearing one being a property test:

```python
set(out.get("cache", {})) <= {"use-cache"}   # for every input
```

That is a correctness test, not a style one — the failure mode it prevents is *absence of an effect*, which a green suite would otherwise hide.

## What ships

| batch | |
| --- | --- |
| 1 | `CachePolicy` + `AttachData.cache` + `SpanData.cache_status/reason` in `packages/url4` |
| 2 | the aigateway mapping — the only place in url4 that knows the string `use-cache` |
| 3 | `Cache-Control` on `GET /`, mirroring `X-Profile` |
| 4 | WS attach ingress, first-attach-wins |
| 5 | convergence — header wins, override logged |
| 6 | per-run threading (**not** on `AigatewayConfig`) + egress |
| 7 | read-back preferring RFC 9211 `Cache-Status` |
| 8 | counters by bypass reason, OpenAPI/AsyncAPI docs |

**Read-back matters more than it looks.** Without it a cache hit bills as a fresh call — an error that *hides savings*, so nobody reports it. `Cache-Status` is parsed as an RFC 8941 List selecting the **aigateway member**, not the first: the list may legitimately carry Envoy or CDN entries, and reading a CDN's hit as ours would report a saving that never happened.

## `max-age` ships inert, deliberately

url4 can neither ask for a freshness bound (closed grammar) nor measure one (no `Age` header). So a stated bound **degrades to opt-out**, behind a single `GATEWAY_REPORTS_AGE = False`. The honouring path — `requires_revalidation` and its tests — is written, correct and unreachable; flipping that one flag is all of D11's second half.

This changed during review. The first implementation did *participate-then-revalidate*, which with no `Age` meant **two gateway calls on every bounded hit** — inside `for _ in range(cfg.web_tool_max_iterations)`, so a four-iteration tool turn became eight calls, multiplied across a fan-out. And since D6 deliberately invites intermediaries to inject `Cache-Control`, `max-age=0` from a browser reload was an accidental trigger. Owner ruled for opt-out.

## Gates

```
url4        ✓ append-only ✓ ruff ✓ format ✓ pyright ✓ pytest --cov  (97.45%, 1100 passed)
url4-cloud  ✓ ruff ✓ format ✓ pyright ✓ check_layering ✓ pytest --cov (>80%)
```

**One accepted exception.** Widening the `IdentityAwareJobRunner` port meant two test **doubles** had to accept the new parameter (+10 lines, imports plus one optional arg). The append-only gate stopped the run — correctly, since rule 5 makes that a Confidence-Gate decision — and the owner accepted it as port conformance: no assertion altered, nothing deleted. The implementing agent explicitly declined to widen `ScheduledRun` because an existing test compares that tuple whole. Full diff quoted in the ledger.

## How it was built, and what that caught

Eight sequential agents (one per batch, each running its stack's gates, stop-on-red), then a ninth reviewing adversarially against the spec's acceptance criteria. It returned 10 findings; two were substantive — the `max-age` cost above, and a spec/code disagreement about whether `max-age` equals `no-store` on a miss (it doesn't, per RFC 9111 — `max-age` bounds *reading*, `no-store` prohibits *storing*; my spec conflated them). The other eight were doc staleness, fixed in spec r8 on #515.

Agents also overruled the plan twice, both correctly: it was internally inconsistent about whether `CachePolicy` has one field or two, and it named a test location that contradicts `packages/url4/pyproject.toml`'s own documented convention.

## Not verified

**Green gates are not evidence this works end to end.** Every test asserts against a mock transport, so the aigateway half of the contract is verified by *reading* `global_controls.py`, not by executing it. Plan Verification step 3 needs a local gateway built from #507 and has not been run.

## Process deviation

**No Linear issue** — MCP unauthenticated; the card names it the only permitted transport. This also breaches **rule 8**: the work spans `packages/url4` + `apps/url4-cloud` and should be an epic with one sub-issue per landing. Executed as one `UNFILED` unit because the epic cannot be created. Branch name, `docs/tasks/` mirror and `Refs:` all need back-filling.

https://claude.ai/code/session_01HAiqcmwueUjWBhsEfsQWGx

### DIFF
--- a/apps/url4-cloud/README.md
+++ b/apps/url4-cloud/README.md
@@ -147,3 +147,60 @@ caller's identity and aigateway decides, including whether an absent identity is
   `config.aigatewayBaseUrl`. Unset ⇒ the endpoint answers `503`; everything else is a code default
   (see the `models_cache_*` fields in `config.py`).
 - Cache behaviour is observable at `/metrics` (`url4_cloud_catalog_*`).
+
+## Per-run cache policy
+
+A different cache from the one above: aigateway's **response** cache, which answers an identical
+model call from a stored corpus instead of dispatching it to the provider again. Design:
+`docs/spec/2026-08-05-url4-cache-policy-spec.md`.
+
+**Caching is ON by default.** Only declining is explicit — the gateway participates unless told
+not to, so a switch that "turns caching off" is the only switch there is. Two carriers, one
+meaning, scoped to the **whole run** (every leaf, every fan-out branch — per-node intent would
+need url4 grammar and is out of scope):
+
+```sh
+# HTTP — the standard RFC 9111 request field on GET /
+curl -H 'URL4-Capability: <jwt>' -H 'Cache-Control: no-store' 'http://localhost:9108/?q=...'
+```
+
+```json
+{"type": "ai.url4.attach", "data": {"from_sequence": 1, "cache": {"participate": false}}}
+```
+
+| directive | effect |
+| --- | --- |
+| *(absent)* | participate — the default |
+| `no-store`, `no-cache` | do not participate |
+| `max-age=<seconds>` | participate under a freshness bound (see **Known-inert** below) |
+| `url4-use-cache` | participate, explicitly — the token that lets the header override a frame opt-out |
+
+Conflicting directives resolve to **not** participating: the worst case of declining is a missed
+hit, while the worst case of participating against a caller who refused is a shared answer they
+explicitly declined. Unknown and malformed directives are **ignored, never 4xx** — a cache
+directive is a hint about cost, not a term of the request.
+
+**Precedence: the header wins**, and the overridden declaration is announced as a `warn`
+`ai.url4.log` on the stream. **First attach wins** on the frame side: a re-attach with a different
+policy leaves the run's policy alone (calls may already have run under it) and warns.
+
+> **INVARIANT — url4 never sends a cache control key other than `use-cache`.** aigateway's cache
+> grammar is CLOSED to that one field, and any other key inside the request body's `cache` object
+> makes the whole request **bypass** the cache — silently, with nothing raised anywhere, even
+> alongside a valid `use-cache: true`. So every directive above collapses to participate/opt-out
+> at url4's own edge (`rest/cache_header.py` → intent, `runner/cache.py` → the wire), and a run
+> that participates sends **no `cache` field at all**. `tests/unit/test_runner_cache_body_field.py`
+> pins it as a property over every input.
+
+**Observability.** Each span carries `cache_status` (`hit`/`miss`/`bypass`) and `cache_reason` —
+the gateway's vocabulary verbatim, so `opted_out` stays distinct from `unsupported_control`. The
+run publishes one summary `ai.url4.log` with its hit, miss and **bypass-by-reason** totals
+(`runner/cache_counters.py`). Not Prometheus: a run is a one-shot Job with no scrape endpoint, and
+the layering gate keeps `url4_cloud.runner.*` out of `url4_cloud.metrics` anyway. **No counter is
+labelled by cache key, prompt or credential** — the gateway's entry key is parsed and stops at
+`runner/cache_readback.py`.
+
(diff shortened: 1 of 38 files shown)


sergio-bershadsky:
## End-to-end verification — run 2026-08-07, all checks passed

The ledger flagged that the aigateway half of this contract was established by **reading** `global_controls.py`, never by executing it — *"green gates are not evidence this works end to end."* That was true until #507 landed. It has now been run.

**What makes it an E2E rather than a restatement of the same belief:** the request bodies are not hand-written. They come out of url4-cloud's own production chain — `parse_cache_control()` → `resolve()` → `policy_to_body_field()` — and go over real HTTP to a real aigateway process running merged `main` (`4f2a55ea`).

| # | check | result |
|---|---|---|
| 1 | default body clears the control gate (`provider_projection`, not a caller-attributed reason) | ✅ |
| 2 | `Cache-Control: no-store` → `bypass` / **`opted_out`** | ✅ |
| 2 | `Cache-Control: no-cache` → `bypass` / **`opted_out`** | ✅ |
| 3 | `max-age=60` and `max-age=0` → `bypass` / `opted_out` (D11 degrade) | ✅ |
| 4 | **negative control** — raw `{"no-store":true}` / `{"ttl":60}` / `{"use-cache":true,"s-maxage":30}` → `unsupported_control` | ✅ |
| 5 | a default run's egress body carries no `cache` key (spec §9.1) | ✅ |

**Check 4 is what makes check 2 mean anything.** If the gateway answered `opted_out` for everything, check 2 would pass for the wrong reason. Sending the v1 vocabulary url4 deliberately no longer emits, and getting a *different* reason back, proves the distinction is real.

### Recorded honestly: what was NOT proven

**`default → miss → hit` was not driven through url4** — unreachable locally, not skipped for convenience:

- ollama inherits `CacheBypass(provider_projection)` from `ProviderPluginBase` and can never hit (its `parameters.py:83-90` says so and names the conformance test enforcing it)
- **anthropic and openrouter are the only providers implementing `global_cache_projection`**, and both pin `api_base` by design (openrouter D7). Neither is stubbable; a real miss needs a paid credential.

url4's default egress body is byte-identical to a pre-#518 body (check 5), so miss→hit is aigateway's own contract on a body url4 does not touch — covered by its suite, verified passing on merged `main` alongside this run: `test_repeated_bare_request_still_hits` (a **bare** request repeated — the exact shape url4 sends), the cross-account and cross-profile hit tests, and `test_chat_global_cache_route.py` (25 passed).

### Ordering fact worth recording

`build_global_cache_plan` evaluates `cache_enabled` → `participates_in_global_cache` → **`controls.participate`** → projection. url4's opt-out is adjudicated at the third step, which is why the *same* provider returns `opted_out` for an opt-out and `provider_projection` for a default. Two different reasons from one provider is the evidence the control gate saw participation — the E2E leans on that rather than taking a header at face value.

`GATEWAY_REPORTS_AGE = False` re-verified on `main`: `git grep -E '"Age"|Cache-Status' -- 'apps/aigateway/src/**/*.py'` returns nothing. D11's honouring half stays correctly dormant.

Also rebased onto `main` (now carrying #515, the spec/plan re-synced to merged #507). Both stacks' gates green locally.

PR #520, opened 2026-08-06
feat(aigateway): expose model, provider, and search contracts

keelancj:
## Summary

- register the OpenRouter model seeds required by the current benchmark and notebook flows
- expose enabled credential providers from the loaded plugin registry
- accept bounded provider-neutral web-search intent and apply Gateway-owned OpenRouter policy
- reject the unadvertised `:online` variant so it cannot bypass deployment domain exclusions

## Architecture

The caller supplies `web_search` and optional excluded domains, never OpenRouter's extensible plugin envelope. The OpenRouter adapter owns the private wire representation and unions caller exclusions with deployment policy.

This deliberately preserves the existing Engine routing decision: verified provider-side retrieval remains provider-side, while routes that cannot enforce DRACO's blocklist continue to use the Runner-owned Tavily loop. This PR does not change benchmark grading or retrieval routing.

## Verification

- `pytest -q`: 2671 passed, 40 skipped
- OpenRouter provider suite: 743 passed
- `pyright`: 0 errors
- Ruff check and format verification passed

Refs: OME-712

### DIFF
--- a/apps/aigateway/src/aigateway/core/plugin_base/_provider.py
+++ b/apps/aigateway/src/aigateway/core/plugin_base/_provider.py
@@ -49,6 +49,7 @@ class ProviderPluginCore[TSettings: PluginSettings](ABC):
     """
 
     custom_llm_provider: str
+    provider_display_name: str
     settings_cls: ClassVar[type[PluginSettings]] = PluginSettings
 
     def __init__(self, settings: TSettings | None = None) -> None:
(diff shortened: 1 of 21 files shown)


HupBaHa (approved):
Reviewed exact head 9f0e8e84 against current main, including the OME-305 global-cache interaction. The :online path now bypasses lookup before dispatch validation, and the adapter revision invalidates pre-fix OpenRouter entries. Focused cache/search/provider tests and AIGateway static gates pass; no PR-scoped blocker found. Hosted Python CI should still run before merge.

PR #523, opened 2026-08-06
feat(benchmarks): HealthBench worst-30% challenge benchmark

khoaguin:
Engine-owned HealthBench worst-30% challenge exam + the researcher-facing notebook, verified end to end against the local stack (smoke score 1.0 with full judge evidence; a limit=2 worst30 Fusion rehearsal scored -1.627 with verdict_coverage 1.0).

### Before
- Trigger: an open-source researcher wants to prove their Fusion beats our baseline on hard health conversations.
- Today: there is no HealthBench exam on the Engine at all — no cases served, no judge, no score; the only HealthBench numbers we have came from a one-off July script nobody can rerun.
- Cost: the entry challenge of the launch narrative simply doesn't exist yet, and any number we quote is unreproducible.

### After
- Same trigger.
- Now: `sf.evaluate(candidate, benchmark="healthbench/worst30")` runs the 157 hardest HealthBench Professional conversations through the Engine — Candidate answers, a pinned GPT-5.4 judge grades every physician-written rubric item, penalties subtract, and the report carries the full per-item evidence trail (judge raw reply, MET/UNMET, points). A one-Case `healthbench/smoke` proves the whole chain for pennies, and `limit=N` rehearses cheaply before a full run.
- Win: a reproducible, auditable challenge number — negative scores rank (unclipped mean by design), so the leaderboard can distinguish -0.35 from -0.15 where the official clip would flatten everything to 0.

### Don't regress
- A Case that cannot be scored fails VISIBLY and forces `score: null` — a dropped hard Case or a defaulted penalty verdict can only cheat in the submitter's favor, and both are structurally excluded.
- The grader prompt is byte-for-byte OpenAI simple-evals (sha-pinned in tests) — including its typos.
- The rubric answer key never leaves the image: clients see Case ids + conversations only.
- Every result decodes with the REAL SDK decoder in unit tests (new cross-package round-trip guard) — the four contract drifts that each cost a live paid run to find now fail in milliseconds.
- Malformed judge replies retry with a FRESH sample (nested-call retry), bounded at 2 — the reference's own recovery loop, not an infinite one.

## Implementation notes
- Scoring is the challenge metric (UNCLIPPED mean), deliberately NOT an official HealthBench score — every description says so.
- `aggregate:<n>` intent (protocol v2) tells the reducer which sliced prefix a `limit=N` run selected.
- Judge model `openrouter/openai/gpt-5.4` registered in all three synced lists (url4.toml, gateway seeds, seed-pin test) — it was missing everywhere, which was the root cause behind every live smoke failure during bring-up.
- `just stack-up` now syncs venvs + migrates the gateway DB; `stack-prepare` bakes healthbench assets — a fresh worktree boots with no manual steps.
- Known issue (separate ticket, must land before the OME-762 baseline rerun): the SDK's synthesis compilation leaks `synthesis_1_member_N_name` lines into Fusion output text, which the judge then grades.

Refs: OME-760, OME-761


khoaguin:
moved implementation to https://github.com/OpenMined/screamingface/pull/544

PR #527, opened 2026-08-06
feat(public-docs): six SF Client user guides against the live SDK

callezenwaka:
Second sub-issue of `OME-666`. Adds six of the parent's ten user-guide pages;
`OME-669` covers the remaining four.

## What's here

```
USER GUIDES
  Connections
  ▸ Compose
      Models
      Fusions
  Benchmarks
  Running an evaluation
  Reproduce & share (URL4)
```

Every API name is checked against `OME-605-screamingface-client-v1` @ `e387aefd`,
and every displayed output comes from a real run against a local engine —
discovery for the first four pages, an `ifeval` evaluation for the last two.
No figure is invented; cells with no captured output show code only.

## The SDK moved

These document a different API than `OME-666` describes, because the client moved
off the abandoned `OME-400` branch: `sf.evaluate()` rather than
`benchmark.evaluate()`, one `Report` rather than `Report` + `StudyReport`, Fusion
synthesizers rather than reducers, `sf.configure()` rather than `sf.config()`, and
no `draco-lite`.

## Beyond the AC

- **Navigation is modelled as groups and links.** `Compose` is a label with no page,
  which the old tree could not express — every entry had to be a link. Now a group
  labels children and a link points at a page, rendered by one recursive `NavTree`.
- **Mobile navigation.** Below 1024px there was no sidebar and the navbar's product
  row collapsed to plain text, so a phone reader could only move by prev/next. The
  sidebar is now off-canvas below `lg` and the product label is a switcher.
- **The commit stamp renders once in the sidebar footer**, not at the foot of each
  guide as the ticket asks. It becomes a semver label when the package ships to PyPI.
- **The Overview lost "Where next"**, which duplicated the sidebar and still said the
  guides were coming.

## Test plan

`public-docs` has no test suite. Gates: `oxlint` and `eslint` clean run bare as CI
does, `build` succeeds, `prettier --check` clean on every touched file. Verified in
light and dark theme, and at 400px / 800px / 1200px.

Refs: OME-668

### DIFF
--- a/docs/tasks/2026-08-06-user-guides-v1.md
+++ b/docs/tasks/2026-08-06-user-guides-v1.md
@@ -1,0 +1,46 @@
+---
+id: OME-668
+linear_url: https://linear.app/openmined/issue/OME-668
+parent: OME-666
+status: In Progress
+type: Task
+priority: P2
+labels: [repo, autonomous, agentic, task]
+created: 2026-08-06
+closed:
+---
+
+# User guides v1 for connections, compose (models, fusions), benchmarks (first only), running an evaluation
+
+Second of six sub-issues under `OME-666` (Documentation for ScreamingFace Client V1). Adds six of
+the parent's ten user-guide pages; `OME-669` covers the remaining four.
+
+Pages, in sidebar order:
+
+```
+USER GUIDES
+  Connections
+  ▸ Compose
+      Models
+      Fusions
+  Benchmarks
+  Running an evaluation
+  Reproduce & share (URL4)
+```
+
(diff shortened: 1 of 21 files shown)


callezenwaka:
Screenshot 2026-08-06 at 21 37 10
Screenshot 2026-08-06 at 21 36 20
Screenshot 2026-08-06 at 21 36 27
Screenshot 2026-08-06 at 21 36 34
Screenshot 2026-08-06 at 21 36 44
Screenshot 2026-08-06 at 21 36 52
Screenshot 2026-08-06 at 21 37 04

PR #528, opened 2026-08-07
feat(benchmarks): IFEval

khoaguin:
Two IFEval stories, both now provably faithful to their papers: `ifeval` is the 2023 Zhou et al. protocol on the official dataset with a parity-proven grader, and `ifeval/lanl-ensemble` is the Skurikhin et al. §2 early-exit ensemble replacing the OpenMined `verifying-ensemble` variant. `ifeval/self-corrective` stays as the solo corrective baseline.

### Before
- Trigger: a researcher publishes an IFEval score or runs the corrective ensemble.
- Today: case ids are 1..n renumberings that don't join to the official dataset; the grader is asserted (not proven) equivalent to the community fork; the pinned HF snapshot silently carries one row whose prompt contradicts its own graded constraints; and the ensemble burns 3N model calls + a judge call per round even when round 1 already passed, under a name that implies paper parity it doesn't have.
- Cost: scores that can't be cross-checked against third-party IFEval results; ~3x token spend for identical ensemble scores; a benchmark menu that confuses "our variant" with "the paper".

### After
- Same trigger
- Now: case ids ARE the official IFEval keys in official file order; a golden-parity test runs our grader and the vendored fork's verdict core side by side over all 541 official rows and demands identical verdict vectors; key 2785 is patched to the official text (any other drift fails the build loudly); and the ensemble is the paper's flow — a round-1 pass costs N member calls and nothing else, the judge speaks only on ties or failures, never-pass falls back to Best-of-N.
- Win: publishable, joinable, cheaper — and the leaderboard entry names the protocol it actually runs.

### Don't regress
- Canonical scoring formulas unchanged (prompt/inst × strict/loose); revision hashes bump exactly where assets or protocol changed.
- `ifeval/self-corrective` untouched; shared member machinery (resolve-candidate, member-record/answer) retained under the LANL route prefix.
- url4 compiler: collection-position references now resolve like body references (`$gate*(body)` gating idiom); existing expressions are unaffected — the previous behavior was a hard runtime error, not a meaning anyone could rely on.

## Known limits (documented in-module)
- The LANL E2E invocation test is `xfail(strict=True)` on **OME-765**: a url4-core decode bug (mid-list iterate source in a paren-stripped map-row body) blocks the attempt-2 continuation at run time. Diagnosis + drafted fix are on the ticket; it ships in its own url4 PR. Until then `ifeval/lanl-ensemble` scores `None` on real runs.
- Non-ASCII Roulette: 2 of 541 official rows (keys 1122, 1129) draw random constraint parameters per check — official-harness behavior; the LANL gate and grading record draw independently on those rows (≤0.37% prompt-level jitter).

## Compat note for @keelanev
`ifeval/verifying-ensemble` disappears from the flat registry; member-machinery routes move under `/benchmarks/ifeval/lanl-ensemble/<revision>/`.

Gates: apps/url4-cloud unit suite 966 passed / 5 skipped / 1 xfailed; packages/url4 suite 1127 passed; golden parity 541×both graders live. Canonical `ifeval` validated end-to-end against a live stack (limit=1 smoke, official key 1000, deterministic per-instruction evidence).

Refs: OME-718, OME-764


khoaguin:
moved implementation to https://github.com/OpenMined/screamingface/pull/543

PR #539, opened 2026-08-08
feat(screamingface): add Python evaluation client

keelancj:
## Summary

- add the first `packages/screamingface` Python Client SDK and its public `sf.Model`, `sf.Fusion`, discovery, connection, evaluation, and report APIs
- compile user-authored candidates into URL4 while keeping benchmark selection, cases, grading, and aggregation Engine-owned
- support Cloudflare Access login, provider connections, model-parameter discovery and preflight, REST/WebSocket run execution, replay, cancellation, progress events, finish reasons, refusals, typed failures, usage, and portable report artifacts
- add the ScreamingFace-branded connection, evaluation, and report widgets plus generated Quickstart, Client Tour, DRACO Lite, and DRACO notebooks
- register the new package with repository ownership, CI, release metadata, documentation, and deterministic distribution checks

## Architectural boundary

This PR adds no benchmark protocol or grading logic to the Client and changes no AI Gateway or URL4 Cloud application code. The Client links Candidates into Engine-owned benchmark expressions through the public Engine contracts.

The hosted Engine defaults to `https://fusion.dev.screamingface.ai`; local development can use `sf.configure(engine_url="http://127.0.0.1:9108")` or `SCREAMINGFACE_ENGINE_URL`.

## Verification

- `561 passed, 1 skipped`
- coverage: `95.09%` with the `95%` gate enforced
- Ruff lint and formatting pass
- Pyright passes with zero errors
- generated notebooks are fresh and output-free
- wheel/sdist distribution checks pass
- repository layering check passes

## Deployment note

Authenticated discovery and benchmark-resource requests against `fusion.dev.screamingface.ai` succeed. Hosted run creation currently returns an origin-side HTTP 500 even for a zero-cost static URL4 expression; that is isolated to the deployed Runner Job launch boundary and is independent of this Client change. Local execution remains covered by the vertical-slice and transport suites.

## Tracking

[OME-605](https://linear.app/openmined/issue/OME-605/implement-the-screamingface-python-client-v1)


IonesioJunior (commented):
## Pre-merge review — 4 blocking findings

Reviewed the full branch (140 files, ~29k lines) across the public API, engine/transport, evaluation core, UI, tests, and packaging. **The overall quality is high** — enforced immutability, an executable public-symbol allowlist, a network-free 0.07s import, prompt injection closed by delegating to the certified URL4 renderer, no credential or Access token persisted anywhere, 561 tests with zero `unittest.mock` against a real RFC 6455 protocol server, and honest coverage (no `omit` patterns, 4 justified pragmas).

The four findings below are the ones I'd block on. Each either spends money incorrectly or discards work the user has already paid for. Details are in the inline comments; this is the summary.

| # | Finding | File |
|---|---|---|
| 1 | Access challenge detection ignores status → a success response can trigger a login + request replay | `_engine/auth.py` |
| 2 | An unsequenced advisory `ai.url4.log` aborts a paid Run | `_engine/contract.py` |
| 3 | Linker cannot see `$candidate` in source position → can emit an unresolved reference into a paid Run | `_evaluation/linking.py` |
| 4 | No async `cancel_active` → async interruption orphans paid Runs | `_core/ports.py` |

### Verified while reviewing

- `uv run pytest -q` → **561 passed, 1 skipped** (the ledger says 533 — stale).
- Coverage **95.09%** statements against the 95 floor; **93%** with `--cov-branch` (branch coverage is not configured, so the headline is softer than it reads, and there is ~4 lines of headroom).
- ruff + pyright clean; wheel/sdist build and the distribution check passes.
- Installed wheel resolves **url4 1.4.0** from PyPI while CI only ever tests the in-repo 1.3.0. Suite passes against 1.4.0 too, so this is latent, not broken.

### Not blocking, but worth a look before release

- `DEFAULT_ENGINE_URL` points at `fusion.dev.screamingface.ai`.
- No `LICENSE` file in the package despite `license = "Apache-2.0"` — the published wheel would ship out of compliance (`packages/url4` has one).
- The release-please manifest pins `0.1.0` as already-released, so the first release would be 0.2.0 and skip the 0.1.0 that the README, ledger, and checks all name.
- `report_view.py:402` uses glibc-only `%-d` in the unguarded `Report._repr_html_` path — every Windows notebook user gets a traceback instead of a report.
- The governing spec/plan (`2026-07-29-OME-605-direct-evaluation`) are not on this branch; the only OME-605 artifacts here declare themselves superseded by a file that isn't present. They exist at `6b070ef3` on `origin/OME-605-screamingface-client-v1` — the client-only branch reconstruction dropped the `docs/` half.

Happy to open a follow-up issue per finding, or to implement any of these on request.

packages/screamingface/src/screamingface/_engine/auth.py
```diff
@@ -0,0 +1,442 @@
...
+        self._access_http.close()
+
+    def sync_auth_flow(
+        self,
+        request: httpx.Request,
+    ) -> Generator[httpx.Request, httpx.Response, None]:
+        request.read()
+        generation = self._authorize_request(request)
+        response = yield request
+        audience = _access_audience(response)
```
IonesioJunior, line 193: ### 🔴 Critical 1 — an Access *success* can be read as a login challenge, replaying a non-idempotent request

`_access_audience` never looks at `response.status_code`. It returns an audience whenever `cf-access-aud` is present, or a `Location` header carries exactly one `kid`. So this line can report a challenge for a response that is not one — and the flow then performs an interactive browser login and blindly re-sends the request at line 198.

The Run start is `GET /?q=<url4>` with `Prefer: respond-async`, and its **success** is a 202 that is *required* to carry a `Location` header (`transport.py:381`). Reproduced:

```python
>>> r = httpx.Response(202, headers={"cf-access-aud": "a"*64})
>>> _access_audience(r)
'aaaa…'   # a 202 success, reported as a login challenge
```

**Severity correction, in fairness:** I checked the server side before filing this. `url4-cloud`'s `_schedule` (`rest/routes.py:145-179`) is single-shot per topic and answers a duplicate start with 409, so this is **not** a live double-spend today — a replay fails the Run confusingly rather than billing it twice. I'd still block on it: a client library shouldn't depend on another service refusing duplicates for its own safety property, and spec §7.1 explicitly forbids blindly repeating an ambiguous start.

**Also note the WebSocket sibling already does this correctly** — `_is_access_websocket_rejection` (`transport.py:433-440`) gates on `status_code in {302, 401, 403}` before applying the same heuristic. The three call sites disagree, and the tests only ever exercise 302, so nothing catches it.

#### Suggested fix

One predicate, three call sites. Take `(status_code, headers)` rather than a response object, since the WS caller has a `websockets` response, not an `httpx` one — both satisfy `Mapping[str, str]`:

```python
_ACCESS_CHALLENGE_STATUSES = frozenset({302, 401, 403})

def _challenge_audience(status_code: int, headers: Mapping[str, str]) -> str | None:
    """The AUD tag from a CHALLENGE response, or None if this is not one."""
    if status_code not in _ACCESS_CHALLENGE_STATUSES:
        return None
    ...  # existing extraction + _VALID_AUDIENCE check
```

I'd keep the set at `{302, 401, 403}` rather than widening to the whole 3xx range — Cloudflare Access doesn't challenge with 301/303/307/308, and widening adds false-positive surface to a change whose entire purpose is narrowing.

**Worth considering (larger, your call):** even with the gate, "log in then re-send" is unsafe for a request that starts paid work, and the safety property ends up resting on Cloudflare's choice of status codes. A per-request, default-deny replay marker via `httpx` `request.extensions` makes it structural — mark `POST /token` and `DELETE /` replay-safe, leave `GET /?q=` unmarked, and have an unmarked request raise a retryable `AuthenticationError("reissue the request")` after login instead of replaying. Cost is near zero in practice: `run()` calls `_mint_sync` first and *that* is replay-safe, so cold-start login happens there before the start ever goes out.

**Adjacent, same bug class:** `_start_sync`/`_start_async` loop over all seven `_ATTACH_RETRY_DELAYS` and only `break` on `_attachment_is_still_registering`. Any *other* non-202 runs the loop to exhaustion, re-issuing the same non-idempotent start seven times. Currently masked by the 409 guard. Worth a `break` in the same change.

packages/screamingface/src/screamingface/_engine/contract.py
```diff
@@ -0,0 +1,499 @@
...
+        self._started_at: datetime | None = None
+        self._result: tuple[str, str | None] | None = None
+        self._root_usage: AccountingUsage | None = None
+        self._last_sequence = 0
+        self._event_ids: set[str] = set()
+        self._consecutive_replay_requests = 0
+        self._stream_reattach_requests = 0
+
+    def accept(self, raw: str | bytes) -> _Accepted:
+        payload = _payload(raw)
```
IonesioJunior, line 62: ### 🔴 Critical 2 — an unsequenced advisory log aborts a paid Run

This line routes only `ai.url4.heartbeat` and `ai.url4.error` to the unsequenced path. Everything else falls through to `_accept_sequence` → `_sequence`, which hard-requires `sequencetype == "Integer"` and a positive digit-string `sequence` (lines 205-211).

But `url4-cloud` — in this same repo — injects advisory notices as **`ai.url4.log`** with no sequence at all. `notices.warn` (`url4_cloud/notices.py:38`) builds a `LogEvent`; `CloudEvent.sequence`/`sequencetype` default to `None` (`url4/streaming/protocol/envelope.py:30-31`) and only `codec.decode(payload, sequence)` stamps them. Both emitters bypass the sequencer: `ws/bridge.py:238` (re-attach cache-policy warning) and `rest/routes.py:283` (cache-control override) go through `_offer`/`notify`, not `_pump`.

Running the real emitter through the real encoder gives, verbatim:

```json
{"specversion":"1.0","id":"af0725e1…","type":"ai.url4.log",
 "sequence":null,"sequencetype":null,
 "data":{"severity_text":"WARN","body":"…","attributes":{…}}}
```

Note the keys are **present and null**, not omitted — a hand-written fixture would likely get that wrong. Today this frame raises `ExecutionError: SF Engine CloudEvent sequence must declare Integer semantics` and kills the Run. The REST notice is queued *before* the run is scheduled and the client attaches before `_start_sync`, so it can be the very first frame the client ever sees.

#### Suggested fix

Discriminate on the presence of a sequence, not on a hardcoded type set:

```python
_UNSEQUENCED_TYPES = frozenset({"ai.url4.heartbeat", "ai.url4.error"})
# Types the control plane may inject out-of-band, outside the broker's sequencer.
# INVARIANT: only frames that mutate no _RunState may appear here — gap detection,
# event-id dedup, and the replay-order guarantee all live on the sequenced path.
_ADVISORY_TYPES = frozenset({"ai.url4.log"})

if event_type in _UNSEQUENCED_TYPES or (
    event_type in _ADVISORY_TYPES and payload.get("sequence") is None
):
    return self._accept_unsequenced(payload, event_type)
```

Three things worth calling out about that shape:

1. **Discriminate on `sequence` alone, not on both fields.** This is load-bearing: `test_engine_contract.py::test_state_rejects_invalid_envelopes` pops `sequencetype` from a *sequenced* `started` frame and requires it to still raise. A both-fields check would silently reclassify that frame as advisory and break the existing test.
2. **Don't generalise past the allowlist.** An unsequenced `terminated` would build a `_RunOutcome` from a stream never checked for gaps; an unsequenced `cost.usage` would corrupt `_root_usage` and therefore the billing figures in the user's Report. `heartbeat`/`error`/`log` are exactly the frames that mutate no state — that's the definition, not a coincidence.
3. **Please log it rather than dropping it.** It can't become a public `Event` (`events.Event.sequence` is a mandatory positive int, and inventing one would pollute replay order), but silently discarding it defeats the reason the server sends it — `notices.py`'s own docstring says *"a dropped declaration nobody is told about is indistinguishable from a bug in the gateway."* `_LOG.warning` is enough, and there's precedent at `contract.py:301`. Same argument applies to `ai.url4.error`, which is validated and discarded today (line 79-81) — that's swallowing server nacks like `"stop not supported: no job runner configured"`.

**Longer term (server-side, separate work):** discriminating on the *absence* of a field is a tolerant-reader shim — "out-of-band by design" and "the broker dropped the sequence" look identical on the wire. A distinct `ai.url4.notice` type would be the durable fix.

packages/screamingface/src/screamingface/_evaluation/linking.py
```diff
@@ -0,0 +1,225 @@
...
+            permanent=True,
+        )
+    return [src(text(render(build(expression))), name="candidate_synthesizer", weight=0.0)]
+
+
+def _text_references(value: object) -> set[str]:
+    """Collect exact URL4 Text references from the parsed Benchmark tree."""
+
+    references: set[str] = set()
+    if isinstance(value, Text):
```
IonesioJunior, line 238: ### 🔴 Critical 3 — the linker is blind to `$candidate` references in source position

`_text_references` collects references from `Text` nodes and raw `str` fields. But URL4 parses a reference in **source position** into a `VarRef`, where the `$` sigil is stripped and the name is a bare string this walker can never match:

```python
>>> build("($candidate)!'x'")
Expression(sources=(VarRef(name='candidate', path=()),), intent=Text(value='x'))
>>> link_candidate("/a/b!'q'", "($candidate)!'x'")
PlanningError: Benchmark URL4 does not invoke the Candidate
```

(`VarRef` *is* a dataclass, so the generic recursion does reach it — it just visits `name == 'candidate'`, a plain string with no `$`, and finds nothing.)

**The total miss is the benign case.** The dangerous one is a partial miss. Reproduced against the real linker:

```python
benchmark = "($candidate_member_2, y='use $candidate_member_1')!'go'"
link_candidate(None, benchmark, members)   # 1 member
# → _LinkedCandidate(url4="(candidate_member_1:0.0:'…', ($candidate_member_2, …))!''",
#                    member_indices=(1,))
```

No exception. Arity check satisfied. The emitted expression ships an **unresolved `$candidate_member_2`** to the Engine, which fails only once the paid Run is underway.

A related symptom: because `_static_member_bindings:117` derives its message from the count it *found*, a 2-member Fusion with one member referenced in source position reports *"Benchmark requires exactly 1 direct Model members, but this Candidate has 2"* — which blames the user's Fusion for the walker's blindness. Worth checking whether anyone has reported that.

#### Suggested fix

Add a `VarRef` branch feeding the **existing** matcher (don't hand-roll a second one, or the `$candidate_result` false-positive guard will drift):

```python
elif isinstance(value, VarRef):
    # A reference SITE is a VarRef or an embedded "$name" in Text; a binding SITE
    # (Binding.name) is a bare identifier the parser cannot prefix with "$", which
    # is why the regex ignores it and this branch has to add the sigil back.
    references.update(_candidate_references(f"${value.name}"))
    references.update(_text_references(value.path))
```

Verified this is correct for every VarRef shape: `$candidate_member_1.answers[0].text` → `path=('answers', 0, 'text')` (field selectors on the resolved value, never references — must not be joined into the name); `$1` positional → no match; `$candidate_result` → still correctly ignored by the trailing lookahead.

Two more gaps in the same walker while you're in here:

- **No `Mapping` branch.** `Source.weight: float | dict | None` and `Source.budgets` can hold a `dict`, which is not a `str`, not a `tuple | list`, and not a dataclass — the walk stops dead.
- **`$$` escapes read as references.** `_candidate_references("$$candidate")` returns `{'$candidate'}`, but `$$` is the escape for a literal `$` and `_template_literal` (line 174) relies on it to keep member display names literal. A `(?<!\$)` prefix on the pattern fixes it.

#### Strongest single change here

After rendering, assert no `$candidate*` reference survives unbound — between `linked = expr(...)` and the `_LinkedCandidate` construction:

```python
# INVARIANT: every $candidate* reference in the shipped artifact resolves to a binding
# emitted above. A reference the AST walk failed to see would otherwise reach the Engine
# and fail mid-run, after the Candidate has already been paid for.
unresolved = _candidate_references(rendered) - bound
if unresolved:
    raise PlanningError(...)
```

`_requirements` reasons over the parsed tree, where a reference has several shapes and the node set can change upstream; `render` collapses them all into one surface form. So this check is representation-independent — it closes the whole bug class rather than this one instance, and turns a mid-Run failure into a plan-time one. (Depends on the `(?<!\$)` fix above, otherwise escaped names in the `candidate_members` struct can false-positive.)

packages/screamingface/src/screamingface/_core/ports.py
```diff
@@ -0,0 +1,57 @@
...
+        candidate: Candidate,
+        on_event: SyncEventObserver | None,
+    ) -> _RunOutcome: ...
+
+    def cancel_active(self) -> None: ...
+
+    def close(self) -> None: ...
+
+
+class AsyncRunTransport(Protocol):
```
IonesioJunior, line 45: ### 🔴 Critical 4 — async interruption orphans paid Runs

`SyncRunTransport` declares `cancel_active()` (line 40); `AsyncRunTransport` does not. `AsyncUrl4CloudTransport` correspondingly keeps no registry of minted capabilities and has no `DELETE /` fallback — there is no async equivalent of `_active_tokens` / `_stop_sync` anywhere.

So on async interruption the only stop attempt is the in-band `ai.url4.stop` frame sent from inside `_run_connected` (`transport.py:234`) — which is dispatched from a task that is *already being cancelled*, and cannot work at all if the socket is what failed. `_run_candidates_async` (`runner.py:217-223`) only cancels tasks; it never calls anything transport-level.

Measured against the fake Engine:

| scenario | minted | DELETEd |
|---|---|---|
| sync, 2 candidates, SIGINT | 2 | **2** ✅ |
| async, 2 candidates, `task.cancel()` | 2 | **0** ❌ |
| sync, 1 candidate, SIGINT | 1 | **0** ❌ |
| async, 1 candidate, `task.cancel()` | 1 | **0** ❌ |

The async path *does* deliver the in-band WS stop, so the gap is precisely the HTTP fallback. Spec §7.1 says the async mode "gives both Clients identical … cancellation semantics", so this is a plan deviation, and the concrete consequence is a paid Run that keeps billing after Ctrl-C.

#### Suggested fix

Mirror the sync design: `_active_tokens` on `AsyncUrl4CloudTransport`, an `_stop_async`, `async def cancel_active(self) -> None: ...` on this Protocol, and a call from `_run_candidates_async`'s interruption handler.

**Two things that are easy to get wrong here** (I got the first one wrong on my first pass):

1. **Ordering is load-bearing — sweep BEFORE cancelling the sibling tasks.** Each Run discards its own token in a `finally` on the way out, so if you cancel the tasks and `gather` them first, the registry is empty by the time you read it and the fallback is a guaranteed no-op. The sync path gets this right by accident of structure — it calls `cancel_active()` first, while the sibling worker threads are still inside `run()` with their tokens registered, and only then cancels the futures. The async version needs the same order.
2. **No `asyncio.Lock` on the token set, and please don't add one.** One transport instance is driven by one event loop, `httpx.AsyncClient` is loop-bound after first use, and there is no `await` between any read and write of the set — so it's already atomic. A lock would *introduce* suspension points into a region that currently has none. Worth a one-line docstring note, since the asymmetry with the sync transport's `threading.Lock` (which is genuinely required, because a thread pool drives it) will otherwise look like an oversight.

Also: give the stop request its own short timeout rather than inheriting the 30s client timeout (otherwise Ctrl-C can block for 30s per orphaned capability, and users hit Ctrl-C again), never swallow `CancelledError` while stopping, and treat 404/409/410 on the `DELETE` as success — with the in-band stop frequently winning the race, the fallback will routinely arrive at an already-finished Run and shouldn't report that as an error.

#### Two scope notes

- **This Protocol change turns the pyright gate red until four existing async test doubles get a no-op stub** — `_AsyncFakeTransport` (`test_draco_vertical_slice.py:224`, plus `_AsyncConcurrentFakeTransport` which inherits it), `_AsyncForbiddenTransport` and `_AsyncReachedTransport` (`test_model_parameter_preflight.py:130,141`). Purely additive; their sync twins already carry exactly that stub.
- **The `len(candidates) == 1` short-circuit bypasses the handler in *both* runners** (`runner.py:174-176` and `206-208`), so single-candidate evaluations orphan their capability in the sync path too — see the table above. Worth fixing in the same change.

Separately, and not a blocker: `cancel_active()` stops every Run owned by the transport, so two concurrent `evaluate()` calls on one Client will stop each other's paid Runs. Relocating the registry so each Evaluation owns its own would fix that, the single-candidate gap, and a mint-window race in one go — but it's a bigger change to both ports and both runners, so probably its own PR.

packages/screamingface/src/screamingface/_engine/auth.py
IonesioJunior: **Fixed in `9bdc728a`.**

The gate landed as described: one `_challenge_audience(status_code, headers)` predicate behind all three call sites, `{302, 401, 403}` only. Plus the default-deny replay marker — `POST /token`, `DELETE /`, catalogue reads, benchmark reads and connection calls declare themselves repeatable; `GET /?q=` does not, and a genuine challenge on it now raises a retryable `access_reauthenticated` after the login rather than re-sending.

Two corrections to what I wrote above, both found while implementing:

**1. My claim about `_start_sync` was wrong.** I said the loop "runs to exhaustion, re-issuing the start request seven times" on a non-202. It does not — `if not _attachment_is_still_registering(response): break` breaks on *anything* that is not the attachment-registration problem it retries, including every other non-202. There is no seven-fold replay. No change was made, and nobody should go "fix" it. Apologies for the noise.

**2. The marker had to cover more call sites than I first scoped.** Marking only mint and stop turned every catalogue read into an error on an expired session — three existing tests caught it immediately. The constant also lives in `_core/wire.py` rather than the Access module, because importing it from there would drag native crypto into `client.py`'s deliberately network-free 0.07s import path.

The three existing login-and-retry tests in `test_authentication.py` now declare their catalogue read replay-safe, which is what `Client._http_get` does in production. Their subject and assertions are unchanged.

packages/screamingface/src/screamingface/_engine/contract.py
IonesioJunior: **Fixed in `e832fd16`.**

Discriminates on the presence of `sequence` alone, as discussed — `ai.url4.log` joins heartbeat and error on the unsequenced path only when it carries no sequence, and lifecycle frames still require theirs. The fake reproduces the real wire shape (sequence keys present and null, since url4's encoder does not pass `exclude_none`).

**One thing changed during the rebase, and it is worth flagging here.** This branch had moved on 36 commits while I was working, and `eda5b722 fix(screamingface): surface event stream failures` had already landed in the meantime. It implements the `stream_failed` re-attach — which was a separate finding of mine, so that one is resolved upstream and needs nothing further.

It also rewrote `_advisory_error` to return `(code, message)` so the caller can act on the code. My version of that same function added a `_LOG.warning`. On the owner's instruction to prefer the remote fix on any overlap, **`_advisory_error` is now the remote's version verbatim** — the log line I had added to it is gone. Only genuinely new code survived the resolution: the advisory-log routing and `_advisory_log` itself, which have no counterpart upstream.

Net effect on the point in this thread: the notice is still validated, still kept out of the ordered Event stream, and still logged (via `_advisory_log`). An `ai.url4.error` is no longer logged, but it is now *acted on* where it matters, which is strictly better.

packages/screamingface/src/screamingface/_evaluation/linking.py
IonesioJunior: **Fixed in `9c45f549`.**

All four parts landed: the `VarRef` branch feeding the existing matcher, recursion into mappings, the `(?<!\$)` lookbehind so an escaped `$$candidate` is a literal, and the post-render assertion that every `$candidate*` reference resolves to an emitted binding.

The silent-poison case is now pinned by a test that fails on the old code with no exception at all — it asserted a linked artifact carrying an unresolved `$candidate_member_2`, which is the failure that only shows up mid-Run after payment.

One incidental note: `_text_references` had to be split into a leaf read and a child walk to stay inside the repo's complexity budget once the new branches were added. Behaviour is unchanged.

packages/screamingface/src/screamingface/_core/ports.py
IonesioJunior: **Fixed in `5131788b`.**

Implemented as the small symmetric fix rather than the scope redesign, by decision.

**The ordering advice in my comment above was necessary but not sufficient, and the reason is worth recording.** I implemented sweep-before-cancel as described and it still deleted nothing. `asyncio.gather` cancels its children *and waits for them to finish unwinding* before re-raising in the parent — so by the time any interruption handler runs, every Run has already executed its own cleanup and deregistered. Instrumenting the sweep showed it plainly: two capabilities minted, **zero** visible.

So ordering alone cannot fix the async case. A cancelled Run now deliberately leaves its capability registered, and the sweep owns clearing it; every other exit retires its own, which keeps the registry bounded. The synchronous twin needs none of this and is untouched — its sibling worker threads are still mid-Run when the sweep reads the registry, which is why that path worked all along.

Also landed: no lock on the registry (one loop, no `await` between read and write, so a lock would add the suspension points it is meant to guard), a 5s stop timeout so Ctrl-C cannot block for 30s per capability, and 404/409/410 treated as stopped since the in-band stop usually wins the race.

Test doubles: the three named above, plus `_AsyncReplayTransport` which arrived in the newer commits and inherited a *synchronous* `cancel_active`. All additive, and amended into this commit so no commit on the branch fails the type gate.

**Still open, deferred by decision** — worth tracking separately: the registry lives on the transport, so two concurrent Evaluations sharing a Client stop each other's Runs; and the `len(candidates) == 1` short circuit bypasses the handler in **both** runners, so single-Candidate Evaluations orphan their capability in the synchronous path too. That second one is measured, not inferred.

IonesioJunior:
## All four Criticals are fixed and pushed

Follow-up to [the review above](https://github.com/OpenMined/screamingface/pull/539#pullrequestreview-4901288332). Each finding has a reply in its own thread; this is the short version.

| # | Finding | Commit |
|---|---|---|
| 1 | Access challenge ignored status → could replay a paid Run start | `9bdc728a` |
| 2 | Unsequenced advisory `ai.url4.log` aborted a paid Run | `e832fd16` |
| 3 | Linker blind to `$candidate` in source position | `9c45f549` |
| 4 | No async `cancel_active` → async interruption orphaned paid Runs | `5131788b` |

Spec and work ledger in `f847e36d` / `f6aec0cc`. Every fix is driven by a test that fails first for the stated reason; for the two end-to-end notice tests I stashed the fix to confirm they genuinely fail without it. `run_gates.py screamingface` is green on every commit: **605 passed, 1 skipped** (from 561), coverage 95.04% against the 95% floor.

### Three things that changed the review's own conclusions

**One reported defect was not real.** I claimed `_start_sync` re-issues the Run start up to seven times on a non-202. It does not — the loop breaks on anything that is not the attachment-registration problem it retries. No change made; corrected in that thread.

**One finding was already fixed upstream.** The branch had advanced 36 commits while this review was being written, and `eda5b722` had already landed the `stream_failed` re-attach. Where its changes overlapped mine in `_advisory_error`, the remote's version was kept verbatim per the owner's instruction; only genuinely new code survived.

**The async fix needed a different shape than the review proposed.** Sweeping before cancelling — the ordering I recommended — still deleted nothing, because `asyncio.gather` waits for its children to unwind before re-raising, so the registry is already empty. Detail in that thread.

### Deliberately not addressed here

The Important and Minor findings remain open, including the ones with the shortest fuse before a public release: the missing `LICENSE` file in an Apache-2.0 package, the release-please manifest that would skip 0.1.0, `DEFAULT_ENGINE_URL` pointing at `fusion.dev`, and the glibc-only `%-d` that tracebacks on Windows notebooks. Two follow-ups from Critical 4 also need tracking — the cross-Evaluation stop blast radius, and the single-Candidate short circuit that orphans a capability in **both** runners.

Happy to take any of these next.

IonesioJunior (approved):

PR #540, opened 2026-08-08
feat(public-docs): write the Installation page

callezenwaka:
## Summary

Fourth sub-issue of `OME-666`. `InstallationPage.vue` was an 18-line stub while the
sidebar has listed it under Get Started since `OME-667`, so nothing on the site
said how to get a working environment.

Two paths, hosted first: install the client and point it at an engine someone else
runs, or run AI Gateway and the Engine yourself.

**Asana:** n/a — technical work does not go to Asana (CLAUDE.md rule 6).

## Components touched

`public-docs` only. Triggers `public-docs-tests.yml`.

## Every command was executed first

- `uv build` produces `screamingface-0.2.0-py3-none-any.whl`
- it installs into a clean Python 3.12 and imports 36 public names, resolving
  `url4` from PyPI
- the `[notebook]` extra resolves `ipywidgets 8.1.8`
- the one-line git install works and pins `e387aefd`, the commit the sidebar stamps

The self-hosted section documents what actually blocked bringing a stack up:
migrations before first start, `AIGW_AUTH_MODE`, the OpenRouter plugin shipping
disabled, and preparing benchmark assets. Those four plus "do I even need my own
engine" are an FAQ of collapsibles, following the `syft-space-hub-docs` pattern.

## Notes

- Adds a `Note` component for the inline callout; the source docs repeat that
  markup per use rather than componentising it.
- Restructured from the parent's spec, which describes Installation as local-first
  with `sf.config(engine=…)` and port `4404` — both wrong against the current client.
- Em-dashes removed from this page at the owner's request, so it reads slightly
  differently from its neighbours.
- `CodeBlock` renders dark in both themes. Left alone: it is shared with the Learn
  pages, and terminal chrome staying dark is defensible.

## Test plan

- [x] `npx oxlint .` and `npx eslint .` clean, run bare as CI does
- [x] `npm run build` succeeds
- [x] `npx prettier --check` clean on the touched files
- [x] Light and dark theme
- [ ] Screenshots attached

`public-docs` has no test suite.

## Cross-service notes

None.

Refs: OME-672

### DIFF
--- a/docs/tasks/2026-08-08-installation-tutorial.md
+++ b/docs/tasks/2026-08-08-installation-tutorial.md
@@ -1,0 +1,41 @@
+---
+id: OME-672
+linear_url: https://linear.app/openmined/issue/OME-672
+parent: OME-666
+status: In Progress
+type: Task
+priority: P2
+labels: [repo, autonomous, agentic, task]
+created: 2026-08-08
+closed:
+---
+
+# Add installation tutorial
+
+Fourth of six sub-issues under `OME-666` (Documentation for ScreamingFace Client V1). Replaces
+the `Installation` stub under Get Started.
+
+The Linear issue has no description, so the spec is `OME-666`'s Installation paragraph —
+restructured, because that paragraph is local-first and names API that no longer exists
+(`sf.config`, engine port `:4404`).
+
+Two paths, in order:
+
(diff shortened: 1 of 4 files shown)


callezenwaka:
Screenshot 2026-08-08 at 12 59 29
Screenshot 2026-08-08 at 12 59 19

PR #545, opened 2026-08-10
fix(url4-cloud): reclaim JetStream streams so runs stop failing with 10047

IonesioJunior:
## Summary

Production `url4-cloud` failed **every** run with `err_code=10047 insufficient storage resources available`.

Each run declared its own JetStream stream reserving **256 MiB** against a 10Gi store — a hard ceiling of **40 streams** — and nothing reclaimed them:

- `max_age` expires **messages**, never the stream object, so an empty stream keeps its full reservation. The invariant comment in `jetstream.py` claimed the opposite, and that false belief is why nothing reclaimed anything.
- The only delete path (`DELETE /`) needs a capability token, and `JwtCodec.sign` sets `exp = iat + iat_window_s` (**60 s**) while `POST /token` always mints a *fresh* topic — so a token can never be re-issued for an existing topic. **Any run longer than a minute was structurally unable to tear its own stream down.**
- `except BadRequestError` never caught it: `ServerError` (500) and `BadRequestError` (400) are siblings under `APIError`, so 10047 escaped as an unhandled 500.

Failure was monotonic and total: past ~40 leaked runs, every subsequent run failed.

## What changed

- **The runner reclaims its own stream** after a drain grace (`run_and_reclaim`), in a `finally` so a failed run reclaims too. The grace exists because `delete_stream` destroys the stream *and its consumers* — deleting on the terminal frame would race a draining client.
- **Lazy sweep on 10047**, retried exactly once, covering runs whose pod died before teardown. No background reaper, no leader election, no extra RBAC. An orphan is *proven*, never guessed: emptied by `max_age`, terminated past the grace, or created long ago with nothing published and nobody attached.
- **`max_bytes` 256 MiB → 50 MB** — ceiling 40 → ~200.
- **`activeDeadlineSeconds` now outlives the teardown it must allow.** It previously equalled the run deadline *exactly*, so k8s killed the pod during the drain grace and every timed-out run leaked.
- **Teardown failures never escape.** `delete_stream` connects lazily, so a broker blip raises outside `APIError` — which would report a good run as Failed, or (inside `finally`) *replace* a failed run's real exception.

`packages/url4` is untouched: `delete_stream` was deliberately **not** promoted to `EventPublisher`, since `EventStream(EventPublisher, EventConsumer)` would resolve to it and silently break local-mode teardown.

## Test plan

`uv run .claude/scripts/run_gates.py url4-cloud` → **all green** (ruff, format, pyright, layering, pytest ≥80% cov). 1147 tests pass; 26 new across two files.

Two prior tests changed under the SDLC Confidence Gate, both owner-approved:
1. `test_schedule_builds_a_run_once_named_spec` pinned `activeDeadlineSeconds == deadline_s` — the exact equality that was the defect.
2. `test_runner_job_env_isolation` enumerated the per-run env set; extended by `STREAM_GRACE_S`. Intent unchanged, assertion still exact.

Gates therefore ran with the documented `--skip-append-only`.

## Reviewer notes

Two review passes (senior + adversarial) ran against this branch. They caught a **critical** hole in the first draft: `messages == 0, last_seq == 0` is not only the *fresh* state but the *permanent* state of a topic whose runner never published — reachable via `ImagePullBackOff`, quota rejection, or a world-resolution crash — which made the sweep unable to clear the very outage it exists for. Fixed via `StreamInfo.created` (server-side, so no runner clock is trusted) + `consumer_count == 0`.

**Deliberately deferred** (recorded in the ledger): spec §3.5's RFC 9457 **503** mapping, so a 10047 surviving the sweep is still a raw 500 on the REST path; plus `_ensured` invalidation on out-of-process deletion, clock-skew on the terminated test, a SIGTERM handler, and multi-deployment brokers.

## Not retroactive — one manual step before this helps

Streams already stranded in production must be cleared once; the lazy sweep only fires once the store fills again:

```sh
nats stream ls --names | grep '^url4-cloud_' | xargs -n1 nats stream rm -f
```

## Known residual risk

A pod killed before its `finally` still leaks until the next sweep. The sweep bounds the *consequence* (a run fails, sweeps, recovers) rather than eliminating the leak. The shared-stream redesign — one stream, subject per run, `max_msgs_per_subject` — removes the failure mode outright and remains the recommended follow-up.

Also note: every test runs against a hand-written `_FakeJetStream`, which cannot catch a wrong assumption about JetStream itself — and two of the confirmed review findings were exactly that class of bug. An integration test against the kind rig would close it.

No Linear item — owner directive to spec and implement directly.
Spec: `docs/spec/2026-08-10-url4-cloud-jetstream-stream-reclamation.md`
Ledger: `docs/work/2026-08-10-no-ticket-jetstream-stream-reclamation.md`

### DIFF
--- a/apps/url4-cloud/src/url4_cloud/adapters/jetstream.py
+++ b/apps/url4-cloud/src/url4_cloud/adapters/jetstream.py
@@ -4,27 +4,46 @@ between the Runner and the App. Subject and stream names are per-topic, derived 
 `url4_cloud.subjects.subject_for`/`stream_for` rather than reimplemented here."""
 
 import asyncio
+import logging
 from collections.abc import AsyncIterator
+from datetime import UTC, datetime
 
 import nats
 from nats.aio.client import Client
 from nats.js import JetStreamContext
-from nats.js.api import AckPolicy, ConsumerConfig, DeliverPolicy, DiscardPolicy
-from nats.js.errors import BadRequestError, NotFoundError
+from nats.js.api import AckPolicy, ConsumerConfig, DeliverPolicy, DiscardPolicy, StreamInfo
+from nats.js.errors import APIError, BadRequestError, NotFoundError
+from pydantic import ValidationError
 
 from url4.streaming.codec import decode, encode
 from url4.streaming.interfaces import EventConsumer, EventPublisher, validate_from_sequence
-from url4.streaming.protocol import OutboundFrame
-from url4_cloud.subjects import stream_for, subject_for
+from url4.streaming.protocol import OutboundFrame, TerminatedEvent
+from url4_cloud.subjects import owns_stream, stream_for, subject_for, topic_of
 
-# INVARIANT: a run's stream must outlive the run itself plus the window in which a client may
-# still attach and replay it. This is the ceiling for BOTH, and it is what stops an abandoned
-# run's frames from sitting in the filestore forever: `purge` deletes the stream on the normal
-# DELETE path, and `max_age` reclaims the ones no client ever gets around to deleting.
+logger = logging.getLogger(__name__)
+
+# INVARIANT: `max_age` bounds the BYTES a run's frames occupy. It does NOT reclaim the stream
+# object — JetStream expires messages and leaves the stream, its consumer state and its
+# filestore directory in place, still holding the whole `max_bytes` reservation. Reclaiming a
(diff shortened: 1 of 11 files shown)


keelancj:
looks good 🙌 core fix handles the stuck streams and nice test coverage 🙂

PR #558, opened 2026-08-11
feat(scoreboard): rebuild leaderboard portal shell on SFDS v2

FilipBolt:
## Summary
- Rebuilds the scoreboard portal's landing (benchmark catalog) and per-benchmark board shell to render entirely from the live API, with a tab strip and `?id=` deep-linking across registered benchmarks.
- Explicit empty/error states: zero benchmarks registered, unknown `?id=`, and API failure each get their own message instead of a blank or broken shell.
- Migrates the portal off its hand-vendored SFDS v1 design system onto the current v2 marketing register (self-hosted Plex Sans/Mono + Parastoo fonts, gold `--gain`, v2 tokens/component recipes byte-matched to the `screamingface-design` skill's `reference/`), including vendoring the actual font binaries (never checked in anywhere in this repo before, including the design skill itself).
- Drops the "built by OpenMined" footer mark per Bennett Farkas's 2026-08-07 flag.
- Submission-row population, cost, and the reproducible/verified toggle are out of scope for this unit (OME-769/770/771 — filed, queued next).

## Open item before merge (not blocking review)
Spec flags the catalog row "subtitle" as provisionally mapped to `Benchmark.description`, pending a one-line confirmation from Irina — see OME-768 spec/ledger for detail.

## Test plan
- [x] `run_gates.py scoreboard` — append-only, ruff check/format, pyright, pytest (167 passed, 2 skipped, ≥80% coverage) — all green
- [x] Ran locally against a fresh migrated DB seeded with `draco`/`ifeval`; verified via Chrome DevTools: zero-benchmark empty state, unknown-`?id=` state, tab strip + deep-linking, per-benchmark empty-submissions state, light/dark theme
- [x] Confirmed no `--gain`-as-success usage remains
- [ ] Irina's sign-off on the subtitle field mapping

Refs: OME-768


FilipBolt:
Decision: merging with the subtitle field mapping still provisional (`Benchmark.description`, per the spec's open question) rather than waiting on Irina's confirmation — priority is getting the live-wired shell + SFDS v2 migration shipped end-to-end now; OME-769/770/771 build directly on top of it either way.

If the mapping needs to change, it's a one-line template fix in `index.html`/`benchmark.html` — no structural risk. Will follow up with a quick patch once confirmed, rather than blocking this merge on it.

CI green, gates green, no data-model changes in this unit.

FilipBolt:
Fixed a review finding: \`portal.css:106\` referenced \`var(--accent)\`, which \`tokens.css\` never defines (only the \`--accent-*\` role family — \`--accent-solid\`, \`--accent-border\`, etc.). Pointed it at \`--accent-solid\` instead (\`e4d3acef\`).

One correction to the original finding for the record: the invalid reference didn't drop the border/underline — it fell back to \`currentColor\` per the CSS custom-properties spec, so the active tab's underline was still fully rendered, just silently in the same color as its own text rather than the intended accent. Verified both before and after in Chrome (computed \`border-bottom-color\` matched \`color\` before the fix; now resolves to \`--accent-solid\`, e.g. \`#e2a35b\` in dark/marketing).

FilipBolt:
Ran a full review pass over this branch and pushed fixes in `bfd2aa3c`. Seven findings addressed, three deliberately deferred.

**Correcting my own PR body first:** the "byte-matched to the design-system reference" claim above was **false when I wrote it**. The branch shipped a *stale* copy carrying exactly the drift the reference annotates as already-fixed — `h1` on `--weight-semibold` instead of `--weight-display` (the reference comments that semibold "was a drift from the 2026-07-16 decision"), the label role on semibold instead of medium, `.kv dt` missing from the label selector, the pre-mask `.badge-verified` recipe, and a dropped nested dark+marketing selector in `tokens.css`. The reference hadn't moved since this branch's merge-base, so it was a stale copy rather than a rebase artifact. Re-copied verbatim; `cmp` is now clean on all three files, so the drift check this migration exists to restore actually works.

**Also fixed:**
- **Sticky-rail regression + floating toggle.** `style.css` ships `.rail { position: relative }` plus a `.rail.stuck` variant its own comment documents as JS-toggled, and this portal supplies no such listener — so the rail scrolled away while `.rail .toggle` (`position: fixed`, no background per the system) stayed pinned, leaving a bare theme-toggle glyph over page content, and over `benchmark.html`'s rightmost table cells at ~1000px viewports. Restored `position: sticky` in `portal.css`. Note this is **one deliberate override** of that file's own "extends — never overrides" header rule; it carries an `AIDEV-NOTE` explaining why and pointing at the system-faithful fix (implement the `.stuck` listener + spacer), which wants its own ticket rather than being rushed in here.
- **"Submissions" column was publishing wrong numbers.** The value is `entries.length` from a best-per-spec leaderboard capped at the route's `MAX_LEADERBOARD_TOP` — so a benchmark with 50 submissions across 3 specs displayed `3`, and anything past 200 displayed a flat `200` with no truncation marker. Relabelled **Fusions**, the term OME-769 already uses.
- Empty `.tabstrip` no longer paints a stray hairline above the masthead or announces an empty navigation landmark on API failure / zero benchmarks.
- Catalog rejections are now caught for the whole chain. The handler was the second argument to the first `.then`, so a throw in the success path or inside the `Promise.all` continuation became an unhandled rejection and stranded the page on "Loading benchmarks…" with no error state.
- OM affiliation mark removed from `spec.html` and `data.html` — it had only been removed from index/benchmark, so it survived one click deep on `spec.html`, which is reachable from every leaderboard row.
- Added `assets/fonts/OFL.txt`. The 20 vendored woff2 files are OFL-1.1 and serving them is redistribution, which requires the license and copyright notices to accompany them. IBM Plex's notice is verified against `IBM/plex` upstream. **Parastoo's is recorded but explicitly flagged unconfirmed in the file** — the family name matches an archived Persian font while SFDS v2 documents Parastoo as a Latin editorial serif; @bennett-farkas authored the system and is the person to confirm the true upstream.

**Deferred (not fixed here):**
- The catalog blocks on N leaderboard fetches before painting any row — inert at today's benchmark count, and OME-769 rewrites this exact code path, so it belongs there rather than as a rushed render-then-fill refactor.
- Spec D10 promised a retry affordance on the API-failure state; it ships as static text. Small, real, follow-up.
- `encodeURIComponent` on slash-containing benchmark IDs still 404s — already tracked as OME-776, which is `blockedBy` on this ticket.

Verified locally against a fresh migrated DB seeded with `draco`/`ifeval`: rail stays pinned at max scroll with the toggle inside its band, active-tab accent underline resolves, `.tabstrip:empty` collapses, `OFL.txt` serves 200, all `var(--…)` still resolve after the re-copy, and gates are green.

apps/scoreboard/portal/index.html
```diff
@@ -59,9 +57,8 @@ <h2>Benchmarks</h2>
       <table id="benchmark-table" aria-label="Benchmarks">
         <thead>
           <tr>
-            <th>Name</th>
-            <th>id</th>
-            <th>Description</th>
+            <th>Benchmark</th>
```
HupBaHa, line 60: OME-768 still specifies **`# submissions`**, but this column now reports the number of best-per-spec leaderboard entries, capped at 200. Renaming it to `Fusions` avoids publishing that value as a false submissions count, but it changes rather than fulfills the Linear requirement.

Please either source the actual submission count or get the requirement explicitly changed in OME-768 before merging.

apps/scoreboard/portal/portal.css
```diff
@@ -81,8 +81,54 @@ table .btn { white-space: nowrap; }
...
+   `.rail .toggle` (position: fixed, no background/border per the system) stayed
+   pinned top-right — leaving a bare "◐ light" glyph floating over page content,
+   and over benchmark.html's rightmost table cells at ~1000px viewports.
+   Restoring `sticky` keeps the rail under the fixed toggle at all scroll
+   offsets, which is the behavior this portal had before the v2 migration.
+   AIDEV-NOTE: this knowingly breaks the "extends — never overrides" rule at the
+   top of this file. The system-faithful fix is to implement the `.stuck` toggle
+   + spacer in JS; that is new untested behavior, so it is tracked separately
+   rather than folded into OME-768. Drop this block when that lands. */
+.rail { position: sticky; top: 0; }
```
HupBaHa, line 134: Non-blocking: the byte-matched SFDS rule adds a `::before` glyph that requests `/assets/remix/System/check-line.svg`, but this portal does not ship that asset, so every verified badge produces a 404. Because the markup already contains a literal `✓`, suppressing the unused pseudo-element locally avoids the failed request.

```suggestion
.rail { position: sticky; top: 0; }
.badge-verified::before { content: none; }
```

HupBaHa (changes requested):
Requesting changes because the current implementation does not yet match two requirements in the governing Linear issue OME-768.

1. OME-768 asks for an empty per-benchmark table structure, while ranked rows and the SOTA treatment are explicitly owned by OME-769. In `apps/scoreboard/portal/benchmark.js`, the current page keeps the existing OME-769 behavior for non-empty responses (ranked rows, summary, climb chart, and SOTA state), while the zero-entry path returns before `renderHead()` and before unhiding `#leaderboard-wrap`, so it hides even the required table shell. Please align the page with the OME-768 shell-only boundary, or first reflect an intentional scope change in Linear.

2. OME-768 asks for `# submissions`. The PR now displays `Fusions`, which is more accurate for `leaderboard.entries.length`, but this changes rather than implements the Linear requirement. The required submissions count needs either an appropriate data source or an explicit scope decision in Linear.

OME-775 and OME-776 remain external blockers and are not findings against this PR. The focused tests and current CI are green, but no test executes the portal JavaScript behavior above.

FilipBolt:
Thanks @HupBaHa — all three addressed. Two fixed in code, one where I'm partly pushing back with evidence.

### 1. Empty table shell — agreed, fixed in `df6f21fa`
You're right, and this was a real gap against OME-768's own acceptance. The zero-entry path returned before `renderHead()` with `#leaderboard-wrap` still hidden, so a registered benchmark with no submissions showed the empty-state line and no table at all. Now the column headers render with an empty `tbody` alongside that line (`renderSummary`/`renderClimb` already self-hide on an empty list, so nothing misleading appears). Verified locally against a seeded benchmark with zero scores: `wrapHidden: false`, `headRowCount: 1`, `bodyRowCount: 0`.

### 1b. Ranked rows / SOTA — pushing back, with evidence
These aren't OME-769 work this PR pre-built; they **already exist on `origin/main`** and this branch inherited them:

```
$ git show origin/main:apps/scoreboard/portal/benchmark.js | grep -c "function renderSummary"  → 1
                                                                     renderClimb              → 1
                                                                     renderBody               → 1
                                                                     renderHead               → 1
```

This branch's entire `benchmark.js` diff adds **only** `initTabStrip()`, `showNotFound()`, and a 404 branch — no row/SOTA rendering. The spec's Context section says as much ("`portal/` already has a working `index.html` … and `benchmark.html` — this unit is a completion + design-system migration, not a from-scratch build"), and D5's wording is "don't **pre-build** submission-row rendering here", which this unit didn't.

Stripping them to match the boundary literally would delete rendering that already ships live, leaving an empty shell on the public board until OME-769 lands. I'd rather not trade working functionality for a boundary description — but say the word if you disagree and want it gone.

### 2. `# submissions` → amended in Linear rather than renamed unilaterally
Fair challenge — renaming did change rather than implement the requirement. So I checked whether the original was satisfiable, and it isn't: `/v1/leaderboard/{id}` returns **best-per-spec** entries capped at `MAX_LEADERBOARD_TOP`, and there's no aggregate endpoint (catalogued in OME-772). A "Submissions" column sourced from it shows **3** for a benchmark holding 50 submissions across 3 specs, and a flat **200** past the cap.

So I've **amended OME-768's scope line** to `# fusions` with the rationale in [a comment there](https://linear.app/openmined/issue/OME-768), flagged to @Irina to revert if she objects. `# fusions` is also what OME-769 already calls this quantity. A true count needs a new aggregate endpoint — its own ticket, not this portal-only unit.

### 3. Verified-badge 404 — fixed in `9ada95ed`, and your suggested fix needed one addition
Good catch, and this one was mine: it came in with the byte-match re-copy, not from `main`. Worth flagging that `content: none` alone **doesn't** stop it — I applied exactly that first, and the request still fired:

```
reqid=18 GET http://localhost:9106/assets/remix/System/check-line.svg [404]
```

Chrome resolves the mask URL even when the pseudo-element isn't generated. Clearing `mask`/`-webkit-mask` too gets it to zero image requests. Left an `AIDEV-NOTE` on the block so nobody "simplifies" it back later.

---

Gates green after each commit; CI green. Note the `portal.css` overrides (rail pinning + this badge block) knowingly break that file's "extends — never overrides" header rule — both carry `AIDEV-NOTE`s explaining why and pointing at the system-faithful fix (implementing the `.stuck` scroll listener + spacer), which I think wants its own ticket rather than being folded in here. Happy to file that now if you'd prefer it tracked before merge.

HupBaHa (approved):
Rechecked at df6f21fa. The empty benchmark now renders the required table shell, OME-768 explicitly reflects the fusions terminology, and the verified-badge asset request is eliminated. I also confirmed the ranked-row/SOTA renderers predate this PR, so the author's scope pushback is valid. Focused tests and CI are green.

PR #559, opened 2026-08-11
Package the local ScreamingFace Studio runtime

itstauq:
## Summary

- add one supervised Python process for AI Gateway and ScreamingFace Engine local mode
- run without Docker, Kubernetes, NATS, or Postgres
- apply AI Gateway SQLite migrations in-process and persist state under a configurable data directory
- expose machine-readable readiness and startup failure records for a future Tauri supervisor
- embed the URL4 runner config and guard it against drift from the container config
- add a credential-free token/WebSocket/URL4 execution smoke test
- build a macOS arm64 PyInstaller `onedir` sidecar using the packaging pattern from syft-space

## Verification

- `uv run --frozen pytest` — 6 passed
- both `/healthz` endpoints return successfully
- Engine discovery returns 15 models and 5 connections
- `screamingface-runtime-smoke` completes the real execution protocol
- frozen executable starts from `/tmp`, applies migrations, announces readiness, serves both APIs, and shuts down both services from one signal
- occupied ports produce a concise non-zero startup failure

## Artifact

```text
apps/screamingface-studio/runtime/dist/screamingface-runtime/
├── screamingface-runtime
└── _internal/
```

The current unoptimized macOS arm64 bundle is approximately 198 MB. Build output is ignored and is not committed.

## Still WIP

- copy/name the onedir bundle as a Tauri resource
- start, monitor, and terminate it from the desktop app
- support local runtime versus a configurable cloud Engine URL
- add release signing, notarization, target-triple naming, and other platform builds
- trim broad LiteLLM collection after the supported provider paths are pinned by frozen-runtime tests
- continue replacing frontend model-page mocks after the local runtime lifecycle is integrated

## Notes

This PR intentionally targets `integration/keelan-all-changes-20260806`, which is the branch this work was based on. Targeting `main` directly would include unrelated integration commits.


itstauq:
Superseded by #590, which replays and verifies this work against current main.

PR #544, opened 2026-08-10
feat(url4-cloud): add HealthBench challenge protocols

keelancj:
## Summary

- ports HealthBench smoke and frozen worst-30 challenge protocols to the current Engine benchmark registry
- preserves native multi-message Candidate conversations and validates malformed chat envelopes before model execution
- adds pinned preparation, official rubric grading, fail-closed aggregation, and complete Case artifacts
- adds an SDK end-to-end challenge notebook using the current Fusion synthesizer API

## Dependency

Draft and blocked by #539. This branch is based on the client contracts introduced there. After #539 merges, this branch must be rebased onto current main so this PR contains only the benchmark work.

## Validation

- focused HealthBench and connector tests: 97 passed
- URL4 Cloud: 1,171 passed, 5 skipped
- ScreamingFace SDK: 620 passed, 1 skipped
- Ruff, formatting, Pyright, and notebook checks pass

## Provenance

Ports and updates the benchmark work from #523 for the current benchmark, Candidate, and client architecture.

PR #566, opened 2026-08-12
fix(screamingface): surface failure identity and failed-state semantics in the report view

khoaguin:
Refs: OME-793 · https://linear.app/openmined/issue/OME-793

### Before
- Trigger: a dev opens the report widget on a run where cases died (the 2026-08-11 HealthBench worst30 fusion run: 3 of 10 cases lost to gateway failures).
- Today: the failure banner repeats `candidate · no evaluation row for this Case reached the aggregate` three times with no case ids; a never-graded case wears the same red INCORRECT badge as a genuinely wrong answer; the failed case's detail pane renders nothing useful; score / pass rate / coverage show bare dashes with no reason.
- Cost: every failure investigation drops to raw report.json; infra failures read as bad model answers in screenshots and team discussion.

### After
- Same trigger.
- Now: the banner groups identical failures into one line naming the cases, code, and underlying error (`candidate · missing_case_row · cases 153, 149, 418 — … (ResolutionError: malformed aigateway response)`); failed cases carry a distinct warning-styled FAILED state; the pane renders the failures chain plus an explicit input-unavailable note; unscored candidates say `score withheld — 3 of 10 cases failed`, and an absent cost carries a "cost not reported by this run" title.
- Win: triage happens in the widget; a screenshot of it is self-explanatory.

### Don't regress
- Graded cases' rubric/check rendering unchanged; prior tests untouched (append-only check green).
- `report.to_json()` byte-identical — display-only change.
- The B1 score-withholding rule itself is untouched; the strip only explains it.

## Test plan
- 7 new tests in `test_report_panel.py`: banner grouping + identity, tri-state failed badge/mark, failure-chain pane, withheld-score strip, cost title, no-code failure fallback, escaped untrusted failure text.
- Full `run_gates.py screamingface` green: ruff, format, pyright, pytest --cov ≥95, notebook determinism, build, distribution.

### DIFF
--- a/docs/tasks/2026-08-12-report-failure-ux.md
+++ b/docs/tasks/2026-08-12-report-failure-ux.md
@@ -1,0 +1,17 @@
+---
+id: OME-793
+linear_url: https://linear.app/openmined/issue/OME-793/surface-failure-identity-and-failed-state-semantics-in-the-report-view
+status: in_progress
+type: bug
+priority: high
+labels: [py-screamingface, agentic, autonomous]
+created: 2026-08-12
+closed:
+---
+
+# Surface failure identity and failed-state semantics in the report view
+
+The report widget discards failure evidence the report already carries: banner lines with
+no case ids, INCORRECT badges on never-graded cases, "None" detail panes, unexplained
+metric dashes. Display-only fix in `_ui/report_view.py`; scope and Before/After in Linear.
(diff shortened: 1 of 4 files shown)


keelancj:
Closing as already landed: the complete OME-793 commit was carried into and merged through #544 as 65de0d7d.

PR #567, opened 2026-08-12
feat(aigateway): add per-provider call usage accounting

HupBaHa:
## Summary

- add bounded per-provider-attempt usage accounting for OpenRouter and Anthropic
- enable accounting by default for every non-streaming chat completion; callers send no activation header
- keep `stream: true` on the existing SSE path without `_aigw` accounting
- capture token, latency, cache, failure, and provider-authored direct-cost evidence without persisting gateway `_aigw` metadata in the response cache
- preserve hidden LiteLLM resend cardinality, collapse redirect hops without inflating economics, and mark ambiguous redirect cycles partial
- ship a closed Draft 2020-12 schema, provider-backed release fixtures, leakage regressions, implementation plan, and runtime contract
- keep the evolving pre-beta taxonomy unversioned; compatibility is not guaranteed until beta

## Current Contract

Non-streaming accounting is gateway-owned and default-on. `X-AIGW-Accounting` is no longer read or required. Legacy values are ignored and are never forwarded to providers or included in the cache key.

Streaming accounting is intentionally out of scope. `stream: true` preserves the existing SSE behavior, creates no accounting session, injects no accounting client, and returns no `_aigw`.

`_aigw` is a gateway-reserved response namespace. It is attached only to returned copies; cached provider JSON remains unchanged.

Runtime contract: `apps/aigateway/docs/usage-accounting.md`  
Implementation plan: `docs/tasks/aigw/2026-08-12-OME-303-per-provider-attempt-usage-accounting.md`

## Scope

AIGateway produces provider-call facts only. Engine attribution, USD conversion, run rollups, persistence, and UI remain out of scope.

## Verification

- usage-accounting suite: `377 passed`
- affected provider/cache/security suite: `119 passed`
- final configured AIGateway gates: `ALL GATES GREEN`
- wheel includes `aigateway/core/usage_accounting/usage_accounting.schema.json`
- documented JSON example validates against the packaged schema
- two-stage independent review: `GO`, no findings
- 1,092 redirect chains probed in the prior correction round
- live network/TLS probe through the accounting client reached expected HTTP `401` from OpenRouter and Anthropic without credentials

## Process Note

The default-on pre-beta contract required owner-approved rewrites of prior opt-in response-shape assertions. The full gate was run with the append-only check explicitly skipped and all runtime/static gates passed. Push used the user-authorized `--no-verify` exception because the repository pre-push hook has no mechanism to encode that approval.

## Residuals

- handler initialization remains fail-fast when local TLS configuration is broken
- full raw JSON parsing remains bounded at 256 KiB; selective extraction is deferred until live response-size evidence justifies it
- no compatibility guarantee applies before beta; incompatible changes update code, schema, fixtures, plan, and spec together

Refs: OME-303


apps/aigateway/tests/unit/usage_accounting/test_classify.py
```diff
@@ -0,0 +1,145 @@
...
+        _outcome, code = classify_conversion_failure()
+        assert code in FAILURE_CODES
+
+
+class TestTheVocabularyIsClosed:
+    def test_no_provider_name_appears_in_any_failure_code(self) -> None:
+        # A code naming a provider would mean the taxonomy had leaked out of core.
+        for code in FAILURE_CODES:
+            lowered = code.lower()
+            for provider in ("anthropic", "openrouter", "openai", "gemini", "ollama", "litellm"):
```
sergio-bershadsky: Wondering if we will add new provider in future will this test be fixed automatically. Or it should become a part of test configuration/bootstrap

docs/work/2026-08-07-OME-303-per-provider-call-usage-accounting.md
```diff
@@ -0,0 +1,794 @@
...
+started: 2026-08-07
+finished: 2026-08-11
+---
+
+# OME-303 — AIGateway per-provider-call usage accounting (evidence contract MVP)
+
+## Intent
+
+Give AIGateway an opt-in, provider-extensible **evidence contract** for locally observed
+provider send admissions. A caller that sends `X-AIGW-Accounting: v1` gets two sibling
```
sergio-bershadsky, line 14: Me and @HupBaHa decided that we need to remove versioning from taxonomy accounting as we don't expect V(n) to be appeared soon, treat current acconting layer as alpha so reverse incompatible changes can be introdused in future until it will become beta. Please change this in current plan and spec. 

Please also provide a plan a spec in this PR.

sergio-bershadsky (changes requested):

apps/aigateway/tests/unit/usage_accounting/test_classify.py
HupBaHa: Addressed in d6a559c6. The guard now loads the real ProviderRegistry and derives normalized provider terms from every registered plugin ID and display name, including compound aliases such as gemini, gemini_cli, geminicli, hugging_face, and huggingface. It fails explicitly if plugin discovery is empty and retains openai/litellm as protocol or implementation leak terms. Focused classification tests pass, the full accounting suite is 367 passed, and the AIGateway pre-push gate is green.

docs/work/2026-08-07-OME-303-per-provider-call-usage-accounting.md
HupBaHa: Addressed in d6a559c6. Accounting schema IDs, transport capabilities, provider extension namespaces, schema/test filenames, and the temporary opt-in value are now unversioned. The current pre-beta policy explicitly allows incompatible changes until beta without putting an alpha marker in the wire. Added the requested tracked plan at docs/tasks/aigw/2026-08-12-OME-303-per-provider-attempt-usage-accounting.md and runtime spec at apps/aigateway/docs/usage-accounting.md. The final accounting suite is 367 passed and all configured AIGateway gates are green.

sergio-bershadsky (approved):

PR #569, opened 2026-08-12
feat(scoreboard): fill the leaderboard board with ranked rows and core columns

FilipBolt:
Linear: [OME-769](https://linear.app/openmined/issue/OME-769/leaderboard-fill-submissions-board-ranked-rows-core-columns-sota-medal)

> **Scope corrected after review.** The original title and body said this PR shipped a **SOTA medal**. It no longer does — the medal was removed in `60927ba1` and moved to `OME-771`. This description now matches what actually ships. Reasoning in "Why the medal isn't here" below.

## Summary
Turns the shell OME-768 landed into the real submissions board: ranked rows, an **Author** column, and **accuracy as a value + proportional bar** (the design system's own `.score-cell` recipe). The highest-accuracy row is highlighted and announced to screen readers as **"(highest accuracy)"** — a claim about the number, nothing more.

A **mark column** is reserved and renders empty; `OME-770`'s frontier marks and `OME-771`'s medal populate it.

## Why the medal isn't here
It was built, then **removed rather than repaired**, because it cannot be made truthful from the data this endpoint returns.

The medal must name the best *reproduced* run. But `/v1/leaderboard` builds entries with `RowNumber().over(spec_id)` ordered by **accuracy alone** — verification is never consulted. So a spec holding a verified `0.80` and an unverified `0.90` returns only the `0.90`, and its verified run is invisible to any client. Deriving a medal from that projection can badge a lower verified spec, or none at all.

A server-computed SOTA field would not have fixed it either: when the winner is a row the table does not contain, there is no truthful row to badge. `OME-771` filters the pool **in the query**, which makes the verified run a real row — so the medal belongs there.

Removed with it: the reproducible-SOTA stat, the badge rendering and its progressive enhancement, and the vendored wave-mark asset. `sotaAccuracy` / `isSota` remain in the logic module, still tested, with a note explaining why nothing calls them yet.

The row wording changed for the same reason: it now says **"(highest accuracy)"** instead of "(state of the art, independently reproduced)". The leader is frequently unverified — the live dev board demonstrates this — so the old wording asserted something false.

## Tests — and a note on the harness
`portal/leaderboard-logic.js` holds the pure decisions (SOTA selection, ordering, bar scaling), tested **RED-first** under Node's built-in runner: **14 tests, zero new dependencies, no `package.json`, no vitest.** I'd assumed this needed a new toolchain; it didn't.

The SOTA tests pin the *invariant* — only a reproducible entry may be presented as SOTA — not the current *source* of that signal, because `OME-771` intends to re-source it from the engine's cache ("have we run this URL4 before"). When that lands, one predicate changes and the tests stay valid.

They're mutation-checked: inverting `isReproducible` fails exactly the 3 tests guarding it.

⚠️ **Not yet wired into CI** — that needs `scoreboard-tests.yml` plus the sdlc card's gate list, which is its own unit (`OME-798`). Run locally with `node --test tests/portal/leaderboard-logic.test.js`.

## Two things worth a reviewer's eye
1. **`renderClimb` changed beyond the ticket's scope, deliberately.** Its gradient fill keyed off the raw maximum. Now that the table highlights the raw-maximum row, the two agree — one story colour asserting one thing.
2. **`Questions` is dropped** (owner-approved). Author + the mark column pushed the table from exactly-fitting to 1205px in a 958px container, putting the url4 copy button behind a horizontal scroll. Questions isn't in OME-769's column list and stays on each spec's detail page. **Full disclosure: the table still exceeds the container by ~134px** — `.wrap.wide` is already the widest container available and the rest is genuine content (a 19-char email, a full timestamp). `.table-wrap` scrolls it; the page body never scrolls; Run Locally is reachable.

## Not built — missing backend fields, not choices
- **Fusion name** — no such field exists (`spec_id` is a technical key). Header stays `Spec` rather than promise a name it can't deliver.
- **Models column** — `ran_with_providers` is *providers*, not models; `providers.length > 1` isn't a valid fusion/solo test.
- **Accuracy range whisker** — no min/max/stddev/CI anywhere in the schema.

Each is the same class of gap `OME-772` catalogues.

## Test plan
- [x] `node --test tests/portal/leaderboard-logic.test.js` → **14 passed**
- [x] `run_gates.py scoreboard` → append-only, ruff check/format, pyright, pytest+coverage — all green
- [x] Seeded a deliberate mix (0.99 **unverified**, 0.62 **verified**, 0.15 with a null author) in Chrome: the highest-accuracy row is highlighted and announced as "(highest accuracy)"; **no medal is rendered on any row**, and the mark column is empty throughout
- [x] Column alignment holds across all rows with the mark column empty
- [x] Null author → em-dash; OME-768's empty shell intact (no stray badge); page body never scrolls horizontally
- [ ] `prefers-reduced-motion` — verified by reading the vendored source (`if (reduced.matches) { still(); return; }`), **not** by running it; the tooling here can't emulate the query

## Follow-ups from review (non-blocking, agreed)
- **Scope wording** — this description and the Linear ticket, done here.
- **The in-table accuracy bar is narrow**, so it gives limited visual comparison. Deferred to the `OME-770` / `OME-771` table changes, where the column set changes anyway; the displayed accuracy value itself is correct. Recorded on `OME-771`.

Refs: OME-769


HupBaHa (approved):
Thanks for the careful work here. I’m approving this PR: the implementation improves the truthfulness of the public board by no longer presenting an unverified highest-accuracy result as SOTA, and I don’t see any correctness issue that should block the merge.

Two non-blocking follow-ups:

1. Please align the PR title/body and Linear scope with the decision to move the SOTA medal from OME-769 to OME-771. The current wording still suggests that the medal ships in this PR.
2. The in-table accuracy bar is currently quite narrow, so it provides limited visual comparison. This can be improved or removed later alongside the OME-770/771 table changes; the displayed accuracy itself is correct.

PR #570, opened 2026-08-12
feat(url4-cloud): derive the web-search mechanism from the provider

IonesioJunior:
Closes OME-797 and OME-800 (epic OME-799).

## The problem

Every route in `url4.toml` carried two flags. `web_tools` meant "we run a Tavily search loop for this model ourselves". `native_web_search` meant "the provider searches and answers with it already done". An operator picking a flag had to know, per model, which mechanism that model could actually carry — knowledge that lived in someone's head, not in the code.

**Both defaulted to off, so a route stayed silently non-searching until somebody remembered to declare it.** Adding a model to the catalog gave you a model that couldn't search, with nothing to tell you.

**The file documented one flag and never defined the other.** The header block explains `web_tools` in six lines. `native_web_search` appears only as values on routes.

**And the split wasn't a design choice — it was covering for a bug one app over.** Five of the eight OpenRouter routes were pinned to Tavily by hand. Not because OpenRouter can't search, but because aigateway hardcoded `engine: "native"` in its search envelope, and OpenRouter honours a forced engine *even for a model that has no built-in search* — which errors. The native-or-Exa fallback only applies while `engine` is left unspecified. So "OpenRouter searches natively" was true per **model**, and the config had to carry that list.

That's the real reason two flags existed. Collapsing them without fixing it would have shipped five broken routes.

## The fix

**One flag, `web_search`, defaulting to true.** A route says only *that* it may search. A route that must not search says `web_search = false`.

**The mechanism is derived, not declared.** `provider_of()` takes the segment before the first `/` (an unprefixed id is Anthropic's, matching the gateway's own convention) and tests it against `WEB_SEARCH_NATIVE_PROVIDERS`. Members delegate to the provider; everything else runs the Tavily loop. Deliberately the segment and not a substring — `openrouter/anthropic/claude-opus-4.8` is an OpenRouter route, and a substring test would hand it to a future `anthropic` entry.

**That set holds only `openrouter`,** because that's the one aigateway plugin declaring a `web_search` parameter. The others register bespoke handlers (codex OAuth, gemini-cli, antigravity, our own anthropic) instead of litellm's stock vendor routes, so litellm's `web_search_options` never reaches them. The docstring on the set says what it takes to add an entry.

**aigateway stops naming the engine,** so OpenRouter picks native-or-Exa itself. That is what makes the provider-level rule true rather than merely convenient. Cache adapter revision bumped `08c → 08d` because the envelope shape changed.

Both halves are in one PR because separating them ships a state where five routes force an engine their models don't carry.

## Things I deliberately didn't do

- **Didn't verify the tool round-trip on the 15 non-OpenRouter routes.** They now search by default where a Tavily key exists, and their OpenAI-shape tool calling has never been confirmed end-to-end. You asked for that to be a later PR. Note the failure mode isn't a quiet downgrade — a backend that rejects the tool call *fails* the request, and all 15 flip in one deploy. A deployment with no Tavily key is unaffected.
- **Didn't add a migration path.** `web_tools`/`native_web_search` are gone, with no shim and no dedicated error. `url4.toml` ships inside the image, so a stale config can't meet new code.
- **Didn't touch cache freshness.** Search results still have no TTL (OME-779/780).
- **Didn't replace our Tavily loop with litellm's own web-search interception,** which could do the same job across every provider. Worth its own ticket.

## Please look at this

**I changed 15 existing test files, and the guard that normally catches that was switched off** (`--skip-append-only`, approved before I wrote code). Every other gate ran normally.

Twelve of them are forced renames — the fields they passed no longer exist. These three want your eyes:

| File | What changed |
|---|---|
| `test_native_web_search.py` | **3 tests removed.** They asserted a two-flag config is rejected, and the retired field's default and type check. None is representable with one flag. Each guarantee moved to a named test in `test_web_search_routing.py`; a comment in the file says which. |
| `test_declared_models_match_aigateway.py` | Its cross-check against aigateway no longer applies, so the surviving guard (both mechanisms stay reachable) moved to `test_web_search_routing.py`. |
| `test_openrouter_web_plugin.py` | 4 assertions that pinned `engine: "native"`. |

**And DRACO moves.** All eight lineup models are `openrouter/*`, so five that ran the Tavily loop now delegate. Two consequences, both of which you accepted, both recorded in `benchmarks/draco/aggregate.py`:

- `REVISION` doesn't hash the mechanism, so runs either side of this change publish under the same benchmark revision.
- `EXCLUDED_DOMAINS` — the leakage control covering arxiv, paperswithcode, semanticscholar, alphaxiv — changes from *enforced* (we filter Tavily results client-side) to *forwarded* (we ask OpenRouter to honour it).

## Testing

- `run_gates.py url4-cloud --skip-append-only` — green. 1169 passed, 5 skipped.
- `run_gates.py aigateway --skip-append-only` — green. 3034 passed, 46 skipped.
- New `test_web_search_routing.py` (25 tests): provider resolution including the `openrouter/anthropic/…` trap, mechanism per provider, the partition invariant (a searching route has exactly one mechanism), declaration defaults, and that the shipped config still reaches both mechanisms.
- Plus an executor-level test that the *production* default offers tools — the existing runner fixture defaults the opposite way, so nothing there covered it.

## What you get

`url4.toml` drops to a list of ids. Adding a model is one line, and it searches. The question "does this model do native search or do we loop?" is answered by the code, in one place, with a docstring saying what it takes to change the answer.

Spec: `docs/spec/2026-08-12-OME-797-unify-web-search.md`

### DIFF
--- a/apps/aigateway/src/aigateway/plugins/openrouter_provider/global_cache.py
+++ b/apps/aigateway/src/aigateway/plugins/openrouter_provider/global_cache.py
@@ -58,7 +58,12 @@ from .settings import (
 # never part of any prior revision's contract for this key path. Abandoning the whole
 # generation here is the same discipline as every other adapter-revision bump: readers must
 # never be served an entry keyed under a semantics this module no longer implements.
-GLOBAL_CACHE_ADAPTER_REVISION = "openrouter-global-cache-2026-08c"
+#
+# 08d (OME-800): `apply_web_search` stopped naming the search engine, so the envelope this
+# key path projects changed shape. Entries written while `engine: "native"` was forced
+# describe a request this module no longer sends — a different upstream retrieval path, and
+# for models without built-in search, one that erred rather than answering.
+GLOBAL_CACHE_ADAPTER_REVISION = "openrouter-global-cache-2026-08d"
 
 
 def project_global_cache_request(body: dict[str, Any]) -> dict[str, Any] | CacheBypass:
--- a/apps/aigateway/src/aigateway/plugins/openrouter_provider/web_search.py
+++ b/apps/aigateway/src/aigateway/plugins/openrouter_provider/web_search.py
@@ -15,7 +15,15 @@ WEB_SEARCH_EXCLUDED_DOMAINS_PARAM = "web_search_excluded_domains"
 EXCLUDE_DOMAINS_KEY = "exclude_domains"
 """OpenRouter web-plugin spelling; its server-tool surface uses a different field name."""
 
-_WEB_SEARCH_POLICY: dict[str, object] = {"id": "web", "engine": "native"}
+_WEB_SEARCH_POLICY: dict[str, object] = {"id": "web"}
+"""The web plugin, with the engine deliberately UNSET.
+
+INVARIANT (OME-800): never name an `engine` here. OpenRouter selects the model's built-in
+search when it has one and falls back to Exa when it does not — but only while `engine` is
+unspecified. Naming `native` forces the built-in path even for a model that has none, which
+errors, and turns "this provider searches natively" into a per-MODEL fact every consumer has
+to track as its own list. `apps/url4-cloud` used to carry exactly that list.
+"""
 
 
 def apply_web_search(body: dict[str, Any]) -> None:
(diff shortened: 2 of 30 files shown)


keelancj (approved):
Looks good 😎 verified locally that openrouter gpt5.5 completes draco-lite with native search!

PR #588, opened 2026-08-13
feat(scoreboard): default new submissions to verified as a placeholder

FilipBolt:
Linear: [OME-820](https://linear.app/openmined/issue/OME-820/default-new-leaderboard-submissions-to-verified-ran-on-openmined) · Follow-up: [OME-821](https://linear.app/openmined/issue/OME-821/distinguish-self-reported-runs-from-openmined-run-submissions-on-the)

Action item from the **2026-08-13 dev huddle** (`[52:20]`): *"make verified status default for leaderboard submissions; create task for future non-verified distinction."*

> **Revised after review.** An earlier version of this PR claimed the default meant "the run executed on OpenMined infrastructure". Review showed nothing supports that. The claim is now **withdrawn**, and the public portal copy was changed in the same commit. See "What review found" below.

## Why

`verified_by_openmined` defaulted to `False` and **no HTTP route can set it** — only an uncalled `store.mark_verified()` helper exists. With `OME-414` (re-run verification) unstarted and unstaffed, every row on the board read **"unverified" permanently**.

## What this does

Two lines of substance:

```diff
-    verified_by_openmined = fields.BooleanField(default=False)
+    verified_by_openmined = fields.BooleanField(default=True)
```

and the public copy on both portal pages no longer claims reproduction.

**The default asserts nothing.** It is a placeholder that stops the board reading "unverified" on every row while no verification exists.

## What review found

Two P1 findings. I verified every premise against the code before acting.

**1. The portal already defined the badge as something stronger.** `portal/index.html:41` and `portal/benchmark.html:42` both read *"'Verified' means OpenMined independently reproduced the run."* Flipping the default would have badged **every row** under that definition — publishing the exact claim this ticket's own spec rejected, in the same commit. Testers read the page, not the docstring. This was the merge-stopper.

**2. Nothing attests execution provenance**, so the weaker "ran on our infrastructure" claim was unsupportable too:

| Premise | Evidence |
|---|---|
| SDK allows independent engine/scoreboard URLs | `client.py:45-51` — separate kwargs, separate defaults |
| Production inherits `authMode: disabled` | `values.yaml:41`; `values-prod.yaml` has no override |
| `submit()` never sets the field | `_submission_to_kwargs` — 15 keys, not this one |

A submission is an unattested client payload. "Runs on OpenMined compute" was an assumption about how the cohort would use the system, not a property the write path establishes. My own spec forbids a *client* declaring its trust tier; defaulting to `True` had the *server* assert a tier it cannot verify, which is worse because it looks authoritative.

## The fix

- `portal/index.html` and `portal/benchmark.html` — reproduction claim withdrawn. Both now say scores are self-reported and that the column does not yet distinguish rows.
- `models/score.py`, `routes/scores.py` — docstrings corrected to say the field asserts nothing, each carrying the invariant below.

**The rule that matters:** this default and that portal copy change together. If one says verified and the other does not, the board publishes a false claim.

Confirmed only those two files ever made the claim, and **no test pinned the wording**.

## Still true from the original change

- **A client cannot set this field.** Absent from `ScoreSubmission`, so sending it is a `422`. Previously protected only by `extra="forbid"` staying put; now pinned by a test.
- **No backfill.** Existing rows keep `false`. Some are local test submissions that did not run on OpenMined infrastructure.
- **No migration, and that is correct.** `makemigrations` reports "No changes detected" — a Tortoise `default=` is applied in Python, not as a column `DEFAULT`. Consequence: a row inserted by **raw SQL** still gets the database default, not `True`.

## Not fixed, deliberately

Finding 2's attestation gap needs a mechanism that does not exist. That is **`OME-821`**, filed and linked `blocked-by` this ticket. What changed here is that the board no longer claims what it cannot support.

Also: until `OME-821` or `OME-414` lands, **do not build a UI that filters or ranks on this field** — every row carries the same value. `OME-771` currently defines its "Reproducible" pool as `verified: true`, which under this default is everything. Recorded on that ticket.

## Test plan

- **173 passed, 2 skipped**; ruff check, ruff format, pyright clean; coverage above the 80% gate.
- **Live probe** against a freshly migrated database:

  | Check | Result |
  |---|---|
  | new submission | `201`, `verified=true` |
  | `GET /v1/scores/{id}`, leaderboard, history | `verified=true` |
  | client sends `verified_by_openmined: true` | **`422`** |
  | pre-existing `false` row | still `false` |

## Note for the reviewer

**Four prior tests asserted the old default** (`test_leaderboard_routes.py:111`, `test_sf_payload.py:87`, `test_scores_routes.py:130`, `test_store.py:102`) and were flipped `is False` → `is True`, one line each. Escalated as a Confidence-Gate decision per the append-only rule and owner-approved. Nothing removed or weakened; unverified coverage is preserved by a new test that creates an explicit `False` row. The final gate run used `--skip-append-only`; every other gate ran normally.

Full narrative in `docs/work/2026-08-13-OME-820-default-verified.md`; the withdrawn claim is recorded as spec §2.1a.

https://claude.ai/code/session_01XdKCiEFpo188AkWG5kSn1G


HupBaHa (changes requested):
Thanks for the detailed write-up. I see why this was implemented as a temporary placeholder, with the real provenance distinction deferred to OME-821.

There are still a few places that treat the placeholder as real verification: the green `✓ verified` badge on the benchmark and spec pages, the `Verified rows` counter and sort, and the `verified` chip/filter in the Python leaderboard view. `isReproducible` also still treats the flag as evidence of reproducibility. Could we hide or relabel these until OME-821 provides a real signal?

Could we also update OME-820/821 in Linear to reflect the revised placeholder decision? They currently still say that verified means the run happened on OpenMined infrastructure.

One smaller test issue: the `mark_verified` tests start with a row that is already `True`, so they would pass even if `mark_verified()` did nothing. Starting with an explicit `False` row would test the actual transition.

Requesting changes because the current UI still presents the placeholder as a real trust signal. The underlying temporary-default approach otherwise makes sense.

FilipBolt:
Thanks — the UI point was right, and it changed the shape of this PR rather than just its wording. All four asks are addressed; branch rebased onto `main`, CI green.

**1. The UI that presented the placeholder as real verification — removed, not relabelled.**

| Surface | State |
|---|---|
| `✓ verified` badge, benchmark board | column and badge removed (`benchmark.js`, `benchmark.html`) |
| `✓ verified` badge, spec history | badge and `<th>` removed (`spec.js`, `spec.html`) |
| "Verified rows" counter | removed; the summary strip is now two cards |
| sort on the Verified column | gone with the column |
| `verified` chip + "verified only" filter, Python view | removed in `OME-832`, merged as #601 |

Relabelling could not have fixed the counter or the filter. Their problem is not that the word is wrong — it is that the value certifies nothing whatever it holds.

**2. `isReproducible` — kept, but nothing can reach it.**

I did not change the predicate, because `OME-771`/`OME-821` will re-point it. What I did was make sure nothing consumes it: the pages call only `L.barWidth` and `L.orderRows`, so `isReproducible`, `sotaAccuracy` and `isSota` are all unreachable. The board's gold row is now chosen by plain highest accuracy, and its accessible text reads `" (highest accuracy)"` — no reproduction claim anywhere in the path. `createVerifiedBadge` is likewise kept and uncalled, with a note saying not to re-wire it before `OME-821`.

Writing this reply is what caught a related overclaim, so it is worth flagging: the note above `isReproducible` said to change the predicate "and nothing else, the tests pin the invariant not the signal". Half right. The test helper builds fixtures as `{ spec_id, accuracy, verified_by_openmined: verified }`, so the field name is baked in and re-pointing means changing that helper too — two lines, not one. The assertions genuinely do not move. Corrected in `d7f772a5`.

If you would rather these were deleted outright and restored by `OME-821`, say so and I will.

**3. `OME-820` and `OME-821` updated in Linear.**

Both now state that the default asserts nothing, and both record which UI came out and why.

While doing that I found a mistake in my own earlier wording, which is worth flagging because it points the wrong way on the exact question you raised. I had written that the field is now *uniform*. It is not — the no-backfill rule means rows predating the change keep `false`. That is **worse** than uniform: a filter or counter on this field would split rows by whether they predate the default change while presenting itself as a verification signal, so it would look like it measured trust and actually measure submission date.

Both tickets and this PR's comments now say that instead. The same wrong word had also shipped in `OME-832`/#601, where it was the recorded justification for removing the client's filter — #610 corrects it there.

**4. The `mark_verified` tests.**

Correct, and they were vacuous — starting from a row that was already `True`, they would have passed if `mark_verified()` did nothing. They now start from an explicit `False` row, so the transition is actually exercised, and idempotency is asserted by calling twice.

I also added a separate test that reads the migration files and fails if any of them would backfill the column, because the schema in tests is built from the models rather than from migrations — so a future data migration would not have been caught by anything else.

**One dependency cleared:** `OME-820`'s follow-ups recorded that this must not merge before `OME-832`, or Colab would show a verified chip on every row. `OME-832` merged as #601, so that gate is gone.

HupBaHa (approved):
Re-reviewed current head against current main. The prior blocking findings are resolved: verification UI is removed across the portal and Python view, the latent reproducibility helpers are unreachable and documented, Linear is updated, and mark_verified now has a meaningful transition/idempotence test. Local Python/JS gates, static checks, CI, and desktop layout verification are green.

Non-blocking follow-up: docs/positioning.md still describes the scoreboard as re-run-verified/never self-claimed and should be aligned with the new self-reported placeholder contract. This does not block the runtime/UI change.

PR #601, opened 2026-08-14
fix(screamingface): remove the verified chip and filter from the notebook view

FilipBolt:
Linear: [OME-832](https://linear.app/openmined/issue/OME-832/hide-the-verified-chip-and-filter-in-the-python-leaderboard-view) · Pairs with **#588** ([OME-820](https://linear.app/openmined/issue/OME-820/default-new-leaderboard-submissions-to-verified-as-a-placeholder))

Client half of the `OME-820` review fix, raised by `@HupBaHa` on #588.

## Why

`OME-820` makes `verified_by_openmined` default to `True` for every submission, as a placeholder that asserts nothing — nothing re-runs submissions (`OME-414`) and nothing attests where a run executed. #588 withdraws the scoreboard portal's verification UI for that reason.

The Python client did not, and **this is the surface that matters most**: the notebook view is what renders in Colab for the tester cohort. It showed a `verified` chip on any row whose flag was truthy — which, after `OME-820`, certifies nothing — plus a "verified only" checkbox whose handler hides `[data-verified=false]`.

Hidden rather than relabelled. The chip could have been reworded; **the filter could not**, because its defect is that the value certifies nothing either way.

`OME-820` does **not** make the field uniform — `D5` forbids backfilling existing `false` rows, so pre-change rows keep it (Dmitry's note on this PR). That makes the filter *worse* than useless rather than merely useless: it would split rows by whether they predate the default change **while presenting itself as a verification filter**. A control that looks like trust and actually sorts by submission date is worse than one that does nothing.

## ⚠️ The deletion had a trap

`_row_chip` was ordered:

```python
if value.verified:                 # -> "verified"
if value.python_source is None:    # -> "baseline"
return ""
```

Removing the first branch does **not** just remove a chip. A candidate with an unforkable url4 falls through to the second branch and is labelled **"baseline"** — presenting a community submission as an imported single-model reference. That is a worse error than the one being fixed.

The predicate now keys on `kind`, which is what actually distinguishes a baseline row (`_baseline_row` sets `kind="single"`, `_candidate_row` sets `kind="candidate"`). Forkability is a property of the url4 expression and says nothing about provenance.

**This was already broken for unverified candidates.** `test_a_candidate_is_never_labelled_baseline[False]` **fails before any production change** — the `verified` branch was merely shadowing it. So the pre-existing mislabel is fixed too, and the guard would have caught the naive deletion turning `[True]` into the same bug.

## What changed

- `_ui/leaderboard_view.py` — dropped `verify_action`, the checkbox markup, the `data-verified` attribute (its only reader was the removed handler), and the `verified` chip branch; baseline predicate now `kind == "single"`.
- `_ui/leaderboard_style.py` — comment only. The `.sf-lb__checkbox` rules are kept unused because `OME-821` restores the control. Deliberately terse: **comments in that file ship to every reader in page source.**
- `Leaderboard.verified_by_openmined` is untouched. The API still returns it; only presentation is withdrawn.

## Test plan

- **All seven `screamingface` gates green**, including the **95% coverage floor**, the notebook check, `uv build` and the distribution check. **783 passed, 1 skipped.**
- New parametrised tests across verified × forkable: no `verified only` control, no checkbox markup, no `data-verified`, no verified chip — **including for a candidate whose `verified_by_openmined` is true**, which the existing assertion at `:787` never covered.
- The mislabel guard above.

## Notes for the reviewer

**One prior assertion inverted, owner-approved.** `tests/test_leaderboards.py:356` asserted `"verified only" in html`. Escalated per sdlc rule 5 rather than edited silently, and **inverted rather than deleted** so it still catches the control being re-added before `OME-821`. Its neighbours already assert absence the same way (`"cost" not in html`, `"mine only" not in html`). The gate run used `--skip-append-only`; everything else ran normally.

**`tests/test_leaderboards.py:787` is now unfalsifiable**, and I left it alone. It asserts the verified chip is absent for an *unverified* entry; with the chip gone entirely it passes trivially. The new tests carry the real coverage. Flagging it rather than quietly relying on it.

**My own tests caught two of my own mistakes**, both worth knowing: `LEADERBOARD_STYLE` is inlined into the same HTML string, so an assertion on the *class name* failed on dead CSS rather than a rendered control (narrowed to the markup); and the explanatory CSS comment I first wrote contained the literal phrase `"verified only"`, which also ships inline and tripped the same assertion.

## Sequencing

**#588 should not merge before this one.** Otherwise new submissions default to a `true` that certifies nothing while the notebook still badges them, against a board that has just withdrawn the claim. Either order works if they land together; this one first is safest.

https://claude.ai/code/session_01XdKCiEFpo188AkWG5kSn1G

### DIFF
--- a/docs/plan/2026-08-14-OME-832-hide-verified-chip.md
+++ b/docs/plan/2026-08-14-OME-832-hide-verified-chip.md
@@ -1,0 +1,48 @@
+# OME-832 — Implementation plan
+
+Spec: `docs/spec/2026-08-14-OME-832-hide-verified-chip.md` · Ledger:
+`docs/work/2026-08-14-OME-832-hide-verified-chip.md`
+
+One file of production code, one stack (`screamingface`). RED before GREEN.
+
+## Step 1 — RED
+
+Append to `packages/screamingface/tests/test_leaderboards.py`:
+
+1. `"verified only"` absent from the rendered board.
+2. The `verified` chip absent for a candidate with `verified_by_openmined=True` — the strong form
+   the existing `:787` assertion does not cover.
+3. `data-verified` absent.
+4. The `baseline` chip still present for a baseline row.
+5. **The mislabel guard:** a candidate with no forkable url4 is not labelled `baseline`. This is
+   the assertion that catches spec §3's trap, and it should fail *both* before the change (chip
+   says `verified`) and against a naive deletion (chip says `baseline`).
+
+## Step 2 — GREEN
+
+`_ui/leaderboard_view.py`:
+
+- delete `verify_action` and the `sf-lb__checkbox` label markup;
+- delete the `data-verified` attribute from `_board_row`;
+- `_row_chip` becomes a single `kind == "single"` test returning the `baseline` chip.
+
+`_ui/leaderboard_style.py`: note the now-unused `.sf-lb__checkbox*` rules, kept for `OME-821`.
+
+## Step 3 — the prior assertion
+
+`:354` asserts `"verified only" in html` and must invert. **Escalate first** (sdlc rule 5), then
+change one line and note why in place.
+
+## Step 4 — gates
+
+`uv run .claude/scripts/run_gates.py screamingface --base origin/main`. Note this stack's list is
+longer than scoreboard's: ruff, pyright, **pytest with a 95% coverage floor**, the notebook check,
+`uv build`, and the distribution check. Removing code can move coverage in either direction, so
+read the number rather than assuming.
+
+## Risks
+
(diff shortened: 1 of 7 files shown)


HupBaHa (approved):
Thanks for handling both the verified presentation and the baseline-label fallback. The implementation and tests look good.

One non-blocking wording note: OME-820 does not make the field literally uniform because existing `False` rows are not backfilled. The stronger rationale is that the field temporarily has no trustworthy verification semantics, regardless of its value. This does not affect the implementation.

PR #611, opened 2026-08-16
feat(scoreboard): register DRACO, IFEval and HealthBench with revision identity

FilipBolt:
## Why

Scoreboard's deployed registry seeds only the legacy `hle` / `livetruth` / `livetruth-latest` demo entries. Verified against the live dev board today:

```
$ curl leaderboard.dev.screamingface.ai/v1/benchmarks
→ hle, livetruth, livetruth-latest        # no real benchmark
```

So the Leaderboard v1 catalogue renders no real board, and a Client submission for a real benchmark fails with `unknown_leaderboard` **after** a successful Engine evaluation — the failure lands at the end of a user's run, not the start. `OME-836/837/838` (merged 2026-08-14) flattened the Engine's identities to `draco`, `ifeval`, `healthbench-worst30`; Scoreboard has never heard of any of them.

Launch-blocking for 2026-08-17.

## What

Registers the three canonical benchmarks with their Engine `REVISION`, and gives the board the revision identity that makes those registrations meaningful.

**The revision was already arriving on every submission** — untyped, inside the free-form `metadata` dict (`packages/screamingface/.../leaderboards.py:333`). Unlike `OME-770`'s cost field there is no upstream producer to wait on; this promotes an existing wire value.

Six focused commits:

| | |
|---|---|
| `d0f14e05` | `Benchmark.revision` + `Score.benchmark_revision`, both nullable, migration `0004` |
| `e9d4d073` | resolve the revision typed-field-first, metadata as fallback |
| `aa18554b` | the revision joins the dedup identity hash |
| `edd2586b` | rank per `(spec_id, benchmark_revision)` |
| `df2e4e57` | expose it on the remaining read paths |
| `39765054` | seed the three benchmarks in the chart |

## Decisions worth reviewing

- **The metadata fallback is not a convenience.** Reading only the typed field would silently drop the revision for every deployed Client; rejecting the metadata copy would `422` every live submission. The copy is read, never mutated or stripped.
- **The revision joins `_content_hash` (owner call).** Without it, the same recipe at a second revision deduped into the first and the second result was discarded — which would have made the new partition unreachable, since that row never existed. **No backfill:** a resubmitted pre-existing recipe now creates a second row instead of deduping. Bounded, and documented beside the hash.
- **Ranking partitions rather than filters.** Filtering to the registered revision reads more literally as "not ranked together", but it *hides* scores instead of separating them, and would empty the retained legacy boards (which have no revision). Partitioning is also backward-safe — every existing row shares a `NULL` partition and groups exactly as before.
- **Legacy demo entries retained** (owner call), so this adds three boards rather than replacing three.
- **`ifeval` has no `dataset_url`** — its dataset is vendored in the Engine, so no public URL is authoritative. Left null rather than invented.

## Test plan

`run_gates.py scoreboard --base origin/main` after every step — ruff, ruff format, pyright, pytest --cov ≥80. Final suite **190 passed, 2 skipped**. RED-first throughout.

The seed config is deploy-time YAML that **no unit test executes**, so it was verified end to end instead of assumed:

1. `helm template` renders all six benchmarks into `SCOREBOARD_SEED_BENCHMARKS_JSON`.
2. That **rendered** payload — not a fixture — was fed through the real `load_benchmarks_json` + `seed_benchmarks`: 6 rows with correct revisions, and **6 rows again after seeding twice** (idempotency).
3. The three revisions were diffed **programmatically** against the Engine definitions (`ALL MATCH`), ruling out transcription drift.
4. `tortoise makemigrations` reports **"No changes detected"**, proving `0004` matches the models.

## Reviewer notes

- **One prior test changed, with owner approval:** `test_get_spec_history_returns_submissions_newest_first` pins the exact key set of the history payload, which this legitimately extends. It stays exact — it gained one member and was deliberately **not** loosened to a subset check. Two commits therefore ran `--skip-append-only`; the gate has no per-change acknowledgement, only a whole-check skip. Every other gate green, and `39765054`'s own test diff is 47 insertions / 0 deletions.
- **Not included:** portal rendering of the revision. The board can now legitimately show one spec twice with no visible explanation — worth a follow-up before revisions actually diverge in production.
- The last commit closes `OME-769`'s stale mirror and ledger (merged 2026-08-13, still reading in-progress in the repo), carried in separately per the precedent `OME-769` itself set.

Refs: OME-775

### DIFF
--- a/apps/scoreboard/charts/scoreboard/values.yaml
+++ b/apps/scoreboard/charts/scoreboard/values.yaml
@@ -101,7 +101,40 @@ seedBenchmarks:
   backoffLimit: 3
   ttlSecondsAfterFinished: 600
   annotations: {}
+  # The `revision` of each entry MUST match the Engine benchmark's computed REVISION exactly
+  # (apps/url4-cloud/src/url4_cloud/benchmarks/<name>/definition.py). A submission carries the
+  # Engine's value, and the board ranks per revision — a mismatch makes every real submission
+  # look incomparable to the registered board. Re-copy these whenever a benchmark's dataset or
+  # protocol revision changes; seeding is idempotent, so redeploying just moves the value.
   benchmarks:
+    # --- Engine benchmarks (OME-775) --------------------------------------------------
+    - id: draco
+      display_name: DRACO
+      description: >-
+        A 100-task DRACO reproduction with official score arithmetic. It uses the successor
+        Judge model, provider-default reasoning, mixed native/Tavily retrieval, and a
+        host-only approximation of the reference blocklist, so its scores are not
+        paper-identical.
+      dataset_url: https://huggingface.co/datasets/perplexity-ai/draco
+      revision: 1c58b3085912e304
+    - id: ifeval
+      display_name: IFEval
+      description: >-
+        The canonical 541-prompt instruction-following benchmark, graded by deterministic
+        strict and loose verification. Each Case invokes the Candidate exactly once.
+      # WHY no dataset_url: the IFEval dataset is vendored inside the Engine
+      # (url4_cloud.benchmarks.ifeval.vendor), so no single public URL is authoritative.
+      revision: 0b88a52b5f10a6d9
+    - id: healthbench-worst30
+      display_name: HealthBench Worst-30% Challenge
+      description: >-
+        The 157 hardest conversations from HealthBench Professional — the 30% that top
+        models score worst on. An AI judge grades each answer against a physician-written
+        rubric; safety mistakes subtract points, so per-case scores can be negative.
+      dataset_url: https://huggingface.co/datasets/openai/healthbench
+      revision: 6cd57aee171fbdc4
+    # --- Legacy demo entries, deliberately retained (OME-775 D2) -----------------------
+    # No Engine revision exists for these; they predate the Engine benchmark catalogue.
     - id: livetruth-latest
       display_name: News Livetruth Latest
       description: OpenMined Livetruth latest demo dataset
--- a/apps/scoreboard/src/scoreboard/routes/leaderboard.py
+++ b/apps/scoreboard/src/scoreboard/routes/leaderboard.py
@@ -34,6 +34,9 @@ class RankedLeaderboardEntry(BaseModel):
 
     rank: int
     spec_id: str
+    # WHY: the board ranks best-per-spec PER REVISION (OME-775), so one spec can legitimately
+    # appear twice. Without this field a client cannot tell why the two are not competing.
+    benchmark_revision: str | None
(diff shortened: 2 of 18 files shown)


FilipBolt:
## Update — the ranking design was wrong on the first pass, and local testing caught it

Pushed `1e4d1bda`. The PR description above described **partitioning only**, and claimed it meant "scores measured against incompatible dataset/protocol revisions are not ranked against each other." **That claim was false**, and I want it on the record rather than quietly amended.

Partitioning stops one revision *displacing* another in the best-per-spec collapse. It does not stop them being *ranked together* — the outer query still orders every surviving row into one accuracy ranking.

Demonstrated against a running server with the real Client, not by review:

```
draco registered at: REV-CURRENT

rank   spec_id             accuracy   benchmark_revision   comparable?
#1     fusion/old-winner   0.90       REV-OBSOLETE         NO  <-- stale
#2     fusion/new-entry    0.80       REV-CURRENT          YES
```

The board's top row was a score that was never measured against the benchmark the board claims to rank. A reader concludes `old-winner` is better. It isn't.

**Every unit test passed throughout** — each asserted its own mechanism (the partition works; the filter works) rather than the end-to-end guarantee the acceptance criterion actually states. That is the real lesson here: *"rows are separated"* is not *"results are not ranked together"*, and a criterion phrased about presentation needs a test that reads the presented output.

### The fix

The board now also **filters to the benchmark's registered revision**. Both mechanisms are needed — the partition keeps each revision's best intact, the filter decides which revision the board is about.

A benchmark with **no** registered revision filters nothing, so the retained legacy demo boards are untouched. Once a benchmark declares a revision, rows that cannot be asserted comparable to it — including `NULL` rows predating this column — no longer rank. They stay stored and stay visible in per-spec history.

Re-verified live after the fix: draco ranks only `REV-CURRENT`; the legacy board still shows everything. Three new tests cover the stale-revision exclusion, the no-registered-revision passthrough, and the pre-revision `NULL` case. Suite now **193 passed, 2 skipped**.

### Also found by the same local run

- **The Python Client cannot see the revision.** The server returns `benchmark_revision` on both the leaderboard entry and the score — verified in the raw JSON — but the Client's read models don't parse it. Client-side gap, `packages/screamingface`. Assigned to `OME-771` along with portal rendering (owner decision: correctness here, presentation there).
- **A false alarm worth recording** so nobody re-investigates: a first run appeared to show a second revision deduping into the first. The test was wrong, not the code — the Client sends `Idempotency-Key: candidate_result.run_id`, which short-circuits ahead of `content_hash`, and both submissions shared a `run_id`. Correct behaviour; real runs always carry distinct run ids.

Full record in the ledger, including the complete local verification matrix.

PR #624, opened 2026-08-17
feat(scoreboard): rename the verification field to verified_by_screamingface

FilipBolt:
## The contract is broken on `main` right now

`OME-858` (#622, merged 14:42) renamed this field in the Client and made its decoder strict. Scoreboard still emits `verified_by_openmined`, so the two disagree and every leaderboard read fails. Reproduced against `origin/main` before starting:

```
sf.leaderboards.get('draco')
→ LeaderboardError: Invalid Scoreboard Leaderboard response:
  Leaderboard entry verified_by_screamingface must be a boolean
```

This blocks `OME-402` — the leaderboard notebook, which is the path testers submit through.

## What this does

Renames the field across the model, response schemas, routes, store projections, portal JS and tests (45 occurrences, 17 files), and adds migration `0005` using `RenameField`.

**`0001_initial.py` is deliberately untouched.** It records applied history; editing it would desynchronise every existing database from its recorded state.

Semantics are unchanged — no badge, filter, ranking rule or reproduction claim is introduced, and the field stays server-owned. Defining a trustworthy signal remains `OME-821`.

## Verified beyond the unit suite

Two things the suite structurally cannot cover, each proven directly:

**1 · Existing data survives the rename.** Built the schema at `0004` with `0005` held back, inserted a `false` row and a `true` row through raw SQL, then applied `0005`:

```
new-row|1        old column present: 0
old-row|0
```

Both kept their original values. This is `OME-820`'s D5 invariant — a drop-and-recreate would have defaulted `old-row` to `true` and published a claim about a run nobody checked.

**2 · The contract is repaired.** Real `sf.Client` against a real Scoreboard: `list()` OK, `submit()` OK, `get()` OK, `get_score()` OK — all four fail on `main` today. No response emits the old key on any read endpoint, and a submission claiming the server-owned field is still rejected **422**.

## Two changes beyond a plain rename

**The migration-backfill guard now checks BOTH names.** `test_no_migration_backfills_the_verified_column` scans *migration sources* for the field. Pre-rename migrations keep the old name, so renaming the guard alone would have left it blind to half the history. Found by reading the test — a mechanical sed would have silently degraded it.

**Migration `0005` declares two parents.** `main` carried two sibling `0004` migrations, both claiming `0003`, from `OME-323` and `OME-775` merging minutes apart. Depending on both converges that branch so the next migration has one unambiguous parent. Not requested by the ticket; free fix for a latent hazard.

## Test plan

`run_gates.py scoreboard` — ruff check, ruff format, pyright, pytest --cov all green: **253 passed, 2 skipped**, portal JS 14 passed. `makemigrations` reports **"No changes detected"** (no model/migration drift).

Ran with `--skip-append-only`: a rename cannot avoid touching prior tests. That is a rule-5 decision, approved by the owner after the diff was shown to be purely the identifier — every changed line is identical once the field name is normalised, with the guard hardening the only substantive addition.

Refs: OME-865

### DIFF
--- a/apps/scoreboard/DEPLOYMENT.md
+++ b/apps/scoreboard/DEPLOYMENT.md
@@ -210,6 +210,28 @@ helm rollback scoreboard <revision> --namespace scoreboard --wait
 
 Helm rollback does not roll back database schema migrations. Keep migrations forward-compatible where possible.
 
+### Breaking migrations and multi-replica rollouts
+
+`values-prod.yaml` sets `replicaCount: 3`, and the migration Job is a `pre-upgrade` hook — it
+finishes **before** the Deployment rolls. So during a production upgrade the old pods keep serving
+against the **new** schema until the rollout completes. A migration that renames or drops a column
+those pods still query makes them fail for that window, and `/healthz` does not touch the database,
+so readiness stays green and Kubernetes keeps them in the Service. Combined with the line above —
+rollback does not revert the schema — a breaking migration is not recoverable by `helm rollback`.
+
+**Before cutting a `scoreboard-v*` tag, check whether any unreleased migration renames or drops a
+column.** If one does, pick one:
+
+- **maintenance window** — scale to 0, migrate, scale back up. Simplest, and fine while the service
+  has no users.
+- **expand/contract** — add the new column, dual-write, backfill, switch reads, drop the old column
+  in a later release. Three deploys, no downtime. Use this once the board has real users.
+
+`0005_auto_20260817_1520` (OME-865, renaming `verified_by_openmined` to
+`verified_by_screamingface`) is the first migration in this category. It is safe on dev, which runs
+a single replica, and was deliberately shipped as a plain rename because production had no users at
+the time. Re-check that assumption before releasing it.
+
 ## Troubleshooting
 
 Inspect resources:
--- a/apps/scoreboard/portal/benchmark.js
+++ b/apps/scoreboard/portal/benchmark.js
@@ -208,7 +208,7 @@
     }
 
     var best = bestAccuracy(entries);
-    // OME-820: the "Verified rows" stat is gone, not relabelled. verified_by_openmined
+    // OME-820: the "Verified rows" stat is gone, not relabelled. verified_by_screamingface
     // now carries no trustworthy verification semantics — nothing re-runs submissions
     // and nothing attests where a run executed — so counting it measures nothing.
     //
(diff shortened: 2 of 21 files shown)


FilipBolt:
Good catch, and the premise check turned up more than the finding claimed. Every point verified:

| Claim | Verified |
|---|---|
| Migration runs as a `pre-upgrade` hook while old pods serve | `charts/scoreboard/templates/job-migrate.yaml:9` |
| Three replicas | `values-prod.yaml: replicaCount: 3` |
| Rollback leaves the schema renamed | `DEPLOYMENT.md:211` |

**Worse than reported:** `/healthz` returns a static `{"status": "ok"}` without touching the database, so a pod broken by the rename still passes readiness and **stays in the Service** while failing real queries. Kubernetes has no way to notice.

**And the ticket's own premise was wrong.** OME-865 says *"this is a clean pre-release rename"* — but `scoreboard.screamingface.ai` is live and serving 4 rows created 2026-06-02. Not pre-release.

## Why it still ships as a plain rename

Owner decision, taken with all of the above on the table:

- **Production has no users** (owner-confirmed). Its 4 rows are June demo data on `hle`/`livetruth`, benchmarks nothing queries, and there has never been a `scoreboard-v*` release tag.
- **Merging does not touch production.** Prod releases only on that tag; dev auto-builds from `main`. This reaches dev only, where `replicaCount: 1` makes the window one pod for seconds — against a board that is *currently* fully broken for the Client (`OME-858` renamed the field client-side, so every read fails strict decoding today). The rename strictly improves it.
- **Expand/contract is three deploys** plus dual-write machinery, to protect four stale demo rows and a rollout window on a service with no users.

## The cheaper fix you suggested was tested, and it does not work

"Preserve the old physical column via ORM mapping" was the better option if it held — one deploy, no migration, no rollout risk. Tortoise supports `source_field`, so I tried it. It fails:

```
ValidationError: verified_by_screamingface
  Input should be a valid boolean, input_value='verified_by_screamingface'
```

`_build_leaderboard_query` is hand-written pypika and bypasses the ORM's field-to-column mapping, so it emits the Python name as a string literal instead of selecting the column. Salvageable by hardcoding the physical column in that one query, but that leaves a permanent Python/DB name mismatch **and** Tortoise's autodetector keeps proposing a `RenameField` on every future `makemigrations`. Recurring friction for a name nobody sees.

## What changed instead

The risk is deferred, not dismissed, and written where a releaser will meet it:

- **`DEPLOYMENT.md`** — new *"Breaking migrations and multi-replica rollouts"* section under *Upgrade And Rollback*, explaining the hook/replica/readiness interaction and naming the two safe options (maintenance window, or expand/contract) with the trigger to re-check.
- **The migration** — an `AIDEV-NOTE` stating it is breaking for a multi-replica rollout, why it shipped anyway, and that **the no-users assumption must be re-checked before cutting a `scoreboard-v*` tag**.

If production gains users before that tag, expand/contract is the right call and the note says so.

PR #626, opened 2026-08-18
feat: replace binary accuracy submissions with benchmark-native Leaderboard scores

khoaguin:
Closes [OME-866](https://linear.app/openmined/issue/OME-866/replace-binary-accuracy-submissions-with-benchmark-native-leaderboard) · also delivers [OME-870](https://linear.app/openmined/issue/OME-870/add-benchmark-explainer-infographics-to-the-example-notebooks) (the tester-facing leaderboard story for tomorrow's push). Cross-stack: `apps/scoreboard` + `packages/screamingface` land together because the wire contract flips atomically (the ticket forbids dual wire formats).

### Before
- Trigger: a tester evaluates DRACO or HealthBench with the Client and publishes the result.
- Today: the submission adapter rejects it before it ever reaches the network — DRACO's fractional rubric scores fail the "every Case grade must be binary" check, HealthBench's negative scores fail the 0..1 range check. Only IFEval can reach the leaderboard. If a submission does land, the returned value floods the notebook with a several-thousand-character repr, the boards say "Accuracy" and render percentages, and nothing explains why a HealthBench score of `-1.143` isn't a bug.
- Cost: two of our three launch benchmarks cannot appear on the public board at all, the day before the tester push — and what does appear is easy to misread.

### After
- Same trigger.
- Now: the Client submits the Engine's `CandidateResult.score` exactly as graded (`0.399`, `-1.143`, `0.5`); the Scoreboard stores, dedups and ranks it without recomputing anything. `submit()` returns a brand card ("Score published") with a **View on the Leaderboard** deep link to the exact spec page on the originating board. Portal and notebook widget say **Score**, render plain benchmark-native numbers, and normalize bars over the on-screen min..max so negative boards read sensibly. Each benchmark notebook opens with an sf-dark infographic explaining dataset → grading → score math. `just stack-up` seeds the local scoreboard from the Engine's own benchmark registry, so Run All works against a fresh local stack with real revisions.
- Win: every registered benchmark round-trips to the public leaderboard unchanged, and a tester can predict what number they'll see — and why — before spending a credit.

### Don't regress
- An unscored result (`score=None`) still cannot be submitted; NaN/±inf are rejected on both sides (now explicitly — previously only as a side effect of the deleted range check).
- Higher score still always ranks higher within a benchmark; revisions still never rank together.
- Existing IFEval rows survive the migration (columns renamed in place, verified on a populated DB).
- No legacy aliases: the old `accuracy` wire field is a 422.

## Root cause
The Client → Scoreboard contract hard-coded IFEval's shape: `accuracy = correct/total` with a 0..1 range, a binary-grade check, a ±0.01 server cross-check, and a dedup hash derived from the counts. Nothing ever *recomputed* the score — the extra machinery only policed it — so the core fix is mostly deletion plus a rename.

## Implementation

**Scoreboard (OME-866)**
- `ScoreSubmission.score`: strict float, `allow_inf_nan=False` — string/bool/null/NaN/inf are 422; JSON integer literals accepted as their float value. `correct_questions` becomes optional (binary-era detail); `total_questions` (case count) stays.
- Route ±0.01 tolerance check deleted, not replaced — the Scoreboard never second-guesses the Engine. `content_hash` keys on the exact submitted score.
- Migration `0006`: renames `accuracy→score` on `scores` + `baselines` in place, rebuilds the ranking index. **Breaking for multi-replica rollout** — see the note in the migration + DEPLOYMENT.md before cutting a `scoreboard-v*` tag.

**Portal (OME-866)**
- "Accuracy" → "Score"; `formatScore` renders plain numbers (percent stays only for genuine shares like the frontier's `open_share`); `barWidth(score, min, max)` with floor `min(0, lowest)` — 0..1 boards keep their absolute origin, negative boards shift instead of emitting negative CSS widths.
- Masthead copy: "Results you can **reproduce**, not just read."
- Note: the portal-side infographics ([OME-872](https://linear.app/openmined/issue/OME-872/render-benchmark-explainer-infographics-on-the-portal-benchmark-page)) were implemented and then **reverted in-branch** at owner request — the leaderboard is a public record of the best models for a wide audience; the explainers are documentation and stay in the notebooks.

**Client SDK (OME-866)**
- `_accuracy_result` → `_score_value`: the only client-side gates are *is a CandidateResult*, *has a score*, *score is finite* (`math.isfinite` replaces the NaN rejection the deleted range check provided by accident).
- Public dataclasses rename `accuracy→score`; `LeaderboardScore` gains a one-line `__repr__`, an SFDS `_repr_html_` card, and a `scoreboard_url` stamped at decode time that powers the deep link (localhost locally, the deployed board in production — never a guess).
- Widget renders plain numbers with clamped normalized bars.

**Notebooks & local stack (OME-866 + OME-870)**
- 06/07/08/09 end with an opt-in (`PUBLISH_RESULT = False`) submit cell; 06/07/08 open with a hand-authored sf-dark infographic whose facts mirror the Engine scorers (IFEval prompt-level strict accuracy 0..1 · DRACO clamped weighted rubric mean 0..1, the paper's own formula · HealthBench worst-30 unclipped mean, negative is normal).
- `just stack-up` seeds `draco`/`ifeval`/`healthbench-worst30` with real revisions straight from `url4_cloud.benchmarks.builtins` (no hardcoded list to drift); `just jupyter` defaults both the engine *and* scoreboard URLs to the local stack so submit cells never silently publish to the deployed dev board.
- `examples/helpers.py` gains `load_candidate_result()` — rebuild a submit-ready result from a saved `report.export()` JSON (interim, lossy; the real `sf.Report.load()` is a flagged follow-up).

Design deltas vs the OME-866 ticket (kept `total_questions`, deferred typed `metrics` + score-presentation metadata to a follow-up, exact-score dedup identity, explicit NaN guards, interim negative-score rendering rule, benchmark-native baseline scale) are recorded as a comment on OME-866 for @keelancj to confirm.

## Test plan
- Cross-stack contract tests: the real Client payload per benchmark (IFEval `0.5`, DRACO `0.399`, HealthBench `-1.143`) submits, round-trips byte-identical, ranks descending with negative/mixed ranges. Edge matrix both sides: missing/null score, NaN, ±inf, bool, string, integer literal.
- Portal Node tests: negative/mixed/all-equal/degenerate bar ranges. Widget tests pin no negative CSS widths, no ×100 rendering, the card's deep link, and the summary repr.
- Gates: `run_gates.py scoreboard` (Python + Node) and `run_gates.py screamingface` (cov ≥95, deterministic notebooks, build + distribution) all green locally. Migration chain 0001→0006 applied on fresh and populated sqlite; a row written at 0005 keeps its value under `score`.
- Live E2E on a local stack: a real saved HealthBench result (`-1.1429`) submitted through the actual client path — 201, idempotent resubmit (same id), byte-identical round-trip, ranked correctly on portal and widget.
- Reviewer note: prior tests were renamed/rewritten as part of this sanctioned contract change (`run_gates.py --skip-append-only`, recorded in `docs/work/2026-08-18-OME-866-benchmark-native-scores.md`); three tolerance-era tests now pin the inverse invariant instead of being deleted. The pre-push hook has no sanctioned-change path yet — flagged as a possible `repo` follow-up, deliberately not changed here.

Refs: [OME-866](https://linear.app/openmined/issue/OME-866/replace-binary-accuracy-submissions-with-benchmark-native-leaderboard), [OME-870](https://linear.app/openmined/issue/OME-870/add-benchmark-explainer-infographics-to-the-example-notebooks)


FilipBolt:
Review notes for current head `bf7f12f` — I recommend addressing these before merge.

1. **P1 — Valid partial reports cannot be reopened.** `load_candidate_result()` reconstructs cases without carrying over `status`, `refusal`, or failures (`packages/screamingface/examples/helpers.py`, around lines 105–114). A refused case with a numeric grade is inferred as `scored` but has no output, while a failed case loses the failure required by the `CaseResult` invariant. I reproduced this from a real `Report.export()` payload:

   ```text
   ValueError: Case Result status 'scored' requires output and a numeric grade and cannot carry refusal or failures
   ```

   Please preserve enough outcome information to construct valid cases and add coverage for exported reports containing refused and failed cases.

2. **P2 — The deployment smoke command returns 422.** The payload in `apps/scoreboard/DEPLOYMENT.md` sends `client_name`, `client_version`, and `client_platform` as top-level fields, but `ScoreSubmission` uses `extra="forbid"` and only accepts this information under the nested `client` field. The payload should use:

   ```json
   "client": {"name": "curl", "version": "0.1.0", "platform": "k3s"}
   ```

3. **P2 — The live public docs still describe the deleted API.** The following pages use `entry.accuracy` / `score.accuracy`, document binary correctness validation, and describe `correct_questions` as mandatory:

   - `public-docs/src/pages/sf-client/guides/LeaderboardsPage.vue`
   - `public-docs/src/pages/sf-client/api/LeaderboardsPage.vue`
   - `public-docs/src/pages/learn/LeaderboardPage.vue`

   Since this PR deliberately provides no legacy alias, the copied examples will fail immediately after the contract flip. These pages should move to `score` and explain benchmark-native scoring.

Validation performed against `bf7f12f`:

- 206 focused Scoreboard tests passed, 2 skipped.
- 89 SDK/UI tests passed, 3 skipped.
- All 19 portal Node tests passed.
- Fresh SQLite migration chain through `0006` passed with the expected `score` schema.
- `git diff --check` passed.

Operational note: migration `0006` is a breaking column rename and cannot be safely rolling-deployed against the documented three replicas. The PR acknowledges this; the release should use a maintenance window or an expand/contract rollout.

FilipBolt (commented):
Second pass on `bf7f12f`, on the presentation layer this time. These are additional to my notes above and don't overlap them. I've deliberately left out the three items already recorded as accepted deltas on OME-866 (dedup rehash, dropping `correct_questions`, the interim negative-bar rule) since those are settled.

1. **P1 — The IFEval description is now Markdown, and the portal renders descriptions as plain text.** `apps/url4-cloud/src/url4_cloud/benchmarks/ifeval/definition.py` rewrites the description to `[IFEval](https://arxiv.org/abs/2311.07911) contains 541 …`. There is no Markdown renderer anywhere in the portal: `portal/benchmark.js:334` does `descNode.textContent = b.description`, and the homepage subtitle goes through `el()` at `portal/main.js:89-93`, which also sets `textContent`. The board will show the literal brackets and URL.

   DRACO and HealthBench descriptions are plain prose, so IFEval would be the only affected one. Local stacks hit it immediately now that `just stack-up` seeds from the Engine registry; deployed boards seed from `charts/scoreboard/values.yaml`, so it surfaces there once someone syncs the text across — and that file would then disagree with the definition.

   Two smaller things in the same hunk: the rewrite drops the official-keys note and the key 2785 patch, and it is the only change in this PR outside `apps/scoreboard` + `packages/screamingface`. Could it move to its own PR?

2. **P2 — The same score renders at two different precisions in one notebook.** Three renderers, two rules:

   | value | report card + submit receipt | board widget | portal |
   |---|---|---|---|
   | `-1.1429` | **-1.1429** | **-1.143** | **-1.143** |
   | `0.8333333` | **0.833333** | **0.8333** | **0.8333** |

   `_ui/report_view.py::_score_text` uses `f"{value:g}"` (6 significant digits) and `_ui/score_view.py` imports it. `_ui/leaderboard_view.py::_score_text` independently uses `f"{value:.4g}"`, and `portal/main.js::formatScore` uses `toPrecision(4)`. The comment on the `report_view` helper states the opposite of what happens: *"matching the submit receipt (score_view) so the same figure never renders two ways in one notebook."* True for `score_view`, not for the board.

   `-1.1429` is this PR's own live E2E value, so a tester sees `-1.1429` in the report and on the "Score published" card, then `-1.143` on the board. Given the PR's goal is that a tester can predict the number they will see, one shared helper would be worth it.

3. **P3 — `score > 0` still gates the hero highlight.** `_ui/report_view.py:277`: `scored = candidate.score is not None and float(candidate.score) > 0`. On HealthBench worst-30 every score is negative, so the score cell never highlights. That is a leftover 0..1 assumption of exactly the kind this PR removes elsewhere — negative is a normal result here, not a null outcome.

4. **P3 — `DEPLOYMENT.md` isn't updated for the second breaking migration.** The "Breaking migrations and multi-replica rollouts" section still reads *"`0005_auto_20260817_1520` … is the first migration in this category."* `0006` belongs in it now. A releaser following that section's own instruction — check whether any unreleased migration renames or drops a column — finds 0005 documented and 0006 not. One sentence. (Complements the operational note in my earlier comment; the rollout risk itself is already acknowledged in the PR.)

5. **Release ordering, not a code change.** `screamingface` 0.1.1 is live on PyPI and the release-please PR (#592) for 0.2.0 was last recut before this branch existed. Since the old `accuracy` field is now a 422 with no alias, 0.2.0 has to be cut from a commit that **includes** this PR, otherwise `pip install screamingface` hands testers a client that 422s on submit against the flipped board. Worth sequencing explicitly given the tester push.

Things I checked that hold up, for the record: the benchmark revision hashes are unaffected (the description is not among the hash inputs at `ifeval/definition.py:32-45`, and `CANDIDATE_RESULT_SCHEMA` is untouched), so the registered revisions on dev stay valid; migration `0006` chains correctly off `0005` and the index it drops genuinely exists from `0001_initial.py:83`, rebuilt in the right order around the rename; every `accuracy` reference in the hand-written pypika leaderboard query was renamed, which is the trap that killed the `source_field` approach in OME-865; the OME-775 registered-revision filter and partitioning survive intact; the frontier stat only consumes `open_share` and the holder label, never the raw score, so negative boards don't break it; `barWidth`'s `min(0, lowest)` floor is sound in both directions and clamped; and the score card escapes every interpolation including the href.

FilipBolt (approved):
Approving — the functional issues from both review passes are addressed in `99ab98c`, the updated checks are green, and I re-verified the loader, deployment payload, public-docs contract, and presentation changes.

Non-blocking bookkeeping follow-up: the commit message and `docs/work/2026-08-18-OME-866-benchmark-native-scores.md` say `packages/screamingface/tests/test_examples_helpers.py` was added, but that file is not present in the PR. The loader fix itself works against a mixed refused/failed export; please either add the claimed regression test or correct the work ledger before merge.

The release-ordering and non-rolling migration rollout remain owner actions as already documented.

PR #631, opened 2026-08-18
feat(scoreboard): adopt the leaderboard-mvp landing copy and UI

FilipBolt:
Closes [OME-874](https://linear.app/openmined/issue/OME-874/replicate-the-leaderboard-mvp-landing-copy-and-ui-on-the-portal). Brings the public landing page to the brand mockup at https://brand.screamingface.ai/leaderboard-mvp/ — copy and UI 

## Before / After

**Before** — a visitor lands on the public board and sees a page describing what we ship but never saying what the board is *for*. The mockup's vocabulary (fusion, reproducible, solo model) is undefined, so a negative HealthBench score or a submitter-less baseline row reads as a bug. Half the mockup's copy was deliberately withheld by `OME-839` because it claimed things we cannot do.

**After** — the page states its purpose and defines every term it uses. The hero, glossary, docs CTA and the `Focus` / `Best reproducible` columns match the mockup; the read-this-first note carries the two of its three sentences that are true on our board.

## The copy decision, stated plainly

`OME-839` adopted only the mockup copy that was true, under a §4 invariant: *no portal copy may claim reproduction, verification status, or cost until the owning ticket lands.* This PR originally suspended that invariant on the owner's call, taking the mockup's note verbatim.

**Review (@HupBaHa) showed that wording is not implementable here, and it was reverted.** Verified against the query rather than the wording: `_build_leaderboard_query` selects `verified_by_screamingface` but never filters on it and orders by `score DESC`, so an unverified `0.99` outranks a verified `0.40`.

The reason is structural rather than editorial. **The mockup's note is coherent only because its default view is filtered to reproducible rows** — "the top is the best verified result" is true *of a filtered board*, and "toggle on self-reported runs" names the control doing the filtering. On an unfiltered board the first is false and the second points at nothing. So two sentences are deleted rather than reworded; no phrasing makes a missing control present.

| | Mockup sentence | Outcome |
|---|---|---|
| S1 | *"By default, the leaderboard only shows results we've reproduced ourselves."* | **Deleted** — describes a default filter that does not exist |
| S2 | *"The top of each leaderboard is the best **verified** result: the current SOTA."* | **One word**: `verified` → `submitted` |
| S3 | *"Toggle on self-reported runs…"* | **Deleted** — names a control that exists nowhere (`OME-771` is Blocked) |

Net deviation from the brand copy: **one word changed, two sentences removed, nothing invented.** Both pages carry the same note; `benchmark.html` has no note in the mockup at all, and the owner's instruction was scoped to the landing page, so that page's wording is ours either way.

**The glossary is untouched and still describes machinery we do not have** — `Reproducible` means "ran on shared compute… anyone can re-run it and get the same score". That is deliberate: it defines what the term *means*, and no row is presented as being in that category. Re-run verification is [OME-414](https://linear.app/openmined/issue/OME-414), the real verified signal [OME-821](https://linear.app/openmined/issue/OME-821), the pool toggle [OME-771](https://linear.app/openmined/issue/OME-771). None has landed.

`models/score.py` keeps its original invariant, plus a note recording that forward-looking wording was tried and why it failed — so the next attempt reaches for the filter rather than the wording. **The copy and the pool filter are one change, not two:** the full mockup wording returns with `OME-771`.

> **Owner action:** S1 and S3 were copy @IrinaMBejan approved verbatim on 2026-08-18. Removing them deviates from that instruction, and was done because review showed they describe behaviour the board does not have. Please confirm before merge.

## Implementation

**Backend** — `Benchmark.focus`, a nullable 120-char editorial line, exposed on `/v1/benchmarks` and seeded through `values.yaml`. Migration `0007` is a plain `AddField` and is **safe for a rolling multi-replica rollout**, unlike `0005`/`0006`.

**Portal** — hero, glossary, note, CTA, table columns. The hero sets the mark as the lowercase "o": it is an ``, never the emoji, because `style.css` records the system's kerning audit (the emoji's glyph box word-spaces the letters and dips below baseline). The asset is vendored app-local, so the public board makes no off-origin request.

**`Best reproducible` costs no extra request** — it reads the payload the page already fetched to count rows. It counts **submissions only, never baselines**: a baseline is an imported reference number with no submitter, and ranking one here would present an outside board's figure as the best result on this one. Pinned by a test.

**A DRY fix, not in the original plan.** `routes/leaderboard.py` carried a second hand-written `BenchmarkSchema` projection. Adding `focus` broke that copy while the store's stayed correct — the identical shape that took `main` red in `OME-852`. There is now one mapper used by both.

## Deliberate divergences from the mockup

Three, all because its data is mock and ours is real:

| | Mockup | Here | Why |
|---|---|---|---|
| Footer 2nd cell | `leaderboard.screamingface.ai · MVP preview · mock data` | `scoreboard.screamingface.ai` | That hostname has **no DNS record** (re-checked 2026-08-18); our data is not mock |
| Stats strip | absent | kept | Three real live counts the mockup predates |
| Dataset column | absent | kept | Links real published JSONL that `data.html` serves |

Also: `benchmarks` targets `index.html#benchmarks`, not the mockup's bare `benchmark.html`, which errors without an `id`. And the mockup renders `Best reproducible` as a **percentage** — not followed, because `OME-866` made scores benchmark-native this morning and HealthBench would print `-114.3%`.

## Test plan

- **Backend, RED first:** `focus` persists through seeding and is `None` when absent; `/v1/benchmarks` serialises `null` rather than omitting the key.
- **Portal JS:** six cases on `bestEntryScore` — empty board, `null` vs a real `0` (0 is a real score on a benchmark that goes negative), highest-not-first, an all-negative board, baselines never counting, malformed entries skipped.
- **Static:** the hero mark is served as `image/png`, and the `.o-mark` `src` resolves to a path this app serves — asserted behaviourally, so a hotlink cannot satisfy it.
- **New opt-in smoke lane** (`pytest -m smoke`) — drift alarms comparing our vendored glossary, note copy and mark digest against the live brand site. **Excluded from CI by design:** an editor at another company must not be able to turn our build red.
- **Gates:** ruff ✓ · ruff format ✓ · pyright ✓ · pytest --cov **323 passed, 2 skipped** ✓ · portal JS **25 passed** ✓. Migration chain `0001→0007` applied on fresh and populated sqlite.
- **Security review** of the branch: no vulnerabilities. It did surface the `index.html`/`benchmark.html` contradiction, which is fixed here, and a **pre-existing** issue out of scope for this PR — `POST /v1/scores` is unauthenticated in the shipped prod config (`authMode: disabled`, client-controlled `submitted_by`). Filing separately.

## Reviewer notes

- **`--skip-append-only`** was used, owner-approved under sdlc rule 5. Exactly **one** prior assertion changed — the hero-copy string this PR replaces — and it was rewritten structurally so the next copy edit will not break it again. The JS test file is flagged by the gate only because it cannot parse JavaScript; `git diff` shows 0 removed lines there.
- **No internal references in served files.** The portal ships unminified and `register_portal` mounts `portal/` whole, so everything in it is public. My first pass put ticket ids and an `AIDEV-NOTE` reading *"none of this is implemented yet"* into `index.html` — visible in View Source directly under copy claiming reproducibility. All removed.

  **Corrected after review:** an earlier revision of this PR body claimed zero internal references, and that was wrong. `assets/mark/PROVENANCE.md` — a file I created *inside* the served tree — shipped a ticket id and a `.claude/` path, because the audit grepped only `*.html`/`*.js`/`*.css`. Fixed in `72ba06f2`, and the real remedy is the guard that came with it: a test fetches every markdown file under `portal/` through the app and fails on `OME-`, `.claude/` or `worktrees/`. Verified by reintroducing the leak and watching it fail.

  Scoped to markdown on purpose — `benchmark.js`, `main.js`, `leaderboard-logic.js` and `portal.css` already carry **~42** such references predating this branch, and a guard that fails on arrival gets skipped rather than fixed. Cleaning those needs its own ticket.
- **`values.yaml` does not propagate.** Deployed environments keep their own values file, so the three `focus` lines need a platform-team sync — same caveat as `OME-775`'s revisions. cc @bitsofsteve
- **`Focus` copy is placeholder**, not brand-approved: *Research reports with citations* / *Instruction following* / *Clinical safety, hardest cases*. Editable in config, not code.
- The name cell keeps its `description` subtitle **and** gains `Focus`, per owner decision. This leaves `OME-768`'s open question (what the catalogue subtitle should be) still open.

Refs: [OME-874](https://linear.app/openmined/issue/OME-874), supersedes [OME-871](https://linear.app/openmined/issue/OME-871)


HupBaHa:
I think this should block before merge.

I verified the underlying behavior, not just the wording. The leaderboard ranks by score, not by `verified_by_screamingface`: a `verified_by_screamingface=false` row with score `0.99` ranks above a `true` row with score `0.40`.

So this copy is contradictory on the current unfiltered board:

> The top of this board is the best reproducible result: the current SOTA.
> Self-reported runs can rank above it...

If self-reported runs can rank above it, then the top of the board is not necessarily the best reproducible/current SOTA. Please remove or reword the “top/current SOTA” claim until the real pool filter lands.

Same issue on the landing page: it tells users to “Toggle on self-reported runs”, but that toggle does not exist on this page yet. That is visible broken UX, not just a nit.

FilipBolt:
Both correct — thanks, this was the right call to block on. Fixed in `b2f7556e`.

I verified the ranking against the query rather than the wording, and it's exactly as you describe: `_build_leaderboard_query` selects `verified_by_screamingface` into the projection but never filters on it, and orders by `score DESC`. An unverified `0.99` outranks a verified `0.40`.

**The contradiction you found in `benchmark.html` was my drafting error, not brand copy.** The mockup has no note box on its benchmark page at all — 0 `class="note"`, 0 occurrences of "Read this first". I wrote that sentence two days ago when we chose to reword that page's disclaimer rather than delete it, and I ported the mockup's sentence structure without noticing it depended on a filter.

## Root cause, which is structural rather than editorial

The mockup's note is coherent **only because its default view is filtered to reproducible rows**. "The top of each leaderboard is the best verified result" is a true statement *about a filtered board*, and "toggle on self-reported runs" names the control doing the filtering. Ported onto an unfiltered board, the first sentence is false and the second points at nothing.

That's why both are deleted rather than reworded — no phrasing makes a missing control present. What ships:

| | Mockup sentence | Outcome |
|---|---|---|
| S1 | *"By default, the leaderboard only shows results we've reproduced ourselves."* | **Deleted** — describes a default filter that does not exist |
| S2 | *"The top of each leaderboard is the best **verified** result: the current SOTA."* | **One word**: `verified` → `submitted` |
| S3 | *"Toggle on self-reported runs…"* | **Deleted** — names a control that exists nowhere (`OME-771` is Blocked) |

Both pages now carry the same two sentences:

> The top of each leaderboard is the best submitted result: the current SOTA. Every row keeps its URL4 expression, so you can rerun any claim before trusting it.

## One partial pushback

Your ask was to "remove or reword the top/current SOTA claim". Removing it entirely would have cost brand copy we didn't need to lose — one word makes it true and keeps the SOTA framing the gold-row treatment already implies. So I reworded rather than removed. Say if you read that differently.

## What is deliberately unchanged

The **glossary** (`Reproducible` / `Unverified` / `Solo model`) stays verbatim. It defines what those terms *mean*; it doesn't assert that any row is currently in the reproducible category — and it reads coherently beside a note that never calls a row verified. That vocabulary is the owner's and I didn't want to touch it beyond what your findings required.

`models/score.py` keeps its original invariant, plus a note recording that the forward-looking wording was tried and why it failed — so the next person reaches for the filter rather than the wording.

## Follow-up

The copy and the pool filter are **one change, not two**. The full mockup wording — default-filtered view, the toggle, "best verified result" — returns with `OME-771`, which is blocked on `OME-821` → `OME-414`.

Also flagging for the owner separately: S1 and S3 were copy she approved verbatim on 2026-08-18, so removing them is a deviation from that instruction, made because the review showed they describe behaviour the board doesn't have.

HupBaHa (approved):
Approved from code-review side.

I rechecked `b2f7556e`: the two blocking copy/UX issues are fixed. The notes now say “best submitted result” instead of claiming verified/reproducible top, and the landing page no longer points users at a missing self-reported toggle.

Verified locally:
- `uv run pytest -q` → 324 passed, 2 skipped, 3 deselected
- `node --test tests/portal/leaderboard-logic.test.js` → 25 passed

CI is green and the PR is mergeable clean. The remaining owner-copy confirmation mentioned in the work ledger is a product sign-off question, not a code-review blocker from me.

PR #633, opened 2026-08-18
feat: run any OpenRouter model — dynamic admission at preflight

khoaguin:
Run **any** model from OpenRouter's public catalog with nothing but an OpenRouter key — no env edits, no url4.toml entry, no restarts. When a run asks for an OpenRouter model the Engine doesn't know, the Engine asks the AI Gateway "does this model actually exist on OpenRouter?". The gateway checks OpenRouter's public model list: real → the model is admitted on the fly and the run just works; typo / provider off / no key → a clear refusal **before any money is spent**, naming the exact knob to turn.

openrouter-dynamic-admission

Epic: OME-878 · sub-issues: OME-879 (aigateway) · OME-880 (url4-cloud) · OME-881 (py-screamingface)

### Before
- Trigger: a researcher points `sf.evaluate()` at a real OpenRouter model that isn't in our curated ~117-model list.
- Today: they must edit `AIGW_OPENROUTER_DEFAULT_MODELS`, add a url4.toml entry, and restart both services — or, with partial config, hit a confusing mid-run `404` after money is already spent.
- Cost: testers give up; silent money waste on mid-run failures.

### After
- Same trigger.
- Now: the model just works with only an OpenRouter key connected. A typo'd id gets "not in OpenRouter's public model catalog — check the id at openrouter.ai/models"; a missing key gets "connect an OpenRouter API key first" — both pre-run, both $0.
- Win: all of OpenRouter's catalog usable instantly; refusals are diagnoses, not dead ends.

### Don't regress
- Curated compiled seeds + the engine↔gateway set-equality drift guard: **unchanged and green**.
- Typos still refuse pre-spend; a catalog outage is its own verdict (`openrouter_catalog_unavailable`), never mislabeled as a typo.
- `~`/`:` variant ids stay excluded (OME-819/OME-873 grammar); ollama's path untouched.
- Restart cleanly forgets admissions (in-memory only, zero persistence, zero teardown debt); a saved notebook silently re-admits at its next preflight.
- An Engine talking to an **older gateway** (no admit endpoint) behaves byte-identically to today.
- No existing test was modified anywhere; 51 new tests across the three stacks.

## How it works (one atomic commit per app)

**1. `feat(aigateway)` — OME-879.** New `POST /v1/models/admit`. Decision ladder, cheapest first: `AIGW_OPENROUTER_DYNAMIC` flag (new, default **on** — prod can pin a closed world without a breaking change) → provider enabled → id shape (`openrouter/<author>/<model>`, no `:variant`, no `~`) → caller's account actually credentialed (reuses the chat path's own resolution, so admission and dispatch can't disagree) → membership in OpenRouter's public catalog, fetched over the existing OME-479 bounded discovery transport and TTL-cached (a notebook's burst of admissions costs one dial). A grant registers the model on app state for the deployment's lifetime: it joins `GET /v1/models` and resolves on `GET /v1/model-parameters`. Refusals are 200 **answers** with a diagnostic code, never HTTP errors.

**2. `feat(url4-cloud)` — OME-880.** Implementation surfaced a plan-vs-reality gap: the Engine has **no run-submission preflight** — the familiar "not available on this Engine" refusal is raised client-side from the `/v1/models` projection, and the Engine's real per-model pre-spend seam is the `GET /v1/model-parameters` miss. So that miss is the admission trigger: for an OpenRouter-shaped id the Engine calls the gateway's admit endpoint; a grant joins an in-memory `AdmittedModels` overlay beside the frozen declared world, invalidates the catalog cache (so `/v1/models` lists it under a fresh ETag), and the fetch proceeds; a refusal passes the gateway's diagnostic body through verbatim on the existing caller-correctable 404 wire. Admitted ids also ride a new `URL4_CLOUD_EXTRA_MODELS` job-env key onto every scheduled run (both adapters, read at schedule time), merged **additively** into the runner's world — so the run process can actually route the model the App just admitted.

**3. `feat(screamingface)` — OME-881** (sub-issue added mid-build once the missing-preflight reality was clear). The SDK used to refuse a missing model from the `/v1/models` listing *before ever making the call that triggers admission*, which would have made the whole feature unreachable. An OpenRouter-shaped missing id now gets one free pre-spend probe of `GET /v1/model-parameters` instead — the Engine admits it (run proceeds) or answers with the gateway's refusal, decoded into a `PlanningError` that relays the gateway's own sentence. Every other missing id (typos, other providers, `~` ids) keeps today's immediate refusal, before any network call.

## Cross-service contract

- Engine → Gateway: `POST /v1/models/admit` `{"model_id": "openrouter/<author>/<model>"}` under the usual identity headers → `200 {"admitted": bool, "code": str|null, "message": str|null}`. Anything else (404 from an older gateway, timeout, unreadable body) reads as "cannot answer" and degrades to today's behavior.
- Engine → Client: admission refusals reuse the existing gateway-shaped `404 {"detail": {code, message, provider, model}}` pass-through — no new wire, one decoder.
- App → Runner: `URL4_CLOUD_EXTRA_MODELS` (JSON array of model ids), additive-only into the declared world; malformed values fail loud at Runner boot (App-written, so a bug, never caller input).

## Rollout

Gateway releases **before** the engine; each side degrades gracefully against an older counterpart. No migrations, no persisted state, no chart changes required (`AIGW_OPENROUTER_DYNAMIC` only if an operator wants it off).

## Test plan

- `run_gates.py` green for all three stacks (ruff, pyright, coverage gates, layering/enterprise checks, notebook determinism).
- aigateway: 16 endpoint tests — the full refusal ladder, idempotent re-admit, TTL cache (one dial per burst), catalog outage vs typo distinction, listing + model-parameters integration, auth.
- url4-cloud: 27 tests — shape gate, admit adapter (grant/refuse/unsupported/unreachable), overlay + ETag movement, `CachedCatalog.invalidate()`, `URL4_CLOUD_EXTRA_MODELS` world merge + both adapters + schedule-time freshness + ambient-value reset, loud-failure on malformed env.
- screamingface: 8 tests — probe fires and run proceeds (sync + async), each refusal code reaches the user pre-spend, older-Engine fallback keeps today's wording, non-admissible shapes never probe.
- **Live acceptance probe: ✅ PASSED (owner-run, 2026-08-19, local stack from this branch + real OpenRouter key).** `openrouter/qwen/qwen-2.5-7b-instruct` (real, unseeded) ran a 1-case IFEval evaluation via dynamic admission, then appeared in `GET /v1/models` and served its 50-parameter details contract. Refusal ladder verified live and pre-spend: a typo'd id → `model_not_on_openrouter` naming openrouter.ai/models (the ladder even caught the plan doc's own id slip — OpenRouter's real id carries a hyphen), a `~` variant and a non-OpenRouter id → today's `model_unavailable` wording, all $0.

Refs: OME-878, OME-879, OME-880, OME-881
Plan: `.dk/plans/2026-08-18-openrouter-dynamic-model-admission.md` (local)

### DIFF
--- a/apps/aigateway/src/aigateway/core/discovery_runtime.py
+++ b/apps/aigateway/src/aigateway/core/discovery_runtime.py
@@ -206,6 +206,12 @@ class DiscoveryRuntime:
         return self._limits
 
     @property
+    def client(self) -> DiscoveryHttpClient:
+        # OME-879: the admit route reuses the SAME bounded transport for its
+        # catalog-membership check, so admission gains no new egress surface.
+        return self._client
+
+    @property
     def cache(self) -> EvidenceCache:
         return self._cache
 
--- a/apps/aigateway/src/aigateway/core/plugin_base/__init__.py
+++ b/apps/aigateway/src/aigateway/core/plugin_base/__init__.py
@@ -21,6 +21,7 @@ from ..cache_ports import PROJECTION_BYPASS_REASON, CacheBypass, GlobalCacheProj
 from ._contract import ProviderPluginBase
 from ._ports import (
     CredentialStrategy,
+    ModelAdmission,
     ModelEntry,
     OAuthCodeExchangeRequest,
     OAuthConfig,
@@ -45,6 +46,7 @@ __all__ = [
     "CacheBypass",
     "CredentialStrategy",
     "GlobalCacheProjection",
+    "ModelAdmission",
     "ModelEntry",
     "OAuthCodeExchangeRequest",
     "OAuthConfig",
--- a/apps/aigateway/src/aigateway/core/plugin_base/_ports.py
+++ b/apps/aigateway/src/aigateway/core/plugin_base/_ports.py
@@ -35,6 +35,33 @@ class ModelEntry:
     litellm_params: dict[str, Any]
 
 
+@dataclass(frozen=True)
+class ModelAdmission:
+    """Outcome of a dynamic model-admission attempt (OME-879).
+
+    Think of it as the gateway's answer to "may this model join the served
+    catalog right now?": either a grant carrying the ``ModelEntry`` to serve, or
+    a refusal carrying a diagnostic ``code`` + ``message`` that name which knob
+    to turn — all decided BEFORE any money is spent.
+
+    INVARIANT: a grant always carries an entry; a refusal always carries a code.
+    Use the two constructors, never the raw dataclass fields.
+    """
+
+    admitted: bool
+    entry: ModelEntry | None = None
+    code: str | None = None
+    message: str | None = None
+
+    @classmethod
+    def granted(cls, entry: ModelEntry) -> ModelAdmission:
+        return cls(admitted=True, entry=entry)
+
+    @classmethod
+    def refused(cls, code: str, message: str) -> ModelAdmission:
+        return cls(admitted=False, code=code, message=message)
+
+
 class CredentialStrategy(ABC):
     """Per-provider credential producer (the auth "port").
 
--- a/apps/aigateway/src/aigateway/core/plugin_base/_provider.py
+++ b/apps/aigateway/src/aigateway/core/plugin_base/_provider.py
@@ -28,6 +28,7 @@ from ..api_key_validation import ApiKeyValidator
 from ..cache_ports import PROJECTION_BYPASS_REASON, CacheBypass
 from ._ports import (
     CredentialStrategy,
+    ModelAdmission,
     ModelEntry,
     OAuthCodeExchangeRequest,
     OAuthConfig,
@@ -37,6 +38,7 @@ from ._ports import (
 if TYPE_CHECKING:
     from ..credential_blob.store import CredentialBlobStore
     from ..oauth.identity import AccountIdentity
+    from ..parameter_discovery import DiscoveryHttpClient, DiscoveryLimits
     from ..profile_index import ProfileIndexStore
     from ..profile_models import AuthMode, AuthType
 
@@ -58,6 +60,35 @@ class ProviderPluginCore[TSettings: PluginSettings](ABC):
     @abstractmethod
     def register_models(self) -> list[ModelEntry]:
         """Return the model_list entries this plugin contributes."""
+
+    async def admit_model(
+        self,
+        model_id: str,
+        *,
+        discovery_client: DiscoveryHttpClient | None,
+        discovery_limits: DiscoveryLimits | None,
+        catalog_cache: dict[str, Any],
+        credentialed: bool,
+    ) -> ModelAdmission:
+        """Decide whether ``model_id`` may join the served catalog at runtime (OME-879).
+
+        The default answer is a refusal: dynamic admission is opt-in per
+        provider, and only a provider with an authoritative public catalog to
+        validate against (OpenRouter today) overrides this.
+
+        Args: ``discovery_client``/``discovery_limits`` are the bounded
+        discovery transport (``None`` when discovery is disabled);
+        ``catalog_cache`` is app-owned mutable scratch the provider may use to
+        TTL-cache its catalog; ``credentialed`` says whether the CALLING account
+        holds a usable credential for this provider — the route resolves it,
+        because credential scoping is account+profile shaped and lives in core.
+        Returns a ``ModelAdmission`` — never raises for a refusal.
+        """
+        del model_id, discovery_client, discovery_limits, catalog_cache, credentialed
+        return ModelAdmission.refused(
+            "dynamic_admission_unsupported",
+            f"provider {self.custom_llm_provider!r} does not support dynamic model admission",
+        )
 
     def conformance_models(self) -> list[ModelEntry]:
         """Return deterministic model representatives for contract conformance.
--- a/apps/aigateway/src/aigateway/main.py
+++ b/apps/aigateway/src/aigateway/main.py
@@ -51,6 +51,7 @@ from .routes import (
     auth_session,
     chat,
     health,
+    model_admission,
     model_parameters,
     models,
     oauth_connections,
@@ -376,6 +377,10 @@ def create_app(settings: Settings | None = None) -> FastAPI:
 
     app.state.pending_auth = PendingAuthTable(ttl_seconds=600)
     app.state.discovery_runtime = _build_discovery_runtime(settings)
+    # OME-879: dynamic-admission state. Deliberately app-state-only (deployment
+    # lifetime): a restart forgets every admission and nothing is persisted.
+    app.state.admitted_models = {}
+    app.state.admission_catalog_cache = {}
 
     for plugin in registry.all():
         auth_router = plugin.auth_router()
@@ -391,6 +396,7 @@ def create_app(settings: Settings | None = None) -> FastAPI:
     app.include_router(oauth_connections.router)
     app.include_router(health.router)
     app.include_router(models.router)
+    app.include_router(model_admission.router)
     app.include_router(providers.router)
     app.include_router(model_parameters.router)
     app.include_router(chat.router)
--- a/apps/aigateway/src/aigateway/plugins/openrouter_provider/admission.py
+++ b/apps/aigateway/src/aigateway/plugins/openrouter_provider/admission.py
@@ -1,0 +1,174 @@
+"""Dynamic OpenRouter model admission — the decision ladder (OME-879).
+
+FEATURE: run any OpenRouter model (OME-878). When a caller asks for an
+OpenRouter model outside the seeded list, this module answers "may it join the
+served catalog right now?" — think of it as a bouncer with a guest list it
+re-reads every few minutes.
+
+The ladder runs cheapest-first, in EXECUTION order:
+
+1. Flag — dynamic admission switched off (`AIGW_OPENROUTER_DYNAMIC=false`)
+   refuses everything, before any other work.
+2. Provider — a disabled provider (`AIGW_OPENROUTER_ENABLED=false`) can serve
+   nothing, so admitting into it would be a lie.
+3. Shape — the id must be `openrouter/<author>/<model>` exactly: no `:variant`
+   (a second route around gateway-owned search/caching, see `is_online_variant`)
+   and no `~` (the engine's colon escape, OME-873 — a `~` id is an encoding,
+   not a model).
+4. Credential — the calling account must hold a usable OpenRouter credential;
+   without one the admitted model could never dispatch, so refuse now with the
+   knob's name instead of mid-run later.
+5. Catalog — only now spend a network fetch: the model must appear in
+   OpenRouter's public catalog (bounded OME-479 transport, TTL-cached in
+   app-owned scratch so a burst of admissions costs one dial).
+
+Worked example: `openrouter/qwen/qwen2.5-7b-instruct` with the flag on, the
+provider enabled and a stored key walks 1-4 for free, costs one catalog fetch
+at 5, and is granted; re-admitting it 10s later hits the cached id set and
+costs nothing. `openrouter/qwen/qwen-typo` walks the same ladder and is refused
+at 5 with `model_not_on_openrouter` — $0 spent either way.
+
+INVARIANT: refusals are values, never exceptions — the route serves them as a
+200 answer, and an outage (`openrouter_catalog_unavailable`) is always distinct
+from a typo verdict (`model_not_on_openrouter`).
+"""
+
+from __future__ import annotations
+
+import time
+from typing import Any
+
+from aigateway.core.parameter_discovery import (
+    DiscoveryError,
+    DiscoveryHttpClient,
+    DiscoveryLimits,
+    fetch_discovery_json,
+)
+from aigateway.core.plugin_base import ModelAdmission, ModelEntry
+
+from .discovery import ALLOWED_ORIGINS, MODELS_URL, parse_catalog_model_ids
+from .settings import (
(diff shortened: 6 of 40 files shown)


khoaguin:
## Review — dynamic model admission

14 candidate findings were adversarially verified; 13 held, 1 was refuted (multi-replica admitted-set split — the gateway is pinned to a single uvicorn worker and `replicaCount: 1` in both Helm value files, and per-process scoping is documented as deliberate). The 10 that matter, most severe first. Findings 1–3 look like merge blockers: they break admitted models exactly in the restart/redeploy paths this PR is built for.

---

### 1. Engine's admitted overlay outlives a gateway restart and never heals
`apps/url4-cloud/src/url4_cloud/catalog/executable.py:195`

The engine's add-only `AdmittedModels` overlay survives a gateway restart, while the gateway's `app.state.admitted_models` is deliberately forgotten. After the restart, `real_model in self._admitted` is still `True`, so the engine skips the idempotent re-POST to `/v1/models/admit` that would heal the divergence and forwards `fetch_model_parameters` straight to a gateway that no longer knows the model. The gateway 404s with `model_not_found`, which passes through verbatim (404 is caller-correctable), and the SDK decodes a **permanent** `PlanningError` — "Model … is not available from the selected provider profile" — a misleading, profile-blaming verdict for a real, previously-admitted model. The overlay never evicts and nothing re-validates on a forwarded 404, so every retry fails until the **engine** is restarted. The admit route's own "saved notebook re-admits after a gateway restart" story never executes, because the re-admit request is never sent.

### 2. `default_model` validated against the pre-overlay tuple
`apps/url4-cloud/src/url4_cloud/world_config.py:278`

`_require_declared(section.default_model, models)` checks the stale local `models` captured **before** `_apply_extra_models` appends the overlay ids to `section.models`. A deployment that sets `URL4_CLOUD_AIGATEWAY_MODEL` to an id that arrives only via `URL4_CLOUD_EXTRA_MODELS` fails world parse on every run with `WorldConfigError "default_route … is not a declared model"` — even though the id *is* in `section.models`, so the error message lies about the world's contents.

### 3. `URL4_CLOUD_EXTRA_MODELS` is read by the App's own parse path
`apps/url4-cloud/src/url4_cloud/world_config.py:296`

`_apply_extra_models` reads the env var inside `parse_config` for **every** caller — including the App's composition root (`build_executable_catalog_service(settings, os.environ)` → `declared_model_ids`) — giving the Job-scoped per-run key a second, unintended meaning as ambient App env. A malformed ambient value raises `WorldConfigError` inside `declared_model_ids`, which `catalog/__init__.py:99-103` degrades to `None`: catalog routes 503 **and** the runner's extra-models provider goes dark, so real admissions silently stop reaching runs. A *well-formed* ambient value is worse: its ids silently join the App's declared projection, becoming listable and forwardable while bypassing the admission ladder entirely. The Job path defends against exactly this staleness (`inprocess.py` pops the key); the App-parse path has no scrub.

### 4. K8s adapter doesn't neutralize ambient `EXTRA_MODELS` from `envFrom`
`apps/url4-cloud/src/url4_cloud/adapters/k8s.py:292`

`K8sJobRunner._env` only appends the overlay when non-empty, and never neutralizes an ambient `URL4_CLOUD_EXTRA_MODELS` arriving via the Helm-owned `envFrom` ConfigMap — while the inprocess adapter unconditionally pops the same key (`inprocess.py:180`). If an operator leaves the key in the runner ConfigMap after an experiment, every K8s Job silently inherits the stale routes (or fails loud with a `WorldConfigError` attributed to an App bug), while the identical local-mode run scrubs the key and behaves fine. The two adapters diverge exactly where the PR's own contract tests assert they render one contract. Fix: always emit a neutralizing explicit env entry (explicit env wins over `envFrom`), mirroring the inprocess pop.

### 5. `invalidate()` destroys the stale-on-error fallback bodies
`apps/url4-cloud/src/url4_cloud/catalog/cache.py:148`

`CachedCatalog.invalidate()` clears every credential's cached entry — including the bodies that `stale_max_s` exists to protect. One admission fires `on_admitted` → `_entries.clear()`; if the gateway hiccups a second later, `_on_failure` finds no entry to serve stale and every `GET /v1/models` answers 502 fleet-wide, where pre-PR the stale body would have covered up to an hour. Bursts of admissions also re-clear the whole cache per grant, stampeding per-credential upstream re-fetches. A gentler invalidation — age `fetched_at` past `ttl_s` while keeping bodies for the stale path — preserves the class's own documented invariant ("an upstream blip must not empty every client's model list").

### 6. `_is_credentialed` collapses reauth/pending states into "no key"
`apps/aigateway/src/aigateway/routes/model_admission.py:67`

The bare `except HTTPException: return False` swallows the 401 `auth_required` (profile in ERROR, with its `reauth_url`) and 409 `profile_pending_auth` that chat and model-parameters surface deliberately. Admission then answers `provider_not_credentialed` / "connect an OpenRouter API key first" — relayed by the engine and surfaced verbatim by the SDK as a `PlanningError` naming the wrong fix. A user with an expired key loops on re-adding the same key; the real fix (reauth) is never named.

### 7. No cap, quota, or teardown on the admitted set
`apps/aigateway/src/aigateway/routes/model_admission.py:109`

Any authenticated caller holding one OpenRouter credential can admit the entire OpenRouter catalog (~500 real ids today; parser cap `_MAX_CATALOG_MODELS=10_000`) with a for-loop — the catalog dial is TTL-cached, so it costs about one upstream fetch. `app.state.admitted_models` has no removal path; every `GET /v1/models` serializes the extra rows for all tenants; engine-side each id joins the add-only overlay, so every scheduled K8s Job carries the full JSON list in `URL4_CLOUD_EXTRA_MODELS` and every run parses the extra `ModelSpec`s — for the rest of both deployments' lifetimes, with no teardown short of restarting both services.

### 8. Any string-detail 404 decodes as "model not installed"
`packages/screamingface/src/screamingface/_engine/catalog.py:314`

`_raise_model_not_installed` treats **any** 404 model-details response with a string detail and a `model` query param as "model not available on this Engine" — including FastAPI's default `{"detail": "Not Found"}` from a server or reverse proxy that lacks the `/v1/model-parameters` route. Previously that fell through to the generic `engine_contract_error` ("Could not load the SF Engine Model details: HTTP 404"), pointing at the deployment. Now `client.models.get()` and the parameter preflight raise a permanent `PlanningError` "Model 'x' is not available on this Engine" for every model — including ones present in the listing — steering users to "fix" correct model ids. The decode fires on all details calls, not just the admission probe the ledger documents, and the shape check can't distinguish the Engine's genuine not-installed body from a route-missing body.

### 9. Unscoped shared admission cache invites cross-provider collision
`apps/aigateway/src/aigateway/routes/model_admission.py:104`

One flat `app.state.admission_catalog_cache` dict is handed unscoped to every plugin's `admit_model` port, and the OpenRouter module writes generic keys (`ids`/`expires_at`) into it. The second provider to implement the port — following OpenRouter as the reference pattern — will read/write the same keys: OpenRouter's `_catalog_ids` would treat provider B's frozenset as a valid fresh OpenRouter catalog within the 300s TTL, refusing real OpenRouter models as `model_not_on_openrouter` and admitting colliding foreign ids into OpenRouter's namespace. One-line fix at the shared seam: hand each plugin `cache.setdefault(provider, {})`.

### 10. Third divergent copy of the admissibility grammar in the SDK
`packages/screamingface/src/screamingface/_evaluation/runner.py:344`

`_defers_to_details_probe` omits the engine's `is_route_legal` check despite its docstring claiming to "mirror the Engine's own admissibility gate", and its hardcoded `openrouter/` prefix means any future dynamically-admissible provider requires a lockstep three-component release. Today, charset-illegal ids like `openrouter/a/b!` are deferred though the engine refuses them locally (wasted probe; correct error only via the fallback decoder). Tomorrow, a second `admit_model` provider ships and every deployed SDK lands its ids in the `hopeless` branch — a permanent `model_unavailable` `PlanningError` before the probe ever fires. Since the probe is free and pre-spend, and finding 8's decoder makes probing a non-admitting engine byte-identical to today's refusal, deferring *any* missing id (no grammar copy at all) was safe.

---

khoaguin:
All 10 findings addressed — one fix commit per app, plus a rebase onto latest `main` (which renamed `apps/url4-cloud` → `apps/screamingface-engine`; the engine commit now lives under the new package name).

| # | Finding | Fix |
|---|---------|-----|
| 1 | Engine overlay outlives a gateway restart | The overlay **heals on a forwarded 404**: an overlay id whose parameters fetch 404s is evicted and re-admitted once — grant → retry the fetch (the saved-notebook story now actually executes), refusal → relayed, bounded to one retry. `AdmittedModels` gained `discard()`; the invariant is now "the overlay never contradicts the gateway's latest answer". |
| 2 | `default_model` validated pre-overlay | `_require_declared` now checks `section.models` **after** `_apply_extra_models` merges — an overlay-delivered default boots. |
| 3 | `EXTRA_MODELS` read by the App's own parse | The key is now an **explicit opt-in** (`include_extra_models=True`) that only the Runner-boot `load_config` passes. The App's parse physically cannot absorb an ambient value — neither the 503 nor the admission-bypass path exists anymore. |
| 4 | K8s adapter doesn't neutralize `envFrom` | The Job env **always** carries an explicit `URL4_CLOUD_EXTRA_MODELS` entry (empty when no overlay) — explicit env beats `envFrom`, mirroring the inprocess pop. The prior exact-set env test gained the key with a dated justification (same pattern as `STREAM_GRACE_S`). |
| 5 | `invalidate()` destroys stale fallback bodies | Invalidation is now **expiry, not deletion**: `fetched_at` is aged past the TTL (only ever moved backward), so the next fetch re-reads upstream but `stale_max_s` fallback bodies survive a post-admission gateway blip. |
| 6 | Reauth/pending collapsed into "no key" | `_credential_verdict` relays `auth_required` ("must be reconnected — reauthorize") and `profile_pending_auth` ("still connecting — finish the connection") as their own admission refusal codes; only a genuinely missing profile words as "connect a key first". SDK decodes both (pending is non-permanent). |
| 7 | No cap on the admitted set | `_MAX_ADMITTED_MODELS = 256`, checked before any credential/catalog work; at capacity the refusal is `admission_capacity_reached` naming the static-seed and restart knobs. Already-admitted ids re-admit idempotently even at capacity. Real eviction (LRU/TTL/admin delete) is deliberately a follow-up design ticket, not invented here. |
| 8 | Any string-detail 404 decoded as "not installed" | The rewrite moved **out of the shared details decoder into the probe call site** — only the probe, which knows it asked about a listing-missing model, reads a bare 404 as "not available on this Engine". A proxy/route-missing 404 on `models.get()` stays `engine_contract_error` (deployment diagnosis), with a test pinning exactly that. |
| 9 | Flat cross-provider admission cache | The route hands each plugin `cache.setdefault(provider, {})` — per-provider compartments, collision structurally impossible. |
| 10 | Third divergent grammar copy in the SDK | `_defers_to_details_probe` is **deleted**: every listing-missing id defers to the free pre-spend probe; the Engine and Gateway are the only admissibility authorities, and a future second provider needs zero SDK release. The two prior availability tests keep every assertion byte-identical — only their fake engines were made faithful (a real Engine 404s details for unlisted models). |

New tests: engine +10 (heal ×5, stale-preservation, overlay default, App-parse isolation ×2, K8s neutralization), aigateway +5 (reauth/pending relay, cap ×2, cache namespacing), SDK +1 (proxy-404 diagnosis) — plus the amended pins. All three stacks: `run_gates.py` green (append-only skipped once, on the in-session owner approval for the three documented test amendments).

keelancj:
@khoaguin it's not open yet so can't actually approve, but looks good 😊

keelancj (approved):

khoaguin:
**Follow-up from owner review of the fix round** (`5801d407`): the 256-model admission cap was bypassable under concurrency — the route's capacity pre-check and its insert are separated by two awaits (credential resolution, catalog dial), so N simultaneous admissions could all pass a pre-check taken at `len == cap-1` and then all insert (TOCTOU; asyncio's single thread doesn't help because every `await` is a scheduling point).

Fixed by making `_store_admission` the definitive enforcement: membership + cap check + insert with **no awaits between them**, atomic under the single-threaded loop; the top-of-route check stays as a cheap pre-dial refusal. Regression test reproduces the race deterministically (the awaited plugin decision itself fills the last slot) and was verified RED against the previous code.

With enforcement now real, the cap is raised **256 → 1024** — above OpenRouter's entire public catalog (~415 models today), so it never constrains legitimate use and acts purely as a runaway backstop.

HupBaHa (approved):
Approved after current-head recheck at 5801d407. I specifically rechecked the Tortoise/ORM boundary: no model or migration changes, admission state stays app-state-only, Postgres migration/lifecycle race checks pass, and the new cap enforcement is an in-memory asyncio race fix rather than DB-dependent behavior.

Checks run: gateway focused admission/profile race tests passed, explicit Postgres migration/lifecycle checks passed, full aigateway non-live suite passed, ruff passed, pyright passed, no-enterprise guard passed.

Non-blocking follow-ups: tighten wording from any OpenRouter model to route-legal non-variant OpenRouter models, fix stale docs mentions of URL4_EXTRA_MODELS vs URL4_CLOUD_EXTRA_MODELS, and keep OME-886 as the long-term queue-wait vs provider-execution timeout fix.

HupBaHa:
Heads up: this PR currently has merge conflicts with main. GitHub reports mergeable=CONFLICTING / mergeStateStatus=DIRTY.

I checked the virtual merge and the conflict is in:

- apps/screamingface-engine/src/screamingface_engine/job_env.py

The checks are green, but merge is blocked until this is rebased/merged with main and the conflict is resolved. After that, I can re-check the final diff if needed.

Issue #642, opened 2026-08-18
Engine truncates candidate result at 1 MiB but reports the run succeeded, so the SDK fails to decode it


siddhant230:
## Description

The Engine truncates a candidate result body to 1 MiB and still terminates the run
`succeeded`. The truncation cuts mid-JSON, so the SDK's decode fails and the whole
evaluation is lost with `ExecutionError: SF Engine Candidate result must be JSON`
(`Code: execution_failed`).

The failure is silent and late: the live progress panel renders **"Evaluation complete ·
succeeded"** (it folds the `Terminated` event, which says `succeeded`), and only afterwards
does `report_from_outcomes` throw. An hour of paid model work is unrecoverable — the
truncated body is the only copy of the result the client ever receives.

The cap is a hardcoded constructor default with no environment override, so there is no
operator knob to work around it.

**Relevant code**

- `apps/url4-cloud/src/url4_cloud/runner/executor.py:564` — `result_cap: int = 1_048_576`
- `apps/url4-cloud/src/url4_cloud/runner/executor.py:449-466` — `build_result()` cuts on a
  byte boundary and appends `…[truncated]`, with no signal to the caller that it did
- `apps/url4-cloud/src/url4_cloud/runner/executor.py:619` — where the cap is applied
- `packages/screamingface/src/screamingface/_evaluation/results.py:129` — where `json.loads`
  raises
- `packages/screamingface/src/screamingface/_engine/transport.py:43-51` — the SDK already
  documents "The Engine truncates a result body to 1 MiB" and raises the WS frame limit to
  8 MiB so a capped result is *deliverable*. Delivering a truncated result only moves the
  failure from the transport to the decoder; the decode path was never addressed.

## How to Reproduce

1. Start the local stack: `just stack-up` (gateway :9105, scoreboard :9106, engine :9108),
   with `URL4_BENCHMARK_ASSETS` pointing at prepared HealthBench assets.
2. Point the notebook at the local engine:
   `os.environ["SCREAMINGFACE_ENGINE_URL"] = "http://127.0.0.1:9108"`, then `sf.connect()`.
3. Build a 5-member Fusion with a synthesizer (any panel large enough to produce verbose
   rubric evidence), e.g. `open_pentagon_diverse` = qwen3.8-max + glm-5.2 + deepseek-v4-pro
   + nemotron-3-ultra-550b-a55b + minimax-m3, synthesized by kimi-k3, `max_tokens=16384`,
   `temperature=0.0`.
[...]

PR #647, opened 2026-08-19
feat: deliver large Evaluation results in full instead of cutting them off at 1 MiB

khoaguin:
Closes OME-892 (sub-issue of OME-888 · GitHub #642).

### Before
- Trigger: a researcher runs a large paid Evaluation whose aggregate result exceeds 1 MiB.
- Today: the Engine silently cuts the result mid-JSON, reports the run **succeeded**, and the researcher discovers via a cryptic `SF Engine Candidate result must be JSON` error that an hour and several dollars produced nothing recoverable (documented instance: 1h 11m, $6.54, 137 model calls). Workaround is running in `limit=10` batches and averaging scores by hand.
- Cost: destroyed paid work, a green panel that lies, manual batch bookkeeping.

### After
- Same trigger.
- Now: the run succeeds and the full result arrives. The Engine parks an oversized result on disk and the result frame carries a **claim ticket** `{id, size_bytes, sha256}`; the SDK redeems it over HTTP (streaming, retried on transient resets) and verifies byte count + sha256 **before** building the Report. Truly enormous results fail loudly with `result_too_large` naming actual and allowed bytes.
- Win: big Evaluations just work; "succeeded" always means "your complete, verified Report is here".

### Don't regress
- Results at or under 1 MiB travel inline, **byte-identical** to today (`artifact` is omitted from the frame unless a ticket is issued) — old SDKs don't notice anything on small runs.
- The progress stream stays responsive: bulk bytes never ride a WS frame, and the spill runs off the event loop so heartbeats can't stall.
- Failed terminations still carry cost/usage telemetry.
- No disk leaks: artifacts are TTL-swept (startup + hourly) — fetching never deletes, so retries, HTTP Range resume, and deduped tickets all keep working.

## Root cause

The result's only delivery vehicle was one WebSocket frame, double-bounded (engine truncation cap 1 MiB; client frame limit 8 MiB). `build_result()` cut the UTF-8 body at the cap, appended `…[truncated]`, and still emitted `Completed` — so the lifecycle terminated `succeeded` while destroying the only copy of the result in flight.

## Design: WS for signaling, HTTP for bulk

Industry-standard split (every WS stack caps messages; Slack/Discord/Jupyter all move bulk out-of-band):

1. **`packages/url4`** — `ResultData` becomes body XOR artifact (validator-enforced); new `ResultArtifact` with sha256-hex patterns that double as the path-traversal guard.
2. **Engine** — `build_result()` forks, hard cap first: > `URL4_CLOUD_RESULT_HARD_CAP_BYTES` (default 1 GiB) → run fails `result_too_large`; > `URL4_CLOUD_RESULT_INLINE_CAP_BYTES` (default 1 MiB) → complete body spilled to a content-addressed store (`artifacts/<sha256>`, atomic writes, idempotent); otherwise inline as today. **The truncation code path is deleted** — a mangled body is unrepresentable.
3. **Engine REST** — `GET /artifacts/{id}` (capability-guarded, `FileResponse`: streaming + Range). Cleanup is **TTL-only**: swept at startup and hourly for the life of the App (crash-proof loop). The sync `GET /` path resolves artifact results to the same bytes.
4. **SDK** — strict ticket decode; redemption happens **after** the WS closes, streams against the ticket's declared size (bounded memory), retries transient resets (3×, backoff), verifies size + sha256 before `json.loads`; mismatch → `result_integrity_mismatch` with both numbers. A legacy `…[truncated]` body from an older Engine now raises `result_truncated` naming the received byte count.

**Compat under version skew** (additive contract): new+new = the feature · new engine + old SDK = small results byte-identical, big results fail loudly (strictly better than today's silent loss) · old engine + new SDK = inline path unchanged plus the marker detection.

## Review hardening

An adversarial review (10 verified findings) ran before this PR; all are fixed in these commits — highlights: delete-on-fetch removed entirely (it broke retry, Range resume, and deduped tickets), fetch retry + streaming size bound, hard-cap-first ordering, sweep race tolerance + `.tmp` collection, byte-identical inline frames via `Field(exclude_if=…)` (a wrap-serializer erased the model's OpenAPI schema), single-encode + off-loop spill. Details in `docs/work/2026-08-19-OME-892-deliver-large-results.md`.

## Test plan

- `run_gates.py` ALL GREEN × 3 stacks: url4 (1,170 tests, cov ≥95) · screamingface-engine (1,761 incl. 2 new end-to-end integration tests driving a real over-cap run through the real executor/lifecycle/WS/REST, cov ≥80) · screamingface (929, cov ≥95 + notebook determinism + build + distribution checks).
- No paid provider calls anywhere; deterministic and network-free (tiny caps make small synthetic results exercise every tier).
- One prior test replaced (`test_over_cap_result_is_truncated_with_marker` pinned the deleted truncation behavior — superseded by 4 fork-contract tests); append-only gate run with the documented `--skip-append-only` acknowledgment.

## Open items (not in this PR)

- **Hosted k8s**: the Runner Job pod and the App pod don't share a disk — `URL4_CLOUD_ARTIFACTS_DIR` must name a shared volume in the chart (flagged via AIDEV-NOTE in `job_env.py`; the artifact-404 message names this cause). Local mode — the bug's repro environment — is complete end-to-end.
- **Multi-candidate partial reports**: OME-888's "Partial Report survives" criterion needs its own ticket — the SDK has no partial-report path today (pre-existing; any candidate failure aborts the evaluation).

## Live-run validation (2026-08-19, real paid run)

The owner ran the full `healthbench-worst30` evaluation (157 cases, ~960 model calls, ~50 min, ~$30) against this branch's engine. The result landed at **2,835,927 bytes — 2.7× over the inline cap**, i.e. exactly the #642 scenario:

- ✅ **Spill validated**: the engine wrote the content-addressed artifact (sha-named file verified byte-identical) and the terminal event carried the claim ticket. Under `main`, this run's result would have been truncated mid-JSON and destroyed.
- 🐛 **Defect found & fixed** (`4a45b6bb`): redemption reused the capability token minted at run **start**; tokens live 60 s, the run took 50 min → `GET /artifacts/{id}` 401'd. No test could see it — tests fetch milliseconds after mint. Redemption now **mints a fresh token inside the fetch retry loop**; a new `protocol_server` seam (`artifact_token_expiry`) expires all run-start tokens once the result streams, reproducing the incident deterministically (sync + async regression tests).
- 💾 **Nothing lost**: TTL-only cleanup (review finding 1+2) kept the parcel on disk; the full result was recovered by hand. The failure mode was a *named* `AuthenticationError` — loud, not the old silent truncate-and-lie.

## Reviewer routing

- **Contract shape** (`ResultData.artifact`, OME-618 alignment): @keelancj — this is the OME-888 decision point.
- **Engine store / sweeper / REST route + url4 model**: @sergio-bershadsky.

### DIFF
--- a/apps/screamingface-engine/src/screamingface_engine/app.py
+++ b/apps/screamingface-engine/src/screamingface_engine/app.py
@@ -7,6 +7,8 @@ tests substitute for the production ones built here from `Settings`.
 
 from __future__ import annotations
 
+import asyncio
+import contextlib
 import logging
 import os
 from pathlib import Path
@@ -15,6 +17,7 @@ from fastapi import APIRouter, FastAPI
 from fastapi.staticfiles import StaticFiles
 
 from screamingface_engine.adapters.factory import build_job_runner
+from screamingface_engine.artifacts import ArtifactStore
 from screamingface_engine.auth import Clock, install_problem_handlers
 from screamingface_engine.benchmarks import EMPTY_BENCHMARKS, BenchmarkRegistry
 from screamingface_engine.benchmarks.builtins import BUILTIN_BENCHMARKS
@@ -28,6 +31,7 @@ from screamingface_engine.metrics import MetricsMiddleware, build_metrics, regis
 from screamingface_engine.ops import router as ops_router
 from screamingface_engine.rest import (
     SubscriberGate,
+    artifact_router,
     benchmark_router,
     catalog_router,
     connection_router,
@@ -47,6 +51,7 @@ _DIAGRAMS_DIR = Path(__file__).parent / "assets" / "diagrams"
 _ROUTERS = (
     router,
     rest_router,
+    artifact_router,
     benchmark_router,
     catalog_router,
     connection_router,
@@ -87,6 +92,15 @@ def create_app(
     app.state.connections = connections
     app.state.benchmarks = benchmarks
     app.state.metrics = build_metrics()
+    # FEATURE: deliver large results in full (OME-892) — the serve side of the spill store.
+    artifact_store = ArtifactStore(Path(settings.artifacts_dir))
+    app.state.artifact_store = artifact_store
+    # WHY startup + periodic: artifacts are deleted on fetch, so the only leak source is
+    # parcels nobody redeemed (crashed runs, vanished clients). The boot sweep clears the
+    # backlog immediately; the periodic task covers a hosted App that stays up for weeks
+    # between redeploys — startup-only would let abandoned parcels pool until the next
+    # restart (owner decision on OME-892).
+    _install_artifact_sweeper(app, artifact_store, settings)
     # WHY: pass a getter, not `catalog` directly — the collector re-reads app.state.catalog on
     # every /metrics scrape rather than capturing the value built here.
     register_catalog_metrics(app.state.metrics, lambda: app.state.catalog)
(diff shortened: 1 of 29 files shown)


keelancj (approved):
The artifact approach makes sense to me and that fresh token fix addresses the long-run issue. So looks good from my side! Is there already a follow-up ticket for the shared storage work in k8s?

IonesioJunior (approved):

PR #656, opened 2026-08-20
fix(scoreboard): trim landing — drop stats strip + Dataset column

IrinaMBejan:
## Summary
Slims the scoreboard portal landing page (`apps/scoreboard/portal/`) for the leaderboard MVP. Frontend-only — no Python/API/model changes.

- **Remove the stats strip** (Benchmarks / Datasets published / Newest benchmark), plus the now-dead `updateIndexStats` (+ its call) and `formatDateOnly` in `main.js`.
- **Remove the Dataset column** from the benchmarks table — the `<th>` and the `<td>` block in `benchmarkRow`, plus the orphaned `publishedDataFileName`. `httpUrlOrNull` stays (exported, used by `benchmark.js`/`spec.js`).
- **Focus column is unchanged** — it keeps rendering the editorial `b.focus` field; `benchmarkSubtitle` (the name-subtitle) is unchanged too.

## Test plan
- `run_gates.py scoreboard` — ALL GREEN (append-only check, ruff check, ruff format --check, pyright, pytest --cov=scoreboard --cov-fail-under=80, `node --test tests/portal/leaderboard-logic.test.js`).
- `node --check portal/main.js` OK; orphan grep clean (no `updateIndexStats`/`formatDateOnly`/`publishedDataFileName`/`stat-*`).
- `benchmarkRow` appends exactly 5 `<td>`s matching the 5-column header.

## Owner-verify
Eyeball the served landing page — confirm no stats strip, no Dataset column, and Focus shows the editorial focus.

Refs: OME-902

### DIFF
--- a/apps/scoreboard/portal/index.html
+++ b/apps/scoreboard/portal/index.html
@@ -55,12 +55,6 @@
 
     <p><a class="btn" href="https://docs.screamingface.ai/" target="_blank" rel="noopener">Get started with ScreamingFace →</a></p>
 
-    <div class="stats" aria-label="Live portal counts">
-      <div class="s"><span class="k">Benchmarks</span><div class="v" id="stat-benchmarks">—</div></div>
-      <div class="s"><span class="k">Datasets published</span><div class="v" id="stat-datasets">—</div></div>
-      <div class="s"><span class="k">Newest benchmark</span><div class="v" id="stat-newest">—</div></div>
-    </div>
-
     <h2 id="benchmarks">Benchmarks</h2>
 
     <p class="faint">Pick a benchmark to open its leaderboard. Inside, you can tab across all benchmarks.</p>
@@ -75,7 +69,6 @@
             <th>Focus</th>
             <th class="num">Fusions</th>
             <th class="num">Best reproducible</th>
-            <th>Dataset</th>
             <th>Open</th>
           </tr>
         </thead>
--- a/apps/scoreboard/portal/main.js
+++ b/apps/scoreboard/portal/main.js
@@ -255,21 +255,6 @@ window.ScorePortal = (function () {
   }
 
   /* ---- index page ------------------------------------------------------ */
-  // Site-root data files (published by the deploy workflow) open in the
-  // in-portal viewer instead of forcing a download — GitHub Pages serves
-  // .jsonl as application/octet-stream with no MIME override available.
-  function publishedDataFileName(href) {
-    try {
-      var u = new URL(href);
-      if (u.hostname !== "scoreboard.screamingface.ai") return null;
-      var m = u.pathname.match(/^\/([A-Za-z0-9][A-Za-z0-9._-]*\.jsonl)$/);
-      if (!m) return null;
-      return ["livetruth-latest.jsonl", "livetruth-masking.dataset.jsonl"].indexOf(m[1]) === -1 ? null : m[1];
-    } catch (e) {
-      return null;
-    }
-  }
-
   // "Subtitle" isn't an explicit field on Benchmark — using `description`
   // provisionally (flagged to Irina on OME-768; easy to swap if she says
   // otherwise, this is the one place it's read).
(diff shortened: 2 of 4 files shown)


FilipBolt:
No objection to the trim itself — dropping the stats strip and the Dataset column reverses calls I made in `OME-874`, and they were judgement calls flagged as reversible in one edit. The dead-code removal is careful: `updateIndexStats`, `formatDateOnly` and `publishedDataFileName` all go, and `httpUrlOrNull` correctly stays because `benchmark.js`/`spec.js` still use it.

**One thing to settle before this merges**, because two open PRs are editing the same cell toward opposite designs.

## The collision

| PR | What Focus becomes |
|---|---|
| **#656** (this, `OME-902`) | `b.description` truncated to 50 chars — stops reading `b.focus` entirely |
| **#657** (`OME-904`) | adds `focus` to the **Engine** benchmark definitions and seeds it through the new Engine catalogue |

Both are open right now. #657 carries `focus` end to end — Engine definition → `catalog_entry()` → seed job → `Benchmark.focus` → `GET /v1/benchmarks` — and this PR removes the only thing that renders it. Whichever merges second either takes a textual conflict in `benchmarkRow` or silently wins.

**If this one lands last, `focus` is authored, transported, stored and served, and displayed nowhere.** That is a worse outcome than either design on its own, and neither PR can see it from inside its own diff.

## Why the duplication is the real question

This PR is candid that the description then appears twice on one row — full under the name, truncated in Focus — and asks in owner-verify whether that reads acceptably. Rendered, a row is:

```
DRACO                                            | A 100-task DRACO reproduction with…
A 100-task DRACO reproduction with official…     |
draco                                            |
```

That duplication is exactly what a separate `focus` field existed to avoid: a short line saying something the long description does not.

Worth knowing this slot has a history. **`OME-768` asked what the catalogue subtitle should be on 2026-08-11** — *"nothing on `Benchmark` is explicitly a subtitle field, provisionally using `description`"* — flagged to @IrinaMBejan and never answered. It has since been three different things in four days: `description` as a name-subtitle (`OME-768`), a new `focus` field (`OME-874`), and now `description` again, truncated (`OME-902`).

Also relevant: the brand mockup populates that column from `b.subtitle`, and its name cell is **just the linked name** — no subtitle, no id. So the mockup's answer is one short line in Focus and nothing repeated under the name. Our duplication comes from keeping the name-subtitle *and* filling Focus from the same source.

## Three ways to settle it

1. **Engine `focus` wins** (#657 as written) — Focus shows the Engine's short line, name-subtitle keeps the full description. No duplication, one authoring site, and it is the mockup's shape.
2. **Description preview wins** (this PR) — then `focus` should be dropped from #657 rather than shipped unused, and the name-subtitle probably goes so the text is not doubled.
3. **Both, distinctly** — Focus shows `focus` when the Engine declares one and falls back to a description preview when it does not. Most forgiving, most code.

I lean 1: it removes the duplication, matches the mockup, and #657 already builds the pipe. But this is a copy/product call rather than an engineering one, so it is yours — and I have an interest to declare, since I added the `focus` field in `OME-874`.

## Two smaller notes

**The Dataset column was the only route to `data.html`.** That page and its `PUBLIC_ARTIFACTS` allowlist still exist and still serve the published JSONL — after this, nothing on the site links to it. Fine if deliberate; worth a follow-up if not.

**50 characters truncates mid-word**, so DRACO reads *"A 100-task DRACO reproduction with official score a…"*. Breaking on the last space before the limit reads better and costs one line — non-blocking either way.

IrinaMBejan:
Updated per the comment, the focus is now untouched. Thanks!!

FilipBolt (approved):
Approving. Thanks for turning that around so fast — leaving `b.focus` alone resolves the collision with #657 cleanly: its Engine-sourced `focus` now has something rendering it, and the description no longer appears twice on one row.

Scope is exactly the two removals now, and the dead-code cleanup is careful: `updateIndexStats` and its call, `formatDateOnly`, `publishedDataFileName` all go, while `httpUrlOrNull` correctly stays because `benchmark.js` and `spec.js` still use it. Header and `benchmarkRow` agree at five columns. 11/11 checks green.

One loose end from my earlier comment, non-blocking and not really yours to fix here: **the Dataset column was the only link to `data.html`**. That page still exists, and `PUBLIC_ARTIFACTS` in `src/scoreboard/portal.py` still serves the three published JSONL files — `_validate_public_artifacts` even fails app startup if they go missing. So after this, real published artifacts stay served and reachable by direct URL, but nothing on the site points at them.

Fine if that is deliberate. If the artifacts are meant to stay discoverable, it wants a follow-up ticket rather than a change here.

PR #657, opened 2026-08-20
feat(scoreboard): seed benchmark text from the Engine catalogue

khoaguin:
Closes OME-904 — https://linear.app/openmined/issue/OME-904/show-benchmark-descriptions-on-the-leaderboard-engine-text-as-the-only

## TLDR
**The problem**: open the leaderboard and every benchmark has no descriptions. 
Screenshot 2026-08-20 at 17 44 27

**Why it was empty**: those descriptions were written — twice. Once in the engine, right next to each benchmark's code. And once again, typed by hand into the scoreboard's deploy settings file.

⚠️ However, neither one reached the scoreboard's Postgre database.

The hand-typed copy lost because our deploy repo supplies its own list of benchmarks, and Helm's rule is that an override replaces a whole list instead of merging into it. So our carefully written list got thrown away and replaced by theirs, which has ids but no descriptions. And the engine's copy lost because nothing ever read it — there was no connection between the engine and the board at all.

💡**The fix**: delete the hand-typed copy. Now, during each deploy, the scoreboard asks the engine directly — "which benchmarks do you have, and what does each one say?" — and saves the answer. One copy, living next to the code it describes, and no way for a deploy setting to blank it out again.

> **Before / after diagram** — how the text travels today vs. after this change: see the [OME-904 issue](https://linear.app/openmined/issue/OME-904/show-benchmark-descriptions-on-the-leaderboard-engine-text-as-the-only). Deliberately not committed; the repo keeps no binaries for it.

## Why it was empty

A benchmark's prose existed **twice**: once in the Engine benchmark definition it describes, once hand-typed into the Scoreboard chart values. In production neither reached the database. The deploy repo supplies its own `seedBenchmarks.benchmarks`, Helm **replaces** lists rather than merging them, and that list carried ids and revisions only — so the chart's copy never entered the pipe at all, and the seed job faithfully wrote `description = NULL`.

## What this changes

The Engine benchmark definition becomes the **only** place a benchmark's text is written.

- `Benchmark` gains optional `focus` and `dataset_url`, so all four fields the board displays have one authoring site. They are validated at construction and published in `GET /v1/benchmarks` only when declared — an absent key, never a null.
- The seed job reads that catalogue at deploy from `seedBenchmarks.engineUrl`. It is public and read-only, so the job holds no Engine credential.
- Chart values keep only which Engine to ask, plus the three legacy demo entries that predate the Engine catalogue and have no Engine definition.
- **Engine rows win** over a configured entry sharing their id, and the shadowed id is named in the job's output. That precedence is what makes the Engine the only copy rather than merely the preferred one — prose reintroduced by a deploy is ignored, not applied.

`revision` rides the same response, so the value the board ranks by and the value the Engine computes can no longer drift apart.

When the Engine is unreachable the job seeds the configured rows, leaves existing rows untouched, and exits **zero** — re-seeding refreshes a populated board, and failing a Scoreboard release on an unrelated service's health costs availability and buys nothing. It exits **non-zero** only when no row carries a revision at all, which is true only before the first successful seed; exiting zero there would publish a board holding nothing but demo entries and call it a success.

## Reviewers: two things worth your attention

**1. The chart's pinned revisions are already stale.** Surfaced by a test written for this ticket, not caused by it:

| benchmark | pinned in `values.yaml` | computed by `main` |
|---|---|---|
| `draco` | `1c58b3085912e304` | `66a463248586b277` |
| `ifeval` | `0b88a52b5f10a6d9` | `1cba769ece27f7ef` |
| `healthbench-worst30` | `6cd57aee171fbdc4` | `39cfd96b068f7230` |

This makes the ticket's suggested quick unblock — re-seed by hand from `values.yaml` — actively harmful: it fills the descriptions while pinning revisions no current Engine will submit against. Deploying this change is the better path. The computed hashes are deliberately **not** pinned in a test either; that would rebuild the hand-copied second copy this PR deletes.

**2. `engineUrl` defaults to empty**, because inventing a hostname in the repo default is a guess. Until the deploy repo sets it, the board seeds only the legacy demos. Happy to add a default if there is a canonical origin.

## Test plan

- `uv run .claude/scripts/run_gates.py screamingface-engine` — ALL GATES GREEN
- `uv run .claude/scripts/run_gates.py scoreboard` — ALL GATES GREEN
- `helm template` verified both ways: `SCOREBOARD_SEED_ENGINE_URL` renders when `engineUrl` is set, and is absent when it is empty.

New tests: `apps/screamingface-engine/tests/unit/test_benchmark_display_metadata.py` (catalogue and detail carry the new fields; absent means absent; blank focus and non-`http(s)` links refused; the installed benchmarks publish what the board shows; display metadata never moves a revision) and `apps/scoreboard/tests/unit/test_seed_engine_catalog.py` (field mapping, unknown catalogue fields ignored, URL joining, transport/status/non-JSON/malformed-entry each surfacing as one seed-owned error, Engine precedence, legacy rows kept, seeded revision comes from the catalogue, unreachable-Engine survivable vs. bootstrap failure).

No migration: `benchmarks` already carries `description`, `focus`, `dataset_url`, and `revision`.

### Before

- Trigger: someone opens the leaderboard in the notebook and scans the benchmark list.
- Today: every entry reads "No description published." They cannot tell what DRACO, IFEval, or the HealthBench challenge test without leaving the notebook.
- Cost: the catalogue looks unfinished, and prose someone already wrote is silently lost between two hand-maintained copies that nothing compares.

### After

- Same trigger.
- Now: each benchmark shows its description, focus line, and dataset link, straight from the definition that describes it.
- Win: the board explains itself, and a future deploy cannot blank the text, because there is no second copy left to lose.

### For the next dev

- Adding a benchmark means writing its text in one file, next to the benchmark. There is no chart entry to remember and no revision to re-copy.
- A deploy that reintroduces prose under a published id no longer silently wins; the seed job ignores it and says so in its log.

### Don't regress

- Seeded `revision` matches the Engine's computed revision, now structurally rather than by discipline.
- Seeding stays idempotent; re-running never duplicates or corrupts rows.
- The portal's Focus column and dataset links keep working, including the em-dash fallback for a benchmark that declares neither.
- Legacy demo entries keep their hand-written values — they are retained, not adopted.

### DIFF
--- a/apps/scoreboard/charts/scoreboard/README.md
+++ b/apps/scoreboard/charts/scoreboard/README.md
@@ -20,7 +20,11 @@ The app chart is database-agnostic. It consumes `SCOREBOARD_DATABASE_URL` from `
 
 ## Benchmarks
 
-The post-install/post-upgrade seed Job runs `python -m scoreboard.seed` and passes `.Values.seedBenchmarks.benchmarks` as JSON. Re-running the Job is safe because benchmark registration is idempotent.
+The post-install/post-upgrade seed Job runs `python -m scoreboard.seed`. Re-running the Job is safe because benchmark registration is idempotent.
+
+Each benchmark's text — display name, description, focus line, dataset link — is read at deploy from the Engine named by `.Values.seedBenchmarks.engineUrl`, over its public `GET /v1/benchmarks` catalog. That Engine definition is the only place the text is written, so `revision` cannot drift from the Engine's computed value and a values override cannot blank the prose (OME-904). Set `engineUrl` per deployment; leaving it empty seeds only the configured entries.
+
+`.Values.seedBenchmarks.benchmarks` is passed as JSON alongside it and carries only the retained legacy demo entries, which have no Engine definition. An entry whose id the Engine also publishes is ignored, and the Job names it in its output.
 
 Disable seeding with:
 
--- a/apps/scoreboard/charts/scoreboard/templates/job-seed-benchmarks.yaml
+++ b/apps/scoreboard/charts/scoreboard/templates/job-seed-benchmarks.yaml
@@ -43,6 +43,13 @@ spec:
                   key: {{ .Values.database.existingSecretKey }}
             - name: SCOREBOARD_SEED_BENCHMARKS_JSON
               value: {{ .Values.seedBenchmarks.benchmarks | toJson | quote }}
+            {{- with .Values.seedBenchmarks.engineUrl }}
+            # The Engine whose benchmark catalog supplies every published benchmark's text
+            # (OME-904). One string, so an override can change WHICH Engine is read but cannot
+            # carry or omit the text itself.
+            - name: SCOREBOARD_SEED_ENGINE_URL
+              value: {{ . | quote }}
+            {{- end }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
 {{- end }}
--- a/apps/scoreboard/charts/scoreboard/values.yaml
+++ b/apps/scoreboard/charts/scoreboard/values.yaml
@@ -114,8 +114,16 @@ seedBenchmarks:
   # copy that used to live here was silently lost. An entry whose id the Engine publishes is
   # ignored by the seed job and named in its output.
   #
-  # Deployment-owned: set this to the Engine this board ranks submissions against. Left empty,
-  # the job seeds only the entries listed below.
+  # Deployment-owned: the Engine this board ranks submissions against.
+  #
+  # AIDEV-NOTE: this MUST be the Engine's in-cluster address (its Service DNS name), never the
+  # public hostname. The public host sits behind Cloudflare Access, which answers 200 with an
+  # HTML sign-in page rather than a 401 — verified 2026-08-20. The seed job then fails to parse
+  # it, falls back, refuses the stale configured entries, and exits zero, so the deploy looks
+  # clean while benchmark text silently stops being refreshed. The job logs a WARNING naming
+  # the content type when this happens.
+  #
+  # Left empty, the job seeds only the entries listed below.
   engineUrl: ""
   # Retained legacy demo entries (OME-775 D2). They predate the Engine benchmark catalogue and
   # have no Engine definition, so their text has nowhere else to live.
--- a/apps/scoreboard/charts/scoreboard/values.yaml
+++ b/apps/scoreboard/charts/scoreboard/values.yaml
@@ -101,43 +101,25 @@ seedBenchmarks:
   backoffLimit: 3
   ttlSecondsAfterFinished: 600
   annotations: {}
-  # The `revision` of each entry MUST match the Engine benchmark's computed REVISION exactly
-  # (apps/screamingface-engine/src/screamingface_engine/benchmarks/<name>/definition.py). A submission carries the
-  # Engine's value, and the board ranks per revision — a mismatch makes every real submission
-  # look incomparable to the registered board. Re-copy these whenever a benchmark's dataset or
-  # protocol revision changes; seeding is idempotent, so redeploying just moves the value.
+  # FEATURE: benchmark descriptions on the leaderboard (OME-904).
+  #
+  # A benchmark's text — display name, description, focus line, dataset link — is written ONCE,
+  # in the Engine benchmark definition it describes
+  # (apps/screamingface-engine/src/screamingface_engine/benchmarks/<name>/definition.py). The
+  # seed job reads it from that Engine's public catalog at deploy, so `revision` cannot drift
+  # from the Engine's computed value and no deploy-time override can blank the prose.
+  #
+  # AIDEV-NOTE: do NOT reintroduce descriptions for an Engine-published benchmark below. Helm
+  # replaces lists wholesale rather than merging them, which is exactly how the hand-copied
+  # copy that used to live here was silently lost. An entry whose id the Engine publishes is
+  # ignored by the seed job and named in its output.
+  #
+  # Deployment-owned: set this to the Engine this board ranks submissions against. Left empty,
+  # the job seeds only the entries listed below.
+  engineUrl: ""
+  # Retained legacy demo entries (OME-775 D2). They predate the Engine benchmark catalogue and
+  # have no Engine definition, so their text has nowhere else to live.
   benchmarks:
-    # --- Engine benchmarks (OME-775) --------------------------------------------------
-    - id: draco
-      display_name: DRACO
-      description: >-
-        A 100-task DRACO reproduction with official score arithmetic. It uses the successor
-        Judge model, provider-default reasoning, mixed native/Tavily retrieval, and a
-        host-only approximation of the reference blocklist, so its scores are not
-        paper-identical.
-      dataset_url: https://huggingface.co/datasets/perplexity-ai/draco
-      revision: 1c58b3085912e304
-      focus: Research reports with citations
-    - id: ifeval
-      display_name: IFEval
-      description: >-
-        The canonical 541-prompt instruction-following benchmark, graded by deterministic
-        strict and loose verification. Each Case invokes the Candidate exactly once.
-      # WHY no dataset_url: the IFEval dataset is vendored inside the Engine
-      # (screamingface_engine.benchmarks.ifeval.vendor), so no single public URL is authoritative.
-      revision: 0b88a52b5f10a6d9
-      focus: Instruction following
-    - id: healthbench-worst30
-      display_name: HealthBench Worst-30% Challenge
-      description: >-
-        The 157 hardest conversations from HealthBench Professional — the 30% that top
-        models score worst on. An AI judge grades each answer against a physician-written
-        rubric; safety mistakes subtract points, so per-case scores can be negative.
-      dataset_url: https://huggingface.co/datasets/openai/healthbench
-      revision: 6cd57aee171fbdc4
-      focus: Clinical safety, hardest cases
-    # --- Legacy demo entries, deliberately retained (OME-775 D2) -----------------------
-    # No Engine revision exists for these; they predate the Engine benchmark catalogue.
     - id: livetruth-latest
       display_name: News Livetruth Latest
       description: OpenMined Livetruth latest demo dataset
--- a/apps/scoreboard/pyproject.toml
+++ b/apps/scoreboard/pyproject.toml
@@ -5,6 +5,9 @@ description = "ScreamingFace benchmark scoreboard service"
 requires-python = ">=3.12"
 dependencies = [
     "fastapi>=0.141.1",
+    # WHY a runtime dependency, not a dev one: the seed job reads the Engine benchmark catalogue
+    # over HTTP inside the production image, which is built with `uv sync --frozen --no-dev`.
+    "httpx>=0.28.1",
     "pydantic>=2.13.4",
     "pydantic-settings>=2.14.2",
     "tortoise-orm[asyncpg]==1.1.8",
@@ -58,7 +61,6 @@ filterwarnings = [
 
 [dependency-groups]
 dev = [
-    "httpx>=0.28.1",
     "pyright>=1.1.411",
     "pytest>=9.1.1",
     "pytest-asyncio>=1.4.0,<2",
--- a/apps/scoreboard/src/scoreboard/scores/store.py
+++ b/apps/scoreboard/src/scoreboard/scores/store.py
@@ -323,6 +323,16 @@ class ScoreStore:
         )
         return benchmark_to_schema(benchmark)
 
+    async def has_registered_revision(self) -> bool:
+        """Has any benchmark row ever been registered with an Engine revision?
+
+        INVARIANT: only a benchmark the Engine publishes carries a revision — the retained
+        legacy demo entries have none — so a False answer means no successful Engine seed has
+        ever run against this database (OME-904).
+        """
+
+        return await Benchmark.filter(revision__isnull=False).exists()
+
     async def list_benchmarks(self) -> list[BenchmarkSchema]:
         rows = await Benchmark.all().order_by("id")
         return [benchmark_to_schema(benchmark) for benchmark in rows]
--- a/apps/scoreboard/src/scoreboard/seed.py
+++ b/apps/scoreboard/src/scoreboard/seed.py
@@ -5,7 +5,9 @@ import asyncio
 import json
 import os
 from collections.abc import Sequence
+from dataclasses import dataclass, field
 
+import httpx
 from pydantic import BaseModel, ConfigDict, Field, TypeAdapter, ValidationError
 
 from .config import Settings
@@ -14,6 +16,23 @@ from .scores.schemas import BenchmarkSchema
 from .scores.store import ScoreStore
 
 SEED_BENCHMARKS_ENV = "SCOREBOARD_SEED_BENCHMARKS_JSON"
+# FEATURE: benchmark descriptions on the leaderboard (OME-904). The Engine's benchmark
+# definitions are the ONLY place a benchmark's prose is written; this deployment names which
+# Engine to ask, and nothing else. A values override can therefore change the Engine consulted
+# but can no longer carry, alter, or omit the text itself.
+ENGINE_URL_ENV = "SCOREBOARD_SEED_ENGINE_URL"
+_CATALOG_PATH = "/v1/benchmarks"
+# The catalogue is a small static document behind an ETag; a deploy hook should not hang on it.
+_FETCH_TIMEOUT_SECONDS = 15.0
+
+
+class EngineCatalogUnavailable(RuntimeError):
+    """The Engine benchmark catalogue could not be read.
+
+    INVARIANT: every transport, status, and payload failure surfaces as this one type. No
+    `httpx` or `json` exception escapes this module, so a deploy log names the thing that
+    failed rather than leaking a library's internals.
+    """
 
 
 class SeedBenchmark(BaseModel):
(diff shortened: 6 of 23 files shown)


khoaguin:
# Review — verified findings

**Verdict: hold the merge.** One bug makes every deploy fail outright, and three more let the exact bug this PR fixes come back silently. Everything below was verified against the code (and, where claimed, actually executed), not just eyeballed.

---

## 🔴 1. The seed job crashes before it starts — every deploy fails

`seed.py` now does `import httpx` at the top of the file. But httpx is listed only in scoreboard's **dev** dependency group, and the production image is built with `uv sync --no-dev` — meaning "install everything *except* the dev group."

Concretely: every `helm install`/`upgrade` runs `python -m scoreboard.seed` as a hook Job → `ModuleNotFoundError: httpx` on line 1 of execution, before it even reads its arguments → the Job retries 3 times, fails, and **the whole Helm release fails**. CI is green because tests install the dev group; nothing builds the production image on PRs, so this only shows up at deploy time.

**Fix:** move `httpx` from `[dependency-groups].dev` to `[project].dependencies` in `apps/scoreboard/pyproject.toml`.

## 🔴 2–4. "Engine is the only copy" only holds when everything goes right

The PR's core promise is: Engine text always wins, and a deploy override can't blank the board again. That promise is enforced on the happy path only. Three related holes, one shared root cause — **configured entries are still allowed to carry a `revision`**:

**2. An Engine outage lets stale values overwrite good rows.** When the catalogue fetch fails, the "published by the Engine" set is empty — so the shadowing rule ("Engine rows win over configured entries with the same id") has nothing to shadow *with*. The deployed values override still carries draco/ifeval/healthbench entries (stale revisions, no prose). One transient Engine blip during any later upgrade and those entries sail straight through: the good Engine-seeded rows get overwritten with `description = NULL` and old revisions, and the job exits 0. That is literally the OME-904 bug, reproduced by the fix. The existing unreachable-Engine test doesn't catch it because its configured entry uses an id the Engine doesn't publish.

**3. The bootstrap guard checks the database *after* writing to it.** The only non-zero exit is "no row anywhere carries a revision" — meant to catch a first deploy where the Engine was never reached. But the check runs *after* this same pass has seeded the configured rows. If those rows carry revisions (today's real deploy values do), the guard sees the revisions it just wrote milliseconds earlier and concludes everything is fine. It's asking "did anyone bring food?" after putting your own empty lunchbox on the table and counting it.

**4. The most likely misconfiguration is completely silent.** `engineUrl` defaults to empty. If the deploy repo upgrades the chart without adding it (very plausible — the deploy repo predates the key, and Helm *replaces* lists, so its old benchmarks list also replaces the new defaults), then: nothing is fetched, no error is recorded, the old stale list seeds, exit 0, zero warnings. Green deploy, null-description board — the original bug, again, with no log line. A configured entry carrying a revision while `engineUrl` is unset is mechanically detectable and should at minimum warn loudly.

**One fix covers most of this:** drop `revision` from `SeedBenchmark` entirely. The legacy demo entries don't have revisions, and the store's own docstring already states the invariant "only a benchmark the Engine publishes carries a revision" — it's just asserted, not enforced. Removing the field makes findings 2 and 3 structurally impossible and turns finding 4's smoking gun into a schema error.

## 🟡 5–8. Sharp edges on the catalogue fetch

**5. One bad entry rejects the whole catalogue.** Scoreboard's `_CatalogEntry` caps title at 255 chars, focus at 120, revision at 64 — but the Engine side doesn't enforce any length caps. So an Engine author can write a 130-char focus line that passes every Engine test ("one authoring site" means they never run scoreboard validation), and at the next deploy the *entire* catalogue — including all the untouched valid benchmarks — is refused as one validation failure. Board silently keeps old text, exit 0. Either match the caps on the Engine side, or skip-and-report the bad entry instead of rejecting the batch.

**6. A redirect makes the Engine permanently "unavailable."** The httpx client doesn't set `follow_redirects=True`, and httpx's default is to *not* follow them (verified: `raise_for_status()` raises on a 301). Any http→https or trailing-slash redirect at the ingress — extremely common — means every seed run forever reports "answered HTTP 301" and keeps the old text, exit 0.

**7. Invalid UTF-8 escapes the error handling and blames the wrong thing.** `response.json()` on a 200 body with broken UTF-8 raises `UnicodeDecodeError`, not `JSONDecodeError` (verified by executing it). None of the fetch's `except` clauses catch it — but `UnicodeDecodeError` *is* a `ValueError`, so it falls into `main()`'s `except ValueError: parser.error(...)`, and the deploy log prints a **command-line usage error** for what is actually a mangled Engine response. Exit 2, survivable-failure path bypassed.

**8. No retry on the single GET.** "Engine unreachable → exit 0" is the right call for a populated board, but it also means Kubernetes never retries the Job (backoffLimit only fires on non-zero exits). A Helm upgrade that rolls Scoreboard and Engine together can land the seed pod's one GET exactly in the Engine's restart window — and whatever changed in that Engine release stays off the board until the *next* deploy. Two or three in-process retries with a short sleep would cover the race

khoaguin:
All eight findings verified and fixed in `88df8b5c`. Two I confirmed by executing rather than reading, and one of those changed how I fixed it.

## 🔴 1 — httpx dev-only: confirmed, release-breaking

`apps/scoreboard/pyproject.toml:61` had it under `[dependency-groups].dev`; `apps/scoreboard/Dockerfile:26` builds with `uv sync --frozen --no-dev`. Exactly as described — the hook Job would have died on import.

Moved to `[project.dependencies]` and relocked. Also added the only test that can defend this, since no runtime test can (tests install the dev group):

```python
async def test_httpx_is_a_runtime_dependency_not_a_dev_one() -> None:
    manifest = tomllib.loads((… / "pyproject.toml").read_text(encoding="utf-8"))
    runtime = {…}
    assert "httpx" in runtime
```

## 🔴 2–4 — confirmed, fixed by one rule

Your diagnosis was right and so was the root cause: configured entries were allowed to assert a `revision`. I did **not** take the suggested fix of deleting `revision` from `SeedBenchmark`, because three prior-cycle tests in `test_seed.py` assert it and removing them is a rule-5 violation. The same protection without touching them:

> **A configured entry is refused if it asserts a `revision` the Engine did not publish in this pass, or if its id names a row that already carries one.**

That closes all three. An Engine blip can no longer let stale entries overwrite good rows (2); the guard now reads the prior state *before* seeding, so it cannot count its own writes (3); and an unset `engineUrl` now names every refused entry loudly instead of silently seeding the old list (4).

Your lunchbox analogy for #3 was exact — the check ran after `seed_benchmarks()` in the same function.

## 🟡 5 — confirmed, fixed on both sides

Entries now validate independently: one unusable entry is reported and skipped, siblings still refresh. **And** the Engine enforces the board's widths (`title` 255, `focus` 120, `revision` 64), so an over-long focus line fails at authoring time — the failure belongs where the text is written, not at deploy.

## 🟡 6 — confirmed by execution

```
follow_redirects default = False
301 -> raise_for_status() raises
with follow_redirects=True -> 200
```

Set per-request, so it applies to an injected client too.

## 🟡 7 — confirmed by execution, and it changed the fix

```
F7 raises: UnicodeDecodeError | is JSONDecodeError: False | is ValueError: True
```

Caught alongside `JSONDecodeError`. But the deeper problem was the CLI: `main()` wrapped the entire run in `except ValueError: parser.error(...)`. I moved the JSON parse out of the run so `parser.error` only ever describes a bad argument — otherwise the next unrelated `ValueError` would have found the same escape hatch.

## 🟡 8 — confirmed

Three attempts with a delay, retrying only what a retry can fix: transport failures and 5xx. A 4xx and a mangled body fail immediately.

---

**Note on the diff:** fixing 5 required changing one already-committed test — batch-rejection was the wrong expectation, so `…_is_reported_as_a_seed_error` became `…_is_reported_and_skipped`, plus a new test for the envelope case that must still raise. Owner-approved under sdlc rule 5; the remaining test edits are mechanical (`fetch` returns a `CatalogRead` so it can report rejects) and `retry_delay=0` so failure tests don't sleep through the real backoff (28s → 0.07s).

Both stacks: ALL GATES GREEN. Thanks — #1 alone would have broken the release.

khoaguin:
Rebased onto `main` — the conflict was with #658, which refactored HealthBench into a `healthbench_benchmark()` factory producing two boards. My change had edited the inline `Benchmark(...)` that refactor replaced.

Resolved by threading the display metadata through the factory rather than special-casing either board:

```python
def healthbench_benchmark(
    *, id, title, description, case_ids, protocol_revision, scoring, mean, selection_sha,
    focus: str | None = None,
    dataset_url: str | None = None,
) -> tuple[Exam, Benchmark]:
```

Both boards now pass them, sharing one `HEALTHBENCH_DATASET_URL` constant so the dataset link isn't duplicated across the pair. The registry test asserts all four benchmarks, so a fifth board added without a focus line fails the suite instead of quietly rendering an em-dash.

Verified against the real catalogue, not just the tests:

```
draco                      focus='Research reports with citations'      dataset=…/perplexity-ai/draco
healthbench-professional   focus='Clinical safety, full official exam'  dataset=…/openai/healthbench
healthbench-worst30        focus='Clinical safety, hardest cases'       dataset=…/openai/healthbench
ifeval                     focus='Instruction following'                dataset=—   (vendored)
```

**Worth noting:** `healthbench-professional` is brand new and has never appeared in `values.yaml`. Under the old arrangement it would have shipped with a null description like the others. Under this one it needs no chart entry at all — its text reaches the board because it exists in the registry. First new benchmark since the change, and it needed zero deploy-config work.

One editorial call for a reviewer: the two HealthBench boards share a dataset and a description style, so the focus line is the only place a reader sees the difference — "full official exam" vs "hardest cases". Easy to reword.

Both stacks: ALL GATES GREEN, including the append-only check.

khoaguin:
CI caught a caller I hadn't looked for, and the type error was the smaller half of it.

`packages/screamingface/_runtime/server.py:141` imports `scoreboard.seed._run`, whose signature this PR changed — pyright flagged the missing argument. But that local runtime seeds its board from `scoreboard_seed_json(BUILTIN_BENCHMARKS)`, and those rows carry revisions. Under this PR's new rule — *configuration may not assert a revision the Engine did not publish* — every one of them would have been **refused**, and a local leaderboard would have come up empty. Fixing only the signature would have shipped a green build and a broken local stack.

The gap was in the API, not the call site. That path is doing exactly what this PR does — derive the board's catalogue from the Engine registry — just in process, because a local stack runs the Engine and the board in one virtualenv. It had no door to come through, so it was arriving as `configured`, which is precisely what the refusal rule exists to reject.

So there are now two adapters onto the same catalogue:

```python
# deployment: read it over HTTP
seed_from_sources(engine_url="https://engine.dev…", configured=legacy_rows)

# local stack: hand in the imported registry, nothing is fetched
seed_from_sources(engine_url=None, engine_rows=registry_rows, configured=[])
```

Rows arriving either way are Engine-owned: they win over configuration and are never refused. The local projection also carries `focus`/`dataset_url` now, so a local board shows what the deployed one shows instead of quietly less.

**The process miss worth naming:** three stacks in this repo import `scoreboard.seed`, and I ran gates for two. Changing a signature means grepping callers across the monorepo, not across the app that owns the file. All three green now — `screamingface-engine`, `scoreboard`, `screamingface`.

FilipBolt (approved):
Approving — the design is right and it fixes a live problem. Four things below, none blocking, but the first needs sorting before this reaches a cluster.

**Disclosure:** I added `Benchmark.focus` in OME-874 yesterday, seeded from `values.yaml`. This supersedes that, and I think it's the better call — one authoring site next to the code it describes beats a hand-copied second one. Reviewing a change to my own work, so weigh accordingly.

**It fixes a live bug**, not a hypothetical: DRACO on dev is registered at `1c58b3085912e304` while Keelan's submission from last night carries `66a463248586b277`, so the board shows **0 entries** while the score sits happily on its spec page. `EVALUATION_PROTOCOL_REVISION` changed on 2026-08-16 (`c3f3f97a`), hours after the revisions were seeded. Your `revision` -sourced-from-the-Engine change makes that class of drift structurally impossible. Note the reseed is still needed today — this stops it recurring, it doesn't heal the existing row.

---

## 1 · Set `engineUrl` to the in-cluster address, not the public hostname

`https://fusion.dev.screamingface.ai/v1/benchmarks` answers **HTTP 200 with a Cloudflare Access sign-in page** — HTML, not JSON. Traced through this code, the public host gives:

1. `raise_for_status()` passes — it is a 200
2. `.json()` raises → `EngineCatalogUnavailable`, no retry (correct — a parse failure will not self-heal)
3. `report.engine_error` set, fall back to configured rows
4. configured `draco`/`ifeval`/`healthbench-worst30` all carry a `revision` → **refused** by Stage 3
5. job exits **zero**

So the feature is permanently off and the deploy looks clean. Everything there behaves exactly as designed — the design just assumes a reachable catalogue, and the obvious URL to paste is the one that fails.

Two suggestions: make the `values.yaml` comment say the address must be in-cluster because the public host sits behind Access, and consider naming the fallback in the log at warning level rather than as ordinary output. Right now the one path that reintroduces stale revisions is also the quietest.

## 2 · Nothing exercises a real Engine response

Every test injects a fake client, which is good design for the retry and precedence logic. But the bug being fixed here **was** a deploy-time path no test executed, and this fix has the same shape: `_CatalogEntry`'s field names and the `{"object": "list", "data": [...]}` envelope are only ever checked against payloads written in this repo.

I verified both against `rest/benchmarks.py` and `catalog_entry()` on `main` and they match today — `id`, `title`, `description`, `revision` are all published, and `focus`/`dataset_url` being optional keeps an older Engine working. So the current shape is fine. A recorded real response as a fixture would keep it that way.

## 3 · Blocking I/O inside an async function

`seed_from_sources` is `async`, but `fetch_engine_benchmarks` uses a sync `httpx.Client` and `time.sleep`. Worst case is roughly 49s of blocked event loop (3 × 15s timeout + 2 × 2s sleep). Harmless in a one-shot Job with nothing else on the loop — worth a note only because the function is public and reads as safe to call from the app, where it would not be.

## 4 · An empty description drops the whole benchmark

`description: str = Field(min_length=1)` means an entry with empty prose fails validation entirely, so it lands in `rejected` and never registers — no row, no board, and a submission to it would fail rather than merely look plain. Existing rows survive because seeding upserts, so this only bites a *new* benchmark.

Probably what you intended (an undescribed benchmark is not ready to publish), but the failure is a missing board rather than missing text, which is a bigger consequence than the field suggests. Worth a line in the docstring either way.

---

Things I checked that hold up: the envelope matches the Engine's real response; `_CatalogEntry`'s required fields all exist on today's `catalog_entry()`, so this is backward compatible with an un-deployed Engine; Stage 1 reading `has_registered_revision()` *before* writing is the right call, since reading after would let the pass count its own writes; and the Stage 3 double refusal is the part that actually prevents a repeat — a configured entry asserting a revision cannot overwrite an Engine-owned row even during an outage. The worked example using `1c58b30…` is a nice touch given that is the exact value stranding Keelan's run right now.

khoaguin:
Thanks Filip — all four addressed in `42164795`, and #1 verified against the live host before I touched anything.

## 1 · Confirmed, and it's worse than a bad default

```
$ curl -sIL https://fusion.dev.screamingface.ai/v1/benchmarks
status=200  content_type=text/html  size=33128
<!DOCTYPE html><html><head><title>Sign in ・ Cloudflare Access</title>
```

Your trace is exactly right, and the part that makes it nasty is that **nothing in the code is wrong** — every layer behaves as designed and the sum is a silently disabled feature. So I treated it as a message problem, in three places:

- The parse failure now names the content type and the likely cause:
  `answered content-type text/html; charset=utf-8 instead of JSON (…). A public hostname behind an authentication proxy answers 200 with a sign-in page: engineUrl must name the in-cluster Engine address, not the public one.`
- The fallback logs at **WARNING**, and says the consequence rather than just the cause: `benchmark text was NOT refreshed this deploy; the board is serving whatever the last successful seed wrote`. You were right that the quietest path was the one that turns the feature off.
- `values.yaml` records why the public hostname is wrong, so the next person doesn't rediscover it.

I did **not** set a default value, since I don't know the in-cluster Service DNS name. If you do, it's a one-line change and I'd rather ship it than leave the field empty.

## 2 · Fair hit, and the shape of it was the point

You're right that every test fed the parser a payload written in this repo — proving the parser agrees with itself, which is the same shape as the bug being fixed. Added `tests/fixtures/engine_catalog.json`, captured from the actual producer (`catalog_entry()` over the real registry, the exact body `rest/benchmarks.py:53` assembles), and a test that parses it end to end.

It passed on the first run — so this is the confirmation that was missing, not a fix. It also pins the thing you checked by hand: the fixture carries `case_count`, `href` and `check_surface`, and the test asserts meeting those never costs a benchmark its row.

## 3 · Documented, not changed

~49s worst case is right (3 × 15s + 2 × 2s). Left synchronous — correct for a one-shot Job — but your point that `seed_from_sources` being `async` makes it *look* safe to call from the app is the real risk, so that's now an `AIDEV-NOTE` on the function telling the next person to convert to `AsyncClient` first.

## 4 · Intended, now stated

Dropping the benchmark is deliberate — an undescribed benchmark isn't ready to publish — but you're right the consequence (no row, no board, submissions refused) is much larger than "missing text" suggests. Stated at the field, including that existing rows survive because seeding upserts, so it only bites a *new* benchmark.

---

On the disclosure: no awkwardness at all — `focus` existing at all is why the portal column was ready to fill, and this only moves where it's authored. And thanks for the DRACO/Keelan detail; that's a much better argument for the change than the one in my description, since it's a submission stranded right now rather than a maintenance hazard. Agreed the reseed is still needed today — this stops the recurrence, it doesn't heal the existing row.

All three stacks green: `scoreboard`, `screamingface-engine`, `screamingface`.

khoaguin:
Two follow-ups in `b398ad93`.

## #4 — reversed: a benchmark without prose keeps its row

Owner call, and the right one. Rejecting the entry costs the benchmark its whole row — no board entry, submissions refused — which is a much worse failure than a row that reads plain. `description` is now optional on the consumer side, blank normalises to absent.

Deliberately **NULL rather than a placeholder sentence**: `_ui/leaderboard_view.py:148` already renders `"No description published."` for an absent one, so the reader-facing wording stays in the single place that owns it — and prose written into the database could never be told apart later from prose an author actually wrote.

Leniency stops where it should: an entry missing `id`, `title` or `revision` still fails, because without those there is nothing to register or rank. And the Engine still *refuses* to define a benchmark without a description. Require it where the text is written; tolerate it where it is read.

## #2 — the honest version

I tried to do what you'd actually want here — one test running both halves in process — and it doesn't work. The Client SDK's venv has both packages' *source* on its path, so `find_spec` says yes, but none of their dependencies are installed (not even pydantic); its local runtime installs those at use time. Verified rather than assumed:

```
fastapi                                   ModuleNotFoundError
pydantic                                  ModuleNotFoundError
screamingface_engine.benchmarks.builtins  ModuleNotFoundError: No module named 'pydantic'
```

So there is no environment where both apps run, by design. What I did instead:

- **The producer pins the contract on its own side** — `test_the_catalog_keeps_the_field_names_the_leaderboard_seeds_from` asserts the catalogue keeps the exact names the board seeds from, and its docstring says which file breaks and how (the benchmark lands in `rejected` and vanishes from the leaderboard while the Engine's own tests stay green). The producer is the side that can break the contract, so it's the side that pins it.
- **Each half names the other**, so a rename finds both in one grep.
- The recorded response stays as the consumer-side check.

**One gap left, and it's yours not mine:** the SDK lane is path-filtered to `packages/**`, so an Engine-only PR never runs these seam tests. Adding `apps/screamingface-engine/**` to that filter would close it — CI ownership, so I've flagged it rather than changed it.

## #1 — was already in `42164795`

Both your suggestions landed in the previous push; flagging in case the diff order hid them:

```yaml
# AIDEV-NOTE: this MUST be the Engine's in-cluster address (its Service DNS name), never the
# public hostname. The public host sits behind Cloudflare Access, which answers 200 with an
# HTML sign-in page rather than a 401 — verified 2026-08-20. …
```

```
WARNING engine catalog unavailable: {…}
WARNING benchmark text was NOT refreshed this deploy; the board is serving whatever the last successful seed wrote
```

All three stacks green.

khoaguin:
Rebased again — `main` had moved 24 commits, and DRACO went the same way HealthBench did in the last rebase: refactored into a `draco_benchmark()` factory, now producing two boards (canonical five-pass plus `draco-3pass` for cache-seeded replays, #674 / `4f072b21`).

Resolved identically to the HealthBench one — `focus`/`dataset_url` threaded through the factory rather than special-cased per board, both call sites passing them, one shared `DRACO_DATASET_URL` constant so the link isn't duplicated across the pair.

The catalogue now publishes five boards:

```
draco                     focus='Research reports with citations'
draco-3pass               focus='Research reports, three judge passes'
healthbench-professional  focus='Clinical safety, full official exam'
healthbench-worst30       focus='Clinical safety, hardest cases'
ifeval                    focus='Instruction following'
```

Both sibling pairs share a dataset and a subject, so in each case the Focus line carries the only difference a reader can actually see — judge-pass count for DRACO, scope and metric for HealthBench. Reword either if they read wrong to you; they're one-liners.

## One test changed, and the reason is worth a sentence

`draco-3pass` is the **third** benchmark to land while this PR has been open, and every one of them broke the recorded-response test — because that test asserted a hardcoded roster of benchmark ids. Red for ordinary work, not for a defect.

It now asserts the invariant it was reaching for: every benchmark in the Engine's response comes out the other side, whatever they are.

```python
assert read.rejected == []
assert len(read.rows) == len(payload["data"])
assert {row.id for row in read.rows} == {entry["id"] for entry in payload["data"]}
```

A rename or an outgrown column still fails it. Adding a board no longer does. The reasoning: a test that goes red for normal work teaches people to make it green by editing the test — which is exactly the reflex you don't want it to meet when it eventually goes red for a real reason.

**The trade, stated plainly:** a benchmark silently *disappearing* from the Engine no longer fails this test, since both sides shrink together. That now surfaces as a deletion in the regenerated fixture, where review can see it. If you'd rather have roster coverage back, the honest place for it is a separate assertion about the fixture, not bolted onto this one.

All three stacks green.

PR #672, opened 2026-08-20
Bound the event bridge by a memory budget, not an event count

IonesioJunior:
## Summary

**This PR is the real fix for OME-906** — unlike #667, merging it SHOULD close that issue.

The 8 192-event count cap was a ceiling on **DAG width**, not backlog health: the engine
gathers over `deps` unbounded and emits each node's `NodeStarted` before it awaits
anything, so a wide fan-in lands its whole event burst in one event-loop slice, before
the drain can run at all. A 100-Case DRACO Fusion is legitimately ~3 500 nodes wide and
sat at the limit. The cap defended **2.1 MB** in a process that accepts a **1 GiB** result.

Two changes:

1. **The hard cap is now a memory budget.** `budget // 512` events (default 64 MiB ≈
   131 072 events), readable from `URL4_CLOUD_BRIDGE_MEMORY_BUDGET_BYTES` exactly like
   the result caps — an operator bounds what the backlog may **cost**, not how wide a DAG
   may be.
2. **The error says which failure shape fired**, keyed on a new `drained` counter beside
   #667's high-water mark: drain progress above zero is one DAG burst wider than the
   budget (names the knob and the DAG); **zero drained events** is a stuck consumer — the
   one case where the old "consumer is not keeping up" accusation was right.

## Stacked on #667 (GitHub stack #673)

The `drained` counter sits beside the high-water mark **PR #667 introduces**, so this
branch is based on `OME-906-pipelined-frame-publishing`, not `main`. The two PRs are
formally linked as a native GitHub stack: review #667 first — merging it auto-rebases and
auto-retargets this PR to `main` (no manual retarget needed), or merge from this PR to
land both layers in one operation.

## Evidence

- **RED-first:** a 9 000-dep fan-in through the real engine reproduced the production
  failure deterministically (`BridgeOverflowError ... 8192 ... consumer is not keeping up`)
  before the fix, and completes after it.
- A never-stopping producer still fails — at the budget (verified with a 10 240-byte
  budget → 20 events), not at a magic count.
- Local gates (`screamingface-engine` stack): ruff, ruff format, pyright,
  `check_layering.py`, pytest — **1 890 passed, 5 skipped, coverage 93.57 %** (gate 80 %).

## Notes for review

- Two pre-existing hard-cap tests were given explicit small budgets: the default cap
  moved 8 192 → 131 072 and made them pathological (one took **838 s**; now 0.26 s).
- The 64 MB default is the spec's judgement call (30× the widest measured burst, 1/16 of
  the result hard cap); it is deploy-time tunable, so a tighter pod can lower it without
  a release.

Spec `docs/spec/2026-08-20-OME-906-bridge-memory-budget.md` · plan
`docs/plan/2026-08-20-OME-906-bridge-memory-budget.md` · ledger
`docs/work/2026-08-20-OME-906-bridge-memory-budget.md`

Fixes OME-906

### DIFF
--- a/apps/screamingface-engine/src/screamingface_engine/job_env.py
+++ b/apps/screamingface-engine/src/screamingface_engine/job_env.py
@@ -253,6 +253,16 @@ RESULT_HARD_CAP_BYTES = "URL4_CLOUD_RESULT_HARD_CAP_BYTES"
 spilling. The result is already in Runner RAM when checked (string + encoded copy ≈ 2-3× its
 size), so size this to pod memory, not disk."""
 
+BRIDGE_MEMORY_BUDGET_BYTES = "URL4_CLOUD_BRIDGE_MEMORY_BUDGET_BYTES"
+"""Memory ceiling for the runner's event-bridge backlog: past it the run FAILS with
+``BridgeOverflowError`` instead of buffering on. Converted to an event count at 512 B per
+event (see ``runner.executor.EVENT_SIZE_ESTIMATE_BYTES``) — a bound on what the backlog may
+COST, not on how wide a DAG may be (OME-906)."""
+
+DEFAULT_BRIDGE_MEMORY_BUDGET_BYTES = 67_108_864
+"""64 MiB ≈ 131 072 events at the 512 B estimate — 30× the widest burst measured (OME-906)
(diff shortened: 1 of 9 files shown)


keelancj (approved):
Approved, fix is looking good and the cached-DRACO overflow issue isn't happening anymore 🙌

PR #607, opened 2026-08-14
Harden the ScreamingFace runtime CLI

itstauq:
## What this PR does

#590 added the pip-installable local runtime. This follow-up makes that runtime safe and comfortable to operate as a normal CLI instead of a thin process launcher.

A researcher can manage the complete local stack with:

```bash
screamingface up
screamingface status
screamingface doctor
screamingface logs
screamingface restart
screamingface down
```

`up` starts AI Gateway, Scoreboard, and the ScreamingFace Engine in the background. The SDK continues to default to hosted services; the CLI prints explicit `SCREAMINGFACE_ENGINE_URL` and `SCREAMINGFACE_SCOREBOARD_URL` exports when a notebook should use the local runtime.

### Safe lifecycle ownership

The runtime no longer assumes that a PID read from disk belongs to ScreamingFace. It creates a private authenticated loopback control endpoint and stores a versioned, user-private state document containing the process identity, control endpoint, resolved service URLs, log path, and start time.

`down` and `restart` authenticate with that endpoint before requesting shutdown. Unreachable or mismatched state is treated as stale without signalling its PID, preventing PID reuse from stopping an unrelated process. Repeated `up` adopts a healthy owned runtime, reports partially healthy runtimes clearly, and refuses foreign port conflicts.

### Configuration and machine-readable status

Ports can be configured independently through flags or matching environment variables:

```bash
screamingface up \
  --gateway-port 19105 \
  --scoreboard-port 19106 \
  --engine-port 19108
```

CLI flags override environment variables, which override the defaults. The resolved Gateway URL is propagated into Engine discovery and execution. `status --json` emits a stable, ownership-token-redacted `screamingface.runtime-status.v1` document. `--data-dir` works before or after a subcommand, and `--version` reports the installed distribution version.

### Diagnostics and benchmark assets

`doctor` performs read-only, no-spend checks for runtime dependencies, writable storage, state ownership, ports, service health, model discovery, provider connections, and benchmark assets. Missing optional setup is a warning; corrupt state or unhealthy owned services is a failure.

Benchmark preparation writes a manifest only after validating its output:

```bash
screamingface prepare --list
screamingface prepare draco
screamingface prepare draco --force
```

Valid current preparation is skipped, partial downloads remain available for retry, and stale assets appear in `prepare --list`, `doctor`, and JSON status. Dataset fingerprints come from the Engine-owned benchmark preparation modules, keeping manifests aligned with the Engine's pinned dataset revisions.

### Service-aware rotating logs

[...]

PR #676, opened 2026-08-20
fix(screamingface): warn on partial leaderboard submissions (OME-922)

keelancj:
Warns when a limited or incompletely graded Candidate is submitted to the public Leaderboard,
without letting the advisory disappear on the documented facade, inside notebooks, or after a
saved-result round trip.

The Scoreboard request remains unchanged. Headless callers receive `sf.EvaluationWarning` at
their own source line before the write; notebooks receive one branded display event after the
Scoreboard confirms the write.

**Linear:** https://linear.app/openmined/issue/OME-922/warn-on-partial-submission-that-only-full-runs-are-ranked

### Before

- **Trigger:** a researcher evaluates with `limit=N`, or loses some Case grades, then publishes
  the resulting Candidate.
- **Today:** the selected Cases can have `coverage == 1.0` while still representing only part of
  the Benchmark. The resulting score looks comparable with a complete run even though it is not.
- **Documented-path gaps found in review:**
  - `sf.leaderboards.submit(...)` blamed the SDK wrapper rather than the user's line and collapsed
    repeated advisories under Python's default warning filter;
  - assigning the returned score in a notebook produced no visible notice;
  - exporting and reloading a limited result replaced the Benchmark total with the selected count,
    so a 2-of-3 run returned as an apparently complete 2-of-2 run;
  - warnings-as-errors could persist the score first and then raise, hiding the returned score id;
  - the advisory claimed the score "may appear on the public leaderboard" although it is emitted
    before the POST, so a failed write said the score was on the board;
  - completeness leaned on the Engine's 4-dp `coverage`, which rounds to `1.0` on a large
    benchmark missing a handful of grades;
  - in a notebook the advisory could be neither escalated with `-W error` nor suppressed with
    `-W ignore`;
  - Colab/Databricks shells were misclassified as terminals;
  - global persimmon tokens recoloured existing Report warnings toward the error palette.

### After

- **Same trigger.** A Candidate is complete only when its Case count equals the Benchmark total
  and Engine-owned grading coverage is exactly `1.0`.
- **Headless:** the exact advisory is emitted as `sf.EvaluationWarning`, attributed past every SDK
  frame to the user's call site. Three submissions on three source lines produce three warnings.
- **Warnings-as-errors:** the warning happens before HTTP in **both** environments, so an error
  policy stops before the Scoreboard changes. In a notebook the warning is emitted inside a
  recording context, so the caller's filters apply there too — `-W error` aborts, `-W ignore`
  silences the branded notice — while IPython's stderr block stays suppressed.
- **Notebook:** one accessible, branded display event is emitted after a successful write even if
  the return value is assigned, collected, or captured by papermill/nbconvert. A final-expression
  score card does not repeat it.
- **Saved result:** Candidate export retains the complete Benchmark Case count while the Report
  root retains the selected Case count, so the documented loader still identifies a limited run.
- **Hosted notebooks:** Colab and Databricks subclasses are recognised through their ipykernel
  base class.
- **Colour:** persimmon is scoped to the new submission notice; existing Report warnings retain
  their distinct amber palette.

## Don't regress

- The Scoreboard payload, endpoint, idempotency key, ranking policy, and response contract do not
  change.
- With ordinary warning policy, valid partial scores are still sent unchanged.
- A warning configured as an error aborts before the POST, in notebooks as well as headless; this
  is normal Python warning behavior, made side-effect safe.
- Notebook presentation occurs only after a successful POST. Failed writes display no success
  advisory.
- Rich display is presentation, never authority: if the notebook display publisher itself fails,
  plain stderr carries the advisory and the already-persisted `LeaderboardScore` is still returned.
- Full, fully graded Candidates remain silent.
- Unscored or otherwise invalid Candidates continue to fail existing validation before HTTP.
- Sync and async submission share the same policy and are tested independently.
- The returned `LeaderboardScore` has no transient notice field: equality, hashing, copying,
  pickling, repr, and retrieval stay unchanged.

## Why the original shape missed real users

The first test group exercised `client.leaderboards.submit(...)` directly and inspected the
returned score card. The documentation and notebooks predominantly use another composition:

```python
submission = sf.leaderboards.submit(report.candidates.only)
```

That public facade adds a stack frame, and assignment suppresses IPython's final-expression
renderer. Saved-result publication adds a third boundary through `report.export()` and
`examples/helpers.py`. Each component looked correct alone; the user workflow was not covered.

The replacement tests therefore start at those public boundaries instead of calling the notice
decorator or inspecting fields that exclude notices by definition.

### One focused submission policy

`_scoreboard/submission_notice.py` owns the complete decision:

```text
selected Case count != Benchmark Case count
                     OR
Engine grading coverage < 1.0
```

It returns the process-local rich notice for an active notebook, or emits the headless warning.
Python 3.12's `skip_file_prefixes` skips the entire SDK package rather than guessing one fixed
`stacklevel`, so both module-level and explicit-Client APIs identify the caller correctly.

### Side-effect-safe ordering

Submission now has an explicit order:

1. validate and construct the existing payload;
2. determine the advisory and emit a headless warning;
3. POST and decode the persisted score;
4. display a reserved notebook notice;
5. return the ordinary score.

That ordering means a warning exception cannot follow a successful write. Notebook display is
guarded separately because it occurs after persistence and therefore must never hide the id.

### Notebook output independent of expression position

`_ui/notice_view.py` sends a normal IPython display event instead of storing presentation state on
`LeaderboardScore`. Assignment, list construction, papermill, and nbconvert all retain display
events. When the score is also the cell's final expression, IPython renders the unchanged score
card after the notice, with the advisory present exactly once.

The warning notice uses:

- canonical persimmon light/dark values from `OpenMined/screamingface-brand` commit `7ea35a1`;
- a visible solid square and visible title/body;
- `role="alert"` for the dynamically inserted warning;
- outer-element `data-notice-code` and `data-notice-severity` attributes;
- escaped copy and attributes;
- severity-specific CSS rather than painting every future severity as a warning.

Existing Report warning tokens and Report rendering are unchanged from `main`.

### Saved-result identity remains truthful

`CandidateResult.to_dict()` now serializes its complete `BenchmarkInfo.case_count`. The enclosing
Report document continues to serialize `Report.case_count` at the root, preserving the selected
Evaluation size. Consequently:

```text
serialized Report root benchmark.case_count       selected count
serialized Candidate benchmark.case_count         complete Benchmark count
```

The documented `load_candidate_result(...)` helper reads the Candidate-owned value and can still
warn when yesterday's expensive run was limited.

### Hosted notebook detection

`running_in_notebook()` checks the shell class's MRO, not only its concrete module. A Colab or
Databricks shell defined in a host-specific module still inherits from ipykernel and receives the
branded display rather than a red Python warning block.

### Structured notice cleanup

`ClientNotice` trims its code/title/body before use, so whitespace or a trailing newline cannot
break stable HTML identity. The abandoned private score field, duplicated Report CSS, global
palette migration, and insensitive leaderboard test block are removed from the final diff.

### Red → green public regressions

The focused public-workflow module has 14 tests covering:

- three documented facade calls on three lines produce three exactly attributed warnings;
- exact message and `sf.EvaluationWarning` category;
- limited selection and incomplete grading independently;
- defensive over-counting;
- unchanged payload and successful return;
- sync/async parity;
- sync/async warnings-as-errors perform zero POSTs;
- actual Report export → documented helper reload → submission remains partial;
- real IPython assignment emits one captured HTML display event;
- final expression has one notice plus the unchanged score card, without duplicate copy;
- full notebook submission emits no display;
- failed notebook POST emits no success display;
- async notebook display;
- broken notebook display falls back without losing the saved score id.

Separate tests cover Colab/Databricks inheritance, notice normalization/palette/accessibility,
existing Report amber tokens, and the selected-vs-complete serialization contract.

### Complete verification

- Focused affected suites: **132 passed**.
- Clean temporary merge with current `origin/main`: no conflicts.
- Current merged tree: **1,006 tests collected; 1,005 passed, 1 skipped**.
- Official `screamingface` stack: Ruff check, Ruff format, Pyright, pytest with the 95% coverage
  floor, deterministic notebook check, wheel/sdist build, and distribution validation —
  **ALL GATES GREEN**.

The merged-tree append-only precheck was skipped because current `main` contains PR #607's
already-landed, owner-approved `test_runtime_cli.py` edit relative to this older PR head. The final
OME-922 diff restores `test_leaderboards.py` exactly to `main`, adds one focused test module, and
only appends tests elsewhere.

The four locally executed example notebooks remain user-owned working-tree changes. They were not
regenerated, staged, overwritten, or included in this PR.

## Deliberately left out

- **Server-side full-run ranking enforcement.** The Scoreboard still accepts and may rank partial
  submissions; changing that policy belongs in `apps/scoreboard`.
- **Persisting advisory state on Scoreboard rows.** `get_score()` reflects the server record and
  cannot reconstruct Client-side submission context without a server contract.
- **Migrating unrelated Client advisories.** `ClientNotice` is reusable, but this PR displays only
  the partial-submission advisory.
- **Renaming the progress module's inherited private notebook alias.** It remains separate from
  the stricter submission detector and can be cleaned up without changing this feature.

## Cross-service contract

None. No Scoreboard endpoint, request schema, response schema, authentication behavior, or ranking
calculation changes. The only serialized adjustment is inside the Client-owned Report artifact:
Candidate Results preserve their complete Benchmark Case count for truthful reloads.

## Second review pass (2026-08-21)

Filip's review is fully addressed by the commits above. A follow-up read found four residual
defects, fixed in `8251b221`:

1. **The notebook path bypassed the warnings machinery** — neither escalatable nor suppressible
   there, so the warnings-as-errors guarantee was headless-only. The warning is now emitted in both
   branches; in a notebook it goes inside `warnings.catch_warnings(record=True)`, which applies the
   caller's filters (an `error` policy raises straight through it, an `ignore` policy records
   nothing) while swallowing the stderr write that IPython would render as a red block.
2. **The advisory asserted publication before the POST had happened.** The clause is cut rather than
   qualified, so the copy is true on both sides of the write.
3. **Completeness trusted a rounded wire metric.** `coverage` arrives as
   `round(gradeable / case_count, 4)`, so ≥20 000 cases missing a few grades read as complete. The
   retained `CaseResult`s are now counted directly.
4. **`report.v1` is replaced in place** — the candidate block's `case_count` changes meaning with
   no version bump and no dual-shape loader, which would have been a legacy fallback. The notebook
   display guard is unchanged and still returns the persisted score id if display fails.

Also reverted a behaviour-neutral `score = _score_value(...)` hoist that was diff noise.

Mutation-verified: forcing `prepare_submission_notice` to return `None` fails **13 of 17** tests in
the module, and all four survivors are legitimately no-op assertions.

**Still open for owner-verify:** the notice's dark palette is reached only through
`prefers-color-scheme` and the JupyterLab/VS Code hooks. Colab sets none of them and its theme is
independent of the OS preference, so a dark-theme Colab user on a light OS may get the light box.
Not verifiable without a real Colab session.

**Follow-ups, deliberately not in this PR:** `_STYLE` is re-emitted with every notice (~1.5 KB into
each saved output cell), and there are now two warning presentation systems (`sf-report__warn` amber
vs `sf-notice` persimmon, each carrying its own theme-detection CSS). Keeping them separate is what
keeps review item 6 fixed.

Fixes OME-922

Refs: OME-922


FilipBolt (changes requested):
Requesting changes on 1–4 and the test group. Everything under "smaller things" is fine as follow-up.

Design's right, these all come down to one thing: the feature works on the path the tests use, not the path the docs teach.


## Blocking

### 1 · The warning points at the wrong line, and only shows once

`_scoreboard/leaderboards.py:403` — `stacklevel=3` is counted for `client.leaderboards.submit()`, but the documented `sf.leaderboards.submit()` goes through one extra function, so the count is off by one.

I ran it: the warning is blamed on `src/screamingface/leaderboards.py:26` instead of the user's own line. Worse, Python only shows a warning once per source line, so **three partial submissions from three different lines in the user's code produced one warning between them, not three.**

`README.md:422` and all five example notebooks use `sf.leaderboards.submit()`. The new test only calls the other form, which is why it passes while the documented way is broken.

### 2 · In a notebook, the note only appears if the score is the last thing in the cell

`_scoreboard/leaderboards.py:401` — the Python warning is turned off inside notebooks, so the note travels only on the HTML card. If the cell doesn't end by displaying the score on its own, the user sees nothing.

I ran it: `score = submit(partial)` gives no warning and no card.

Every documented example does exactly that. `examples/08_healthbench.ipynb:236` and `09_corrective_loops.ipynb:208` collect submissions into a **list**, which has no HTML display at all; `00_quickstart.ipynb:194`, `06_draco.ipynb:229`, `README.md:422` and `public-docs/.../learn/LeaderboardPage.vue:19` all assign to a variable. It's also lost when notebooks run under papermill or nbconvert.

### 3 · Half the check stops working on a saved run

`_scoreboard/leaderboards.py:411` — when a result is saved with `to_dict()`, `case_count` is written as the size of *that run*, not the size of the benchmark. So on reload the two numbers always match and the check can never fire.

I ran it: a live 2-of-3 run is correctly flagged partial. Save it and load it back, and it reports as a full run.

That breaks `examples/helpers.py:146`, which is the documented way to publish a run you saved earlier — exactly what someone does with yesterday's expensive run. Live runs are fine (`_evaluation/benchmark.py:60` sets the real size), so this is only about saved and replayed results.

### 4 · With warnings-as-errors, the user loses a score that was already saved

`_scoreboard/leaderboards.py:403` — the note is raised *after* the POST has succeeded.

I ran it with `simplefilter('error')`: the POST went through and the score was stored on the server, then `submit()` raised, so the return value — including the score id — was gone. There's no way to get it back.

`-W error`, `PYTHONWARNINGS=error` and `filterwarnings=error` are common CI settings, and **this PR's own test suite uses that setting in six tests.**

---

## High

### 5 · Colab gets the red block this PR was meant to remove

`_environment.py:18` — the check looks at the shell class's own module name instead of also checking what it inherits from, so Colab (`google.colab._shell`) and Databricks are treated as plain terminals and get the big red warning.

`README.md:241` lists Colab as supported. And the other notebook check in the codebase *does* recognise Colab, so `progress.py` draws its rich panel in the same session where this decides it isn't a notebook.

### 6 · The colour change makes warnings and errors much harder to tell apart

`_ui/style.py:14` — moving from amber to persimmon affects **existing** Report surfaces, not just the new note. Measured: the hue gap between warning and error drops from 43° to 16° in light mode and 41° to 14° in dark. On the badge fill, the colour difference drops from 9.6 to 3.8 for normal vision, and **from 5.5 to 2.3 for someone with red-green colour blindness.**

`report_view.py:104` states the rule outright: warning must never look like the error red. Text contrast actually stayed fine (5.67→5.75), so no contrast check catches this, and no test covers the three surfaces that changed.

This is also the part of the PR furthest from the ticket. It's described as owner-approved — worth saying who approved it, since it changes what users see on surfaces this PR otherwise doesn't touch.

---

## The tests don't currently catch a broken feature

This is the group I'd fix regardless of the rest.

I checked by breaking the feature on purpose — making the function that attaches the note return the score untouched. **7 of the 11 new tests failed, and 4 still passed.** Three of those four are meant to pass (they test that nothing happens on a full run). The fourth is not:

- **`tests/test_leaderboards.py:1464`** — `test_notebook_submission_notice_is_value_neutral_and_stays_off_the_wire` can't fail. It compares two fields that are explicitly excluded from comparison, and searches a request body that this feature never writes to. It never actually checks that a note was attached. **The one test named for proving the note exists cannot tell when it's missing.**
- **`:1393`** — eight colour checks pass on a card with no note in it at all, because those colours are in every card. They also tie the colour change to a behaviour test, so undoing one breaks the other.
- **`:1305`** — the one test checking the exact wording passes the message straight into `pytest.warns(match=…)`, where `.` matches any character, so it's looser than it looks. Six tests set warnings-as-errors individually, overriding the project's narrower setting. And `assert repr(submitted) == repr(plain)` at `:1485` can never fail, because `__repr__` is hand-written and doesn't include notes.

## Smaller things

- **`leaderboard.py:106`** — the field described as temporary is actually a normal stored field: it survives pickling and copying, `replace(score, score=0.9)` carries it onto a score that was never submitted, and with `metadata=None` a partial and a clean score hash the same — so putting both in a set keeps the clean one and the note disappears.
- **`leaderboards.py:89`** — fetching the same score again with `get_score()` returns it with no note, and headless callers get no note on the returned object either. So `-W ignore` removes the only signal completely, and there's no way to ask in code whether a submission was partial.
- **`leaderboards.py:403`** — uses a plain `UserWarning` rather than `sf.EvaluationWarning`. Our docs say otherwise in three places, including the `simplefilter("error", sf.EvaluationWarning)` recipe at `ErrorsPage.vue:30` and the claim at `:202` that it "is the one the Client emits."
- **`_ui/report_view.py:46`** — four new CSS rules duplicate `.sf-report__warn` at `:158`, which uses the same three colours and **already shows this same advisory** at `:429`, for the same condition. Two looks, two places to maintain.
- **`_ui/score_view.py:76`** — every severity gets the warning styling and `role='status'`, so an `info` note renders in full persimmon. `role='status'` only announces text that arrives *after* the element is on the page, so this one is never read aloud; the rest of the repo uses `role='alert'` for static messages. `data-notice-severity` matches no CSS and sits on a hidden span, so the severity is carried by colour alone.
- **`_ui/style.py:14`** — only one of the five copies of the design tokens was updated. The old amber still ships in the scoreboard portal, aigateway-ui, public-docs and the design system's own reference file, so the SDK and the portal now show the same kind of warning in two different colours.
- **`_evaluation/progress.py:11`** — `_in_notebook` here means "is ipykernel loaded", while in `leaderboards.py` it means the stricter "are we really in a notebook". Same name, two meanings, which is the confusion `_environment.py` was created to end.
- **`_notices.py:27`** — `ClientNotice` checks for blank values but doesn't trim them, unlike the four similar classes around it. So `code="partial_submission\n"` is accepted and the newline ends up in the HTML attribute, breaking the CSS selector that the "stable identity" claim depends on.

FilipBolt (approved):
Approving — I re-ran the checks against `8251b221` rather than reading the diff.

All four blockers are fixed:

- **The warning points at the caller's own line**, and three submissions from three different lines now give three warnings instead of one.
- **It's an `EvaluationWarning`**, not a bare `UserWarning`.
- **Partial survives a save and reload** — the exported candidate keeps `case_count: 100` next to the report's `2`.
- **`-W error` stops before the POST**, so nothing is stored and no score id is lost.

Two things you caught that I hadn't flagged:

The notebook path used to skip the caller's warning setting completely, so `-W error` didn't stop and `-W ignore` didn't silence. Both now behave the same in a notebook as headless — I checked all three settings. "A notebook is not an escape hatch" is the right invariant.

And the rounding on `coverage` really does hide missing grades. At 10,000 cases one ungraded case gives 0.9999, but the same single missing grade at 100,000 cases rounds to a clean 1.0 and would have read as a full run. Reading the grades directly is the correct fix.

Two notes, neither blocking:

- I didn't run the notebook rendering itself (it needs a real IPython display), so I'm trusting `_run_notebook_cell` on that one.
- `catch_warnings` changes a process-wide setting, as your own note says. It holds one `warn` call with no I/O, so I think it's fine — worth another look only if submissions ever run from more than one thread.

PR #691, opened 2026-08-23
fix(runtime): make embedded shutdown cancellable

itstauq:
## Summary

- make the embedded runtime's external shutdown waiter cancellable
- construct the Uvicorn-backed server type lazily so SDK-only imports remain lightweight
- include `python-multipart` in the runtime extra, as required by AI Gateway routes
- cover cancellation with a subprocess regression test

## Why

Studio runs the shared runtime CLI in foreground mode and controls it through the desktop lifecycle. Cancelling the previous `asyncio.to_thread(shutdown_event.wait)` task did not stop its worker thread, which could leave the frozen sidecar waiting during shutdown.

The runtime server module is also imported by SDK-only tests and tooling. Creating the embedded Uvicorn server class at import time incorrectly required the optional `uvicorn` dependency. The class is now created and cached only when a local service is built.

Finally, clean wheel and PyInstaller verification exposed that AI Gateway imports form routes requiring `python-multipart`. Adding it to `[runtime]` makes the installable runtime complete rather than relying on an incidental development dependency.

## Suggested review order

1. `packages/screamingface/src/screamingface/_runtime/server.py` — cancellable shutdown waiting and lazy server construction.
2. `packages/screamingface/tests/test_runtime_cli.py` — the subprocess cancellation regression test.
3. `packages/screamingface/pyproject.toml` and `uv.lock` — the missing Gateway runtime dependency.

## Verification

Tested in clean isolated environments both without and with the runtime extra:

```sh
uv run --isolated --frozen pytest tests/test_runtime_cli.py -q
uv run --isolated --frozen --extra runtime pytest tests/test_runtime_cli.py -q
```

- base SDK: 27 passed
- SDK with runtime extra: 27 passed
- clean runtime import: `import multipart; import aigateway.main` passed
- GitHub Actions: Python 3.12 and 3.13 passed
- CodeQL passed

This is a prerequisite for the Studio sidecar integration in #593.


keelancj (commented):
Looks good but just couple things before merge

- The test passes on the buggy code, ran it a bunch times and got CancelledError every time because the worker hasn’t reached event.wait() before cancellation. Sleep ~0.1s first and assert the task is cancelled or no thread leaks

- polling is_set() can miss a quick set/clear. Looping over await asyncio.to_thread(event.wait, 0.05) preserves the edge while remaining cancellable

- Minor/follow-up: scoreboard.wait in server.py:251 uses the same pattern but the subprocess cleanup currently makes it safe

itstauq:
Good point about ensuring the waiter has actually started before cancellation. I’ve updated the regression test to run in a subprocess, synchronize on waiter startup, and use a timeout. This catches the original blocked executor thread without risking hanging the pytest process.

I don’t think the `set()`/`clear()` pulse applies to this shutdown event: it is created as a one-way shutdown latch, only `set()` is called, and it is never cleared. The `is_set()` polling therefore cannot miss the shutdown signal in this code path. Repeated timed `event.wait()` calls would also leave small gaps between waits, so they would not provide a strict edge guarantee in the general case.

keelancj (approved):

PR #692, opened 2026-08-23
feat(screamingface-engine): publish live evaluation progress

keelancj:
## Summary

- publish complete `screamingface.evaluation-progress.v1` snapshots from terminal Benchmark cases
- give each built-in Benchmark one Evaluation adapter that reuses its authoritative grader and scorer
- discover benchmark identity and selected count from the independently executable canonical URL4
- keep terminal facts operation-driven, fail-open, privacy-bounded, and isolated from generic Log ownership
- lazily load only selected per-case grading assets instead of repeating full-corpus validation
- leave generated URL4, `packages/url4`, model/Judge calls, cache identity, final results, and generic executor behavior unchanged

## Review hardening

- identified Case terminals reconcile anonymous Candidate-failure placeholders instead of being frozen out
- projection failures remain terminal failed Cases and cannot stall completion
- discovery covers Iteration body, intent, and reducer templates through one shared registry traversal
- the shared Case endpoint sends a typed terminal observation without changing its authoritative return

## Dependency

Depends on the generic run-log seam in #689. Retarget to `main` after #689 merges.

## Test plan

- `uv run .claude/scripts/run_gates.py screamingface-engine --base 014386e79472d63a3ab21489e4dbf25e1d6a97a4` — all gates green
- focused progress and Evaluation-adapter suite — 57 passed
- broader Benchmark protocol, aggregation, failure-policy, identity, and paid-call regression suite — 141 passed
- verified no changes to `packages/url4` or `runner/executor.py`

Asana: N/A

Refs: OME-932


keelancj:
Superseded after design review. This draft was stacked on the closed, unmerged PR #689 rather than current origin/main, and the adapter needs redesign before implementation. OME-932 remains open. We will preserve the useful terminal-once counting, cumulative snapshots, selected-order scoring, privacy, and fail-open tests, while first locking a refreshed OME-934 generic seam in a documentation-only PR. The replacement must avoid dataset I/O during run-scope opening, consume typed Case observations directly, and decline ambiguous Benchmark discovery.

PR #697, opened 2026-08-24
feat(aigateway): admin cache-snapshot upload (OME-952)

IonesioJunior:
## What

Gateway half of the OME-951 spec (approved):

- **Routes** `GET /v1/admin/cache/info`, `POST /v1/admin/cache/snapshots` (multipart, 202), `GET /v1/admin/cache/snapshots/jobs[/{id}]` — all behind `CurrentAdmin`, audited via `AdminAuditRoute`.
- **Job runner** (`app.state.cache_upload_runner`): single in-flight slot (second upload → 409), owned task, bounded history, states `validating → loading → merging → complete | failed | refused` with counters, warnings, and machine refusal codes.
- **Loader** (`core/request_cache/bulk_loader.py`): staging table + asyncpg COPY (bytes verbatim — uploaded SQL is never executed), then one-transaction **merge** (`ON CONFLICT (key_hash) DO UPDATE`, content from snapshot, local `id`/`created_at`/`hit_count`/`last_hit_at` survive, fresh `gen_random_uuid()` ids — the live table has no DB-side defaults) or guarded **replace** (refuses when live > staged without `acknowledge_loss`).
- **Revision guard**: manifest sha256 + row-count + revision constants vs the live gateway (`core/request_cache/revisions.py` registry; the OpenRouter plugin registers its adapter constant at load — hexagonal direction preserved). Mismatch → refused; `force` override recorded on job + audit line.
- **Schemas**: `AdminCacheInfoOut`/`AdminCacheJobOut`/`AdminCacheJobList` in `admin_schemas.py` for the typed console client.
- Config: `AIGW_CACHE_UPLOAD_MAX_BYTES` (default 256 MiB). Dep: `python-multipart`.

## Notable defects the integration run caught (both fixed + regression-tested)

1. The slicer leaked the dump **epilogue** into COPY across batch boundaries — pg_dump 18's epilogue (blank line → `uuid \\` error) — fixed with a closed-latch on the terminator.
2. Merge INSERT column count mismatch — live table has **no DB-side defaults** for `id` (verified against the running cluster), so ids are generated in SQL.

## Gates

`ruff` clean · `pyright` clean · `check_no_enterprise` OK · `pytest` **3755 passed / 54 skipped, cov 92.28%** · `AIGW_TEST_PG=1` integration **5/5** (testcontainers Postgres 16, genuine `pg_dump` fixtures)

Depends on spec PR #696 (docs only). OME-953 (console UI) stacks on this branch.

Refs: OME-952

PR #699, opened 2026-08-24
feat(repo): snapshot-cache emits a revision-guard manifest (OME-954)

IonesioJunior:
## What

- `snapshot-cache` now writes `<name>.manifest.json` beside the archive: schema tag, `generated_at`, `row_count`, `sha256` of the archive, and the **deployed gateway's** cache-key revision constants (probed from the cluster's aigateway image via `kubectl run --command`, never the host checkout — the checkout may be older or newer than what runs).
- The admin upload (PR #697) verifies this sidecar: checksum, row count, revision constants; mismatch → refused, `force` override recorded. Without a manifest the load proceeds with the console's `revisions_unverified` warning.
- Snapshot artifacts (`*.sql.gz`, `*.manifest.json`) gitignored.
- **This is the script's first tracked commit** — it was previously an untracked local file. ⚠️ The hardcoded OpenRouter API key in the untracked copy is **removed** here (env-only, per the script's own SECURITY NOTE). **Rotate that key** — it existed in plaintext on disk.

## Verification

Live run against the kind stack: 204,765-row snapshot → manifest with matching revisions (`aigw-parameter-contract-2026-08b`, `openrouter-global-cache-2026-08d`) and a sha256 that matches the archive. `bash -n` clean.

Refs: OME-954

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ release/
 # excluded only via .git/info/exclude, which is local to a clone and never travels, so a
 # fresh clone saw them as untracked.
 .claude/worktrees/
+*.sql.gz
+*.manifest.json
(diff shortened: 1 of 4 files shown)


IonesioJunior:
Superseded by #701 (single-PR consolidation per owner decision — local tooling stays untracked).

PR #714, opened 2026-08-24
feat(screamingface-benchmarks): GDPVal text subset

siddhant230:
## Summary

Onboards GDPval — OpenAI's real-work benchmark — as `gdpval-text`, the prose-only slice of its 220-task open gold set: 102 cases graded per rubric criterion by the same judge DRACO pins, plus the SDK CLI command and an example notebook to drive it.

GDPval-text at a glance: 220 open gold tasks filtered to 102 prose cases, reference documents parsed once at build time, one judge call per rubric criterion (median 44 per task, roughly 4,500 per full run), case score = earned over positive points with negatives subtracting and no clamp, board score = plain mean over 102 cases - deliberately not the official pairwise-vs-human metric

Detailed info about the 102 cases we pick 
Screenshot 2026-08-27 at 09 38 37

The board is deliberately **not** GDPval's published metric, and its catalogue description says so: the official metric is blinded expert *pairwise* comparison against a human professional's deliverable, which is unreachable here. The valid claim is fusion versus solo on this board — not parity with the GDPval leaderboard.

**Asana:** _n/a — technical work; tracked in Linear as [OME-971](https://linear.app/openmined/issue/OME-971/onboard-the-gdpval-text-subset-as-a-rubric-graded-benchmark)_

## Components touched

- `apps/screamingface-engine` — new `benchmarks/gdpval/` package (14 modules), registered in `BUILTIN_DEPLOYMENT`
- `packages/screamingface` — `prepare gdpval` in the CLI, `examples/10_gdpval.ipynb`, GDPval's document parsers pinned into the `runtime` extra

Cross-cutting per CLAUDE.md §8; folded into one ticket by owner decision rather than split into an epic.

## Test plan

- [x] Ran both stacks' full gates via `run_gates.py` — `screamingface-engine` and `screamingface`, all green
- [x] `pytest` — 2,131 pass on the engine; 100 new GDPval tests across 8 files
- [x] Preparer run end to end against the live dataset: `gdpval: baked 102 cases … (7 tasks excluded for unusable references)`
- [x] `screamingface prepare gdpval` through the CLI, writing a manifest
- [x] Single live judge call against `:9105` to verify reply parsing
- [x] Rebased on `origin/main`, re-ran both gate suites
- [ ] **A full graded run producing a score — not yet done** (see below)
- [ ] Gateway request/refresh change → n/a
- [ ] Screenshots / recording — n/a

## Cross-service notes

No other owner's paths are touched, but four things reviewers should weigh.

**1. No score has been produced yet.** Everything is verified against the dataset, the test suite and one live judge call, but no candidate has been graded end to end. A five-task pilot costs ~$3 and is the honest gate before merge.

**2. Four exhaustive enumerations were extended**, plus one prior test changed deliberately. `test_benchmark_protocol`, `test_benchmark_display_metadata` (×2), `test_benchmark_deployment` and the new family guard all grew to include the sixth board; none were relaxed. Separately, the verdict audit key moved from `raw` to the contract's `raw_output` — without it a rejected judge reply is recorded as invalid with nothing to inspect.

**3. Two defects that only integration caught.** `prepare` assigned `rubric_id` from a 0-based enumerate while `case_score` indexes from 1, so every criterion would have been scored against the wrong point value, silently. And the dataset pin copied from `inspect_evals` predates the rubrics by five months (`rubric_json: null` on all 220 rows) — their implementation never reads the rubrics; this board is built entirely on them.

**4. Cost shape is unusual.** Grading fans out one judge call per criterion — ~4,498 per candidate on a full run, roughly $51 at the pinned judge's rates. Fine, but worth knowing before anyone schedules a sweep.

`CHECK_THRESHOLD = 0.5` for the corrective-loop check surface is marked provisional in the code: DRACO's 0.7 was set against known baselines, and nothing has been measured on this board yet.

## Review order

Sitting 1 — contract (what "gdpval-text" even is):
1. docs/spec/…OME-971….md then docs/plan/… — the intent you'll judge everything against
2. gdpval/pins.py — pinned dataset + parser revisions; everything downstream inherits these
3. gdpval/subset.py — which 102 of 220 tasks are "text" and why; wrong filter = wrong benchmark
4. gdpval/definition.py + builtins.py — how the benchmark registers (id, revision, routes)

Sitting 2 — build-time pipeline (answer-key correctness):
5. gdpval/prepare.py → ingestion.py → records.py — download, parse reference docs (pdfplumber/python-docx), bake case records. Bugs here silently corrupt what every candidate is asked.
6. gdpval/rubric_filter.py — which rubric criteria survive; median-44 claim comes from here
7. Dockerfile.benchmark — parsers present in the image that runs prepare

Sitting 3 — run-time pipeline (scoring correctness):
8. gdpval/exam.py + prompts.py — what the candidate sees
9. gdpval/case_evaluation.py → verdict.py → check_policy.py → scoring.py → aggregate.py — one judge call per criterion → parse → policy → per-case score → board number. Check the math against the spec's worked example.
10. gdpval/runtime.py — asset loading, strict decode
11. SDK side: _runtime/cli.py (the prepare gdpval command), 10_gdpval.ipynb — does the story a user reads match what the code does

Tests last per area, not in one lump — read each test_gdpval_*.py right after its module and ask "does this test defend the invariant or just the implementation?"


packages/screamingface/scripts/build_notebooks.py
```diff
@@ -1015,6 +1016,130 @@ def _healthbench_e2e() -> NotebookNode:
...
+
+A case score is points earned over points winnable. Penalties subtract without widening the
+denominator, so a case can score **below zero** — that is intended, not a bug: a harmful answer
+should rank below one that said nothing.
+
+A case the judge could not fully grade reports `None` rather than `0.0`. The two are different
+facts, and collapsing them would make a judge outage look like model weakness."""),
+        nbformat.v4.new_code_cell("""\
+for case in fusion_report.candidates.only.cases:
+    print(case.id, case.status, case.grade.score if case.grade else None)"""),
```
khoaguin: **Confirmed bug (user-facing):** this generated cell iterates `for case in ...cases:` and reads `case.id`, but `CaseResult` is a frozen slots dataclass whose field is `case_id` (`packages/screamingface/src/screamingface/case_result.py:255-260`). The notebook's closing per-case-scores cell always raises `AttributeError` — after the user has already paid for the solo + fusion runs. The notebook gate only checks deterministic regeneration, never execution, so CI can't catch it. Fix: `case.case_id` here, then regenerate `10_gdpval.ipynb`.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/case_evaluation.py
```diff
@@ -0,0 +1,168 @@
...
+        raise ValueError(f"Case {selected} carries no Case record")
+    return {
+        "schema": CASE_EVALUATION_SCHEMA,
+        "case_id": selected,
+        "case": case_record,
+        "rubric_evaluations": rows,
+    }
+
+
+def decode_case_evaluation(value: object, expected_case_id: int) -> dict[str, Any]:
```
khoaguin, line 99: **Confirmed silent-scoring bug:** `decode_case_evaluation` dropped healthbench's stricter decode checks (duplicate-`rubric_id` set + evidence/evaluation `rubric_id` cross-consistency — healthbench raises for both at its `case_evaluation.py:147-171`). Aggregate keys `_verdicts` on `evidence.rubric_id` (aggregate.py:577) but `_checks` on `evaluation.rubric_id` (aggregate.py:370), so a schema-valid-but-inconsistent envelope scores quietly wrong instead of failing visibly.

Verified example: `points=[5,3]`, rows `{rubric_id:1, evidence:1, met:true}` and `{rubric_id:1, evidence:2, met:false}` → `_checks` dedups to one UNMET check (pass_rate 0.0) while `_verdicts` yields `{1:true, 2:false}` → case_score 5/8 = 0.625. The write-time binders that would enforce agreement are never called on the aggregate read path (aggregate.py:523 uses decode only). Fix: restore the two consistency checks in decode.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
```diff
@@ -0,0 +1,314 @@
...
+        return extract_reference_text(task_id, file_name, reader=built[suffix])
+
+    return read
+
+
+def _fetch(task_id: str, file_name: str, urls: Mapping[str, str], cache: Path) -> None:
+    """Download one reference into the cache, unless it is already there."""
+
+    destination = cache / file_name
+    if destination.is_file() and destination.stat().st_size > 0:
```
khoaguin, line 253: **Confirmed cache-poisoning path:** the cache-hit check is existence + nonzero size, and the download writes straight to the final path (`destination.write_bytes(...)` at line 266, non-atomic). If the process dies mid-write for a multi-MB PDF (Ctrl-C, SIGKILL, disk-full OSError swallowed by the retry loop at 267-270), the truncated non-empty file is treated as a valid cache hit on every subsequent run. Then either extraction fails with guidance that steers toward excluding a healthy task — or, worse, if the truncated PDF still parses past `MIN_VIABLE_CHARS`, truncated reference text is silently baked into the answer key. Fix: write to a `.part` temp file, then `os.replace`.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
```diff
@@ -0,0 +1,314 @@
...
+        try:
+            with urllib.request.urlopen(url, timeout=_FETCH_TIMEOUT_S) as response:
+                destination.write_bytes(response.read())
+        except OSError as exc:
+            last = exc
+            if attempt + 1 < _FETCH_ATTEMPTS:
+                time.sleep(_FETCH_BACKOFF_S * (attempt + 1))
+        else:
+            return
+    raise IngestionError(
```
khoaguin, line 283: **Confirmed error-reporting gap:** `IngestionError` (raised by `_fetch` on exhausted retries and by `extract_reference_text`) escapes `prepare()` (no try/except at 279-293), but the Docker-build orchestrator catches only `BenchmarkAssetPreparationError` (`benchmarks/prepare.py:55-59`). One flaky HF CDN download across the 85 reference files → the CI image build dies with a raw traceback instead of the clean OME-925 operator line. The standalone `main()` here already catches `(PrepareError, IngestionError)` (line 303) — only the Docker path is unguarded. Fix: wrap or translate `IngestionError` into `BenchmarkAssetPreparationError` at the `prepare()` boundary.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
```diff
@@ -0,0 +1,314 @@
...
+        f"task {task_id}: reference {file_name!r} could not be downloaded after "
+        f"{_FETCH_ATTEMPTS} attempts: {last}"
+    ) from last
+
+
+def prepare(out: Path, *, assets_root: Path | None = None) -> dict[str, Any]:
+    """Bake the GDPval text-subset assets into ``out``, returning its audit summary."""
+
+    rows = load_rows()
+    cache = assets_root or (out / "_references")
```
khoaguin: **Confirmed image bloat:** the default download cache `out / "_references"` lives inside the served asset dir. `deployment.py:118` calls `prepare(out=/opt/benchmarks/gdpval)` → cache lands at `/opt/benchmarks/gdpval/_references`, and `Dockerfile.benchmark`'s runtime stage copies `/opt/benchmarks` wholesale — so all 85 original PDF/DOCX binaries (tens of MB) ship in every Runner image and local SDK assets dir, despite never being read at runtime (references are already flattened into `cases.json`). This contradicts the PR's own `cli.py` comment defining the downloads as "a build cache rather than a served asset". Fix: default the cache outside `out`, or drop `_references` before the runtime copy.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
```diff
@@ -0,0 +1,314 @@
...
+    rootfs — never needs the network or the original binaries.
+
+    AIDEV-NOTE: the concrete readers are built LAZILY, on first use of each extension. 66 of the
+    102 selected tasks carry no reference files at all, and eagerly constructing them would make
+    `pdfplumber` and `python-docx` hard requirements of a build that may never open a PDF — and
+    would make this function untestable without both installed.
+    """
+
+    cache.mkdir(parents=True, exist_ok=True)
+    factories = {".pdf": pdf_reader, ".docx": docx_reader, ".doc": docx_reader}
```
khoaguin, line 230: **Plausible (dormant) reader-map bug:** `.doc` routes to python-docx, which parses only OOXML zips — a legacy binary OLE `.doc` would raise a misleading corrupt-file error. And `.txt`/`.md`, declared allowed prose formats in `subset.py`, have no reader at all (`IngestionError: no reader for '.txt'`). Dormant today because all 85 pinned references are PDF/DOCX, but the subset filter admits all five extensions, so a dataset re-pin can trip it. Cheap fix now: read `.txt`/`.md` as plain text and either drop `.doc` from the admitted set or fail it with an honest "legacy .doc unsupported" message.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/aggregate.py
```diff
@@ -0,0 +1,590 @@
...
+    for case_id in case_ids:
+        row = by_id.get(case_id)
+        input_value = row.get("input") if isinstance(row, Mapping) else None
+        if not isinstance(input_value, str) or not input_value.strip():
+            raise AggregateError(f"GDPval Case {case_id} has no public input")
+        selected.append(SelectedCase(case_id=case_id, input=input_value, metadata={}))
+    return selected
+
+
+def aggregate(
```
khoaguin, line 113: **Design conversation, not a quick patch:** this 590-line reducer is a function-for-function clone of `healthbench/aggregate.py` (~23 private helpers share names, signatures, and logic; the only real delta is mean vs clipped_mean). Unlike `scoring.py`/`exam.py`, it carries no documented WHY-not-shared defense. The failure-visibility ladder now lives in two copies that already diverge structurally — a fix in one board silently misses the other, and the next rubric benchmark clones it a third time. Suggest hoisting the reducer skeleton into the shared `benchmarks/aggregation.py`, parameterized by scorer, row decoder, and failure vocabulary (none of which is revision-frozen). Fine as a follow-up ticket if you'd rather not grow this PR.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/verdict.py
```diff
@@ -0,0 +1,192 @@
...
+        return None
+    text = _without_fences(raw.strip())
+    try:
+        decoded = json.loads(text)
+    except json.JSONDecodeError:
+        decoded = _first_json_value(text)
+    return decoded if isinstance(decoded, dict) else None
+
+
+def _without_fences(text: str) -> str:
```
khoaguin, line 148: **Duplication with a proven drift cost:** `_without_fences`/`_first_json_value` are the third byte-identical copy of the judge-reply JSON recovery (`healthbench/verdict.py:169-186`, `draco/verdict.py:138-155`, plus a fourth variant in `rubric_check.py:444-450`). The work ledger for this PR records exactly this drift biting once already: gdpval initially mirrored HealthBench's fence-less parser while pinning DRACO's judge, and every Case failed at rubric 1 until the fence-strip was re-copied in. The next judge formatting quirk gets fixed in one module while the other boards burn retries (~4,498 judge calls per gdpval run). Decode leniency isn't hashed into any board's revision, so a shared helper is revision-safe.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/runtime.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+def _rubric_tasks(root: Path, case_ids: tuple[int, ...]):
+    """The fan-out point: one Candidate submission in, N ready-to-send judge tasks out."""
+
+    def rubric_tasks(request: Request) -> str:
+        try:
+            case_id = positive_case_id(request.intent)
+            answer = candidate_answer(request.context)
+            raw_cases = _read(root / "cases.json", "GDPval cases")
```
khoaguin: **Confirmed avoidable I/O:** the handlers re-read and re-parse the multi-MB `cases.json` (which embeds the flattened text of all 85 references) with zero caching. `/rubric-tasks` does one disk read + two `json.loads` per call (`_request_text` at runtime.py:240 and `bind_case → _decode_cases` at case_records.py:69) — ~204 full parses per 102-case run; `_cases` (runtime.py:119-125) parses in preflight (line 105) then again at lines 122/269, plus ~102 rubric-file opens repeated at `aggregate()`. Cheap fix: have preflight return the parsed list and feed `_select_cases`; cache the parsed `{case_id: row}` dict in the closure — baked assets are immutable for the process lifetime.

docs/work/2026-08-24-OME-971-gdpval-text-subset.md
```diff
@@ -0,0 +1,381 @@
...
+preparer's four-attempt retry with backoff is what stands between that and a red build. If image
+build time becomes a problem, the fix is caching the references as a build layer, not dropping
+the retry.
+
+**Caught only by CI.** Every local path — `uv run`, the SDK CLI, the direct module invocation —
+had the parsers available one way or another. The image build is the one environment that
+installs its preparer dependencies explicitly, so it was the only place the omission could
+surface.
+
+## Outcome (fill at the end — required before COMMIT)
```
khoaguin, line 408: **Process nit:** the Outcome section is committed as raw template placeholders (`<vs planned>`, `<sha — message>`, `<run_gates.py result line / counts>`) while every task above is marked DONE and the PR is up. Repo CLAUDE.md (AI SDLC rule 2) requires the outcome filled at finish — please fill before merge.

khoaguin (commented):
Automated multi-angle review (8 finder passes, adversarially verified — 9 confirmed, 1 plausible). Overall: the wiring is sound — the full handler chain (cases → rubric-tasks → verdict → case-evaluation → aggregate) traces correctly and 332 targeted tests pass on the branch. Inline comments below, roughly ranked.

**Must-fix before merge:** the notebook's final cell crashes (`case.id` vs `case_id`), the lenient `decode_case_evaluation` lets an inconsistent grading envelope score wrong instead of failing, and the reference-download cache can be poisoned by a truncated partial file.

**Worth a conversation, not a quick patch:** the 590-line aggregate clone and the third pasted copy of the judge-reply JSON recovery — the work ledger shows the parser copy already caused one full-run failure during development.

**Checked and cleared:** the `datasets==5.0.0` vs `5.0.1` Dockerfile/pyproject skew pre-exists this PR; scoring.py/exam.py healthbench-mirroring carries documented WHY-not-shared defenses.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/aggregate.py
khoaguin: Will do later

packages/screamingface/scripts/build_notebooks.py
khoaguin: Fixed in b7e288b0 — generator now emits `case.case_id`; the notebook cell was patched surgically (outputs preserved). A new SDK test pins every `case.<attr>` in every example notebook to the real `CaseResult` fields, closing the bug class.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/case_evaluation.py
khoaguin: Fixed in b7e288b0 — `_decode_rubric_evaluations` now enforces unique positive `rubric_id` plus nested rubric/evidence agreement on schema, case_id and rubric_id. The 0.625 exploit envelope now raises. 7 new RED-first tests in test_gdpval_case_evaluation.py.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
khoaguin: Fixed in b7e288b0 — bytes land in a `.part` sibling and only a finished download is renamed into place via `os.replace`; the failure path unlinks the temp. Tests pin that the destination path is never written in place.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
khoaguin: Fixed in b7e288b0 — `prepare()` translates `IngestionError` into `PrepareError` (a `BenchmarkAssetPreparationError`) at the boundary, so the Docker orchestrator reports the clean OME-925 operator line. Test asserts the translated type and message.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/prepare.py
khoaguin: Fixed in b7e288b0 — the default cache moved to a temp-dir path keyed by the dataset revision (`gdpval-references-<rev12>`), outside the served asset tree; `--references`/`assets_root` still override. Runner images stop shipping the 85 binaries.

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/runtime.py
khoaguin: Fixed in b7e288b0 — `preflight` returns the cases.json text it validated, `_cases` and `_rubric_tasks` memoize per closure (raw text, per-case request text, rubric items). Only successes are cached, so broken assets keep failing loudly. Read-once pinned by tests.

docs/work/2026-08-24-OME-971-gdpval-text-subset.md
khoaguin: Filled in b7e288b0 — Outcome carries actual files, the commit range, both gate suites' results, and the deviations (no graded run yet; two consolidations deferred to follow-ups).

apps/screamingface-engine/src/screamingface_engine/benchmarks/gdpval/aggregate.py
siddhant230: do you need help here? Also; thanks for the review and fixes :-D

PR #719, opened 2026-08-25
feat(scoreboard): support private leaderboards

FilipBolt:
Closes OME-894.

HealthBench worst-30 is the public **entry challenge**, so its submissions must not be publicly visible: staff see everyone's, participants see only their own. Built as a general capability — a benchmark is public or private — because the entry challenge will not be the last board that needs it.

Privacy is enforced in the **API**, not the portal. The portal is static JavaScript against a public API, so hiding rows in the page would leave `curl /v1/leaderboard/healthbench-worst30` serving the whole board. Three of the four read paths leaked.

## What a private board does

| Path | Behaviour |
| -- | -- |
| `/v1/benchmarks` | listed, marked `visibility: "private"` — participants must find it to enter, and the catalogue carries no scores |
| `/v1/leaderboard/{id}` | `entries: []` for everyone; the caller's own rows go to `my_submissions`, with `scoped_to_caller: true`. Anonymous gets both empty **with** the flag |
| `/v1/leaderboard/{id}/{spec}/history` | served only for the caller's own spec, otherwise 404 |
| `/v1/leaderboard/{id}/frontier` | 404 — an aggregate over everyone by definition |
| `/v1/scores/{id}` | 404 unless the caller submitted it — score ids are handed out by the submission response |
| `POST /v1/scores` | refused without a verified submitter; per-submitter dedup so two participants sharing a recipe keep separate rows |

**`entries` is the public ranking, and a private board has none** — so it is empty for everyone, including a participant. Own rows are a different concept, which is why `my_submissions` exists and why nothing in it carries a rank. `rank` stays a required `int`.

## Decisions worth review

**A private board 404s a nonexistent spec too.** The ticket asked for 404-not-403 on another participant's spec, since a 403 confirms it exists. A public board answers an unknown spec with `200` and an empty list; keeping that here would have made "someone else's" distinguishable from "no such spec" by status alone. Both return an identical 404 — body and headers included.

**An owner-scoped board skips the registered-revision filter.** That filter stops incomparable numbers being *ranked* against each other, and a private board presents no ranking. Keeping it would hide a participant's own submission with no explanation — the silent disappearance that stranded a real DRACO run on 2026-08-19.

**Baselines stay visible.** Imported LMArena / Artificial Analysis numbers are public third-party data, not participant submissions, and a participant needs the line to beat. The one call the ticket does not cover — say if it should be hidden.

## The submission path, which took most of the review

Making reads owner-scoped was the easy half. Dedup is a read primitive, and it leaked in ways review surfaced one at a time:

- **Writes fail closed.** Under `auth_mode=disabled` the body's `submitted_by` is trusted, so forging an address and submitting a matching recipe returned the victim's stored row. A private board now refuses writes without a verified submitter, *before* the dedup lookup — refusing after it would still confirm whether a matching row exists.
- **A key mapping is honoured only when the caller may read its target.** What the key points *at* decides, never what the request claims. The lookup is global rather than per-benchmark, so a board that turns private leaves reachable mappings behind. A private target is honoured only for its **verified owner** — verified is what makes the ownership comparison mean anything, since an unverified caller can claim any address.
- **`sfp-` and `sfu-` are server-owned key namespaces.** Public keys are stored verbatim, so a client could otherwise occupy the namespace the server generates into. Both prefixes are escaped, and no value the generator produces is storable verbatim — asserted as a property, not a prefix list.
- **A reserved-namespace mapping is honoured only if this scheme wrote it.** `IdempotencyKey.scheme` records the writer. A legacy `sfp-` row carries no scheme, so a value planted before the namespaces were reserved cannot replay as a scoped key however plausible its stored `submitted_by` looks. Deliberately not backfilled — `NULL` *is* the meaningful value.
- **Migrations `0009` and `0010`.** `0009` clears both namespaces of client-written rows; it is a `pre-upgrade` hook, so it runs before any new pod can have written a new-scheme row and can only ever delete legacy ones. `0010` adds the nullable `scheme` column.

## Stale visibility, which took the rest of it

`visibility` is the only authorisation input another process can change mid-request — the seed job flips boards on every deploy. A decision read from it and acted on later is wrong, and this was found at five layers in turn: the write path, then every read path, then a response body carrying a stale copy, then one exit of one branch, then the replay revalidating the wrong benchmark.

- **Writes** read `visibility` once, revalidate before returning a row and again before persisting one, and take `select_for_update()` on the benchmark row inside the insert transaction. A request whose rules changed underneath it is refused as **409** — nothing is wrong with it and a retry gets one consistent view.
- **Reads** re-decide instead of erroring, which a read can and a write cannot: after re-reading, the private response for that path is produced — scoped entries, or the 404s history and frontier already give. Only the branch about to return public data re-reads; private → public mid-read returns the scoped shape, which is less than the caller could have had and never more.
- **The response body** states the verdict too. `scoped_to_caller: true` beside `visibility: "public"` was possible, and `visibility` is the field D4 added for clients to act on.
- **A replay revalidates the returned row's board, not the requested one.** They can differ — the key lookup is global, so a mapping can point at a row on another benchmark. Revalidating the request's board and then returning that row let a private row out through a public request. Both replay paths now re-apply the read gate to the row actually being returned.

**Correcting an earlier version of this description.** It said the persist path took `select_for_update()` on the benchmark row, and that the lock was untestable because SQLite no-ops it. The second half was wrong, and it was hiding the first half being wrong too: the code called `select_for_update()` and then `values_list()`, which builds a fresh query without copying the lock state, so `FOR UPDATE` was silently dropped and **no lock was taken on any dialect**. The claim was made about PostgreSQL and never rendered against PostgreSQL.

It is taken now, and `test_the_persist_path_really_locks_the_row` renders the SQL of *the query production runs* on the asyncpg dialect and asserts `FOR UPDATE` — no server needed. A test that built its own query would have kept passing while this one regressed, which is why `visibility_query()` is a method rather than inline.

The remaining limit, stated accurately: SQLite does not implement the lock, so its *effect* is exercised only on PostgreSQL. Revalidation alone shrinks the window to the commit interval; it does not eliminate it.

**And the class is now asserted, not just its instances.** `tests/unit/guards/test_visibility_exit_guard.py` walks every function deriving a decision from `visibility` and checks, per `return` and `raise`, that a revalidation must have executed first. Exits that legitimately need no guard are listed with the reason. A new unguarded exit fails CI; so does an entry that has since been guarded, so the list shrinks rather than over-permitting. Four regression shapes are proven caught, not assumed.

## Three things this PR reverses

All three were mine, and all three are worth flagging rather than leaving to be spotted in the diff.

**A blanket refusal of private key mappings became an ownership check.** An earlier revision refused *every* mapping pointing at a private row, because `submitted_by` is forgeable under `auth_mode=disabled`. That broke `same key replays the original` on every private board — the contract a key exists to provide, and one the content hash cannot supply since it only catches an identical retry. Coupling the identity decision to the store is what made the comparison sound.

**A post-upgrade purge job was added, then deleted.** It cleared client-written rows from the reserved namespaces during the rollout window. With the ownership rule restored it is not load-bearing for privacy, and it was deleting live mappings — breaking the same contract for public callers whose raw key occupies a reserved prefix. Deleting it removed **279 lines**.

**And the marker column I argued against is in this PR.** That argument said a marker column "would fix that case and not the private one, and would carry a schema change forever to serve one rollout". It was right about the *purge* and wrong as a claim about provenance generally: a `sfp-` row planted before the namespaces were reserved carries a forgeable `submitted_by`, and no rule reading only the row's contents can tell it from one we wrote. `0010` adds the column. The rollout argument does not apply, because this is not for a rollout — the distinction is permanent.

## Staff access

No admin API, per the ticket: nothing to secure, guess, or accidentally expose. `python -m scoreboard.export_private_submissions --benchmark healthbench-worst30` reads the database directly and keeps the submitter's full address, which the public API deliberately trims.

## Test plan

The regression guard was written **first**, before any privacy behaviour existed, because the risk was breaking the public board while securing the private one. Gates green after every commit. **No test that predates this PR is modified** — the only lines differing from `main` in the store suite are imports, so the append-only gate passes unaided.

Beyond the suite:
- Every fix is **mutation-checked** — reverting it re-fails its test. Several fixes shipped without a test that bound to them and were caught this way rather than by reading.
- `store.py` and all four route modules are at **100% statement coverage**, including the concurrent-retry handler that was uncovered when review first flagged it.
- A **real uvicorn run** with `auth_mode: cloudflare_headers` and two seeded participants exercised every path for anonymous / alice / bob.
- The **PostgreSQL-only regressions now run in CI**, against a `postgres:17-alpine` service. They were opt-in behind `SCOREBOARD_TEST_DATABASE_URL`, which CI never sets — so one of them had silently stopped reaching PostgreSQL at all and was green-by-skip everywhere anyone looked. A guard asserts the job still names every module that skips itself without a database.
- `tortoise makemigrations` reports no drift.

## Before this ships

**Confirm the runtime `authMode` with @Stephen.** The chart ships `authMode: disabled` (`values.yaml:41`) and `values-prod.yaml` overrides nothing — verified by rendering it. With auth disabled a private board is inert in both directions: reads fail closed, writes are refused, and *participants would not see their own rows*. Staff still have the export. The deployed value is not verifiable from this repo (OME-730).

Flipping HealthBench to private is config, not code: set `visibility: private` on its `seedBenchmarks` entry and re-run the seed job. Nothing here makes any existing board private.

**Rolling back below this release publishes every private submission.** Privacy is enforced by code reading `Benchmark.visibility`; the database only stores the column. Executed rather than argued: base `454253da`'s `src/scoreboard/` contains no occurrence of `visibility` at all, and run against a database holding one private board with three submissions its `leaderboard()` returned all three, submitter addresses included.

The chart cannot guard this, and I checked rather than assumed — `helm rollback` executes the **target** revision's hooks (`execHook(targetRelease, release.HookPreRollback, …)`), and the pre-privacy revision's stored manifest has none, so a `pre-rollback` hook added here would be theatre. The control is an operator preflight:

```
python -m scoreboard.check_rollback_safety
```

Non-zero while any board is private, listing each with its submission count. Read-only, and no override flag — clearing the refusal by flipping a board public *is* the leak, performed deliberately. The floor is the tool's own running version, not a hardcoded constant that would have been a guess at a release number that did not exist yet. `DEPLOYMENT.md` carries the invariant, the activation ordering rule and the fail-closed procedure, and the generic rollback snippet now points at it.

The one thing the preflight structurally cannot check is whether a *safe rollback target exists* — that is release history, not database state. Tracked as `OME-1027` (High, Week 4), which must close before any board is flipped private.

OME-909 remains the missing half: a participant now reliably sees their own row, including one measured against another revision, but nothing yet tells them why it would not rank.


HupBaHa (changes requested):
Thanks for the thorough implementation and the two internal review rounds. The main privacy filtering looks solid, and the local quality gates are green. I found four remaining blockers before this is safe to merge:

1. **Private submissions are readable through unauthenticated POST dedup**

   In `auth_mode=disabled`, `_resolve_submitter()` trusts the body's `submitted_by`. An unauthenticated caller can claim another participant's email and submit a matching recipe. The dedup path then returns the participant's existing private row, including its stored metadata.

   I reproduced this: the forged POST returned `200` with the victim's original metadata. The same path also allows forged submissions to be stored under another participant's identity.

   Private-board writes need verified identity, or making a board private must be rejected while authentication is disabled.

2. **The scoped idempotency key is incompatible with PostgreSQL**

   `_scoped_idempotency_key()` stores `email + "\x00" + key` in a `VARCHAR(255)` column. PostgreSQL does not permit NUL characters in any character type, so private submissions carrying an `Idempotency-Key` will fail in production. Concatenation can also exceed the column's 255-character limit.

   Please use a bounded, PostgreSQL-safe namespace and cover it with a PostgreSQL-backed test.

3. **Identity-scoped responses do not prevent shared caching**

   Private leaderboard, history, and direct-score responses currently have neither `Cache-Control` nor `Vary`. A shared cache can therefore reuse one participant's response for another caller at the same URL.

   The private success and privacy-preserving error responses should emit an explicit policy such as `Cache-Control: private, no-store`, with appropriate `Vary` handling. Even if the current proxy does not cache these responses, keeping the policy at the privacy boundary avoids turning a future proxy/configuration change into a data leak.

4. **The concurrent retry handler uses the unscoped key**

   In `ScoreStore.submit()`, the `IntegrityError` branch calls `_resolve_existing(idempotency_key, content_hash)` instead of using `stored_key`. A concurrent private retry can therefore bind or resolve the raw global key, undoing the participant isolation added by this PR.

   This branch is currently uncovered, and the PostgreSQL concurrency tests are skipped without `SCOREBOARD_TEST_DATABASE_URL`.

Local verification completed:

- 409 tests passed, 2 PostgreSQL tests skipped
- 86.46% coverage
- Ruff check and formatting passed
- Pyright passed
- 25 portal tests passed

Because these affect the privacy boundary and the production PostgreSQL path, I don't think the PR is merge-ready yet.

FilipBolt:
All four addressed in `fb81fc29`, rebased onto current `main`. Each was reproduced before being touched — thank you, two of them were production-only.

**1. Forged writes on a private board.** Reproduced exactly: the attacker's POST returned the victim's row, `metadata` and `url4_expression`.

The route now refuses writes to a private benchmark whenever `auth_mode` is `disabled`, and the refusal sits **ahead of `store.submit()`** — refusing after the dedup lookup would still tell a forged request whether a matching row exists. Reads already failed closed in this mode (D2), so writes now match and a private board is inert in both directions rather than half-open.

`Benchmark.exists()` is deliberately kept as the existence gate rather than folded into the visibility read: it is the seam `test_post_score_store_unavailable_returns_503` patches, and that prior test is untouched. Cost is one extra indexed lookup on a non-hot write path.

**2. NUL in the scoped key.** Confirmed against a real `postgres:16`:

```
ERROR:  invalid byte sequence for encoding "UTF8": 0x00
```

You were right that this was production-only — SQLite accepts it silently, so every test passed while every private submission carrying an `Idempotency-Key` would have failed once deployed. The stored key is now `sfp-` + sha256: 68 printable ASCII characters, well inside `VARCHAR(255)`, with the separator surviving only inside the digest input.

`tests/unit/scores/test_idempotency_postgres.py` is the PostgreSQL-backed test. It opens its own connection rather than using `tortoise_db`, because that fixture routes through `postgres_schema_database_url`, which calls `asyncio.run()` inside a running loop and errors before any test body executes — OME-430, still live, I confirmed. It is mutation-proven: restoring the NUL separator fails it with the error above.

**3. Shared caching.** There was no `Cache-Control` or `Vary` anywhere in the app. Private board, history, frontier and direct-score responses now carry `Cache-Control: private, no-store` and `Vary: X-User-Email, Origin`. The refusals carry it too — they are identity-dependent as well, and replaying one participant's 404 to another would deny them their own history. A public board is untouched and stays ordinarily cacheable, guarded by a test.

**4. Unscoped key in the retry branch.** Confirmed by recording what `_resolve_existing` was called with: `['bob@example.test\x00shared', 'shared']` — the second is the raw key from the `IntegrityError` handler. It uses `stored_key` now, and the test forces the branch rather than relying on a real race.

**Verified end to end** against a running server: bob replaying alice's idempotency key with the same recipe gets his own row and never sees her metadata; the same private board under `auth_mode=disabled` returns 403; the cache headers are present on the wire.

One thing I found while fixing rather than in review: my first PostgreSQL test called `Tortoise._drop_databases()` in cleanup, which destroyed the database and would have broken a shared CI PostgreSQL. Caught only because a mutation check could not reconnect on the second run. Cleanup is now scoped to the rows the test creates.

No prior test was modified.

**One consequence worth flagging for deployment.** Fix 1 makes the `authMode` question load-bearing rather than advisory: with authentication disabled a private board now refuses writes as well as reads, so the feature is fully inert until a deployment runs `cloudflare_headers` with `allowedNetworks` set. Previously participants merely could not *see* their rows; now they cannot submit either.

HupBaHa (changes requested):
Re-reviewed at `fb81fc29`. The four findings from the prior review are addressed, but the new scoped-key implementation introduces one privacy blocker and its PostgreSQL regression test is destructive against the shared-database mode it claims to support.

1. **[P1] Public raw idempotency keys can collide with private scoped keys** (`apps/scoreboard/src/scoreboard/scores/store.py:258-279`, `:414-420`). Private keys are stored as predictable `sfp-<sha256(email + NUL + key)>`, while public keys are still stored verbatim and may use exactly that shape. `_resolve_existing()` checks the global key before content hash or benchmark. I reproduced this on the PR head: create a victim private submission with key `predictable-key`, then submit to a public benchmark using the computed `sfp-...` value as the raw `Idempotency-Key`; the public submit returned `created=False`, the victim private score UUID, and `{'secret': 'victim-only'}` metadata. The prefix does not make namespaces disjoint unless public keys cannot occupy it. Please make the public/private namespaces actually disjoint and add the public-to-private collision regression.

2. **[P1 test safety] The PostgreSQL test deletes all idempotency rows in the configured database** (`apps/scoreboard/tests/unit/scores/test_idempotency_postgres.py:46-75`). It initializes directly against `SCOREBOARD_TEST_DATABASE_URL`, asserts `IdempotencyKey.all()` contains exactly two rows, and its `finally` executes `IdempotencyKey.all().delete()`. On any reused/shared test database, pre-existing data makes the assertion fail and is then deleted, despite the comment saying cleanup removes only this test's rows. The static `private-pg` benchmark ID can also overwrite or collide with existing data. Please isolate the test in a unique schema/database or use unique rows and narrowly delete only those rows.

The previous four fixes otherwise look correct: private writes fail closed under disabled auth before dedup; the concurrent handler uses `stored_key`; private responses and identity-dependent refusals carry `private, no-store` plus `Vary`; and the stored private key is bounded and PostgreSQL-safe.

Verification on the PR head: focused affected suites 152 passed / 3 skipped; full Scoreboard suite 422 passed / 3 skipped / 3 deselected; Ruff check, Ruff format check, Pyright, and diff check passed. `SCOREBOARD_TEST_DATABASE_URL` is not available locally, so the new PostgreSQL test was among the skipped coverage; the namespace leak was reproduced through the exact installed Tortoise 1.1.8 store path.

HupBaHa (changes requested):
Two new blockers remain at `f7fc06b7`; the prior namespace-collision and shared-PostgreSQL cleanup findings are fixed.

[P1] `apps/scoreboard/src/scoreboard/routes/scores.py:183` checks visibility separately from `ScoreStore.submit()` at `apps/scoreboard/src/scoreboard/scores/store.py:555`. In `auth_mode=disabled`, `_resolve_submitter` has already accepted the body claim. If the seed/config path flips the board public -> private between those reads, the route passes and the store persists the request with private-board semantics and an unverified claimed identity. I reproduced the exact ordering by flipping visibility immediately before the store call: the response was 201, the board was private, and attacker-controlled metadata was stored under the claimed victim address. The verified-identity decision needs to be coupled to the visibility decision that actually governs persistence (or the store must refuse a private write when identity was not verified); another earlier visibility read does not close the race.

[P2] `apps/scoreboard/src/scoreboard/purge_reserved_idempotency_keys.py:49` deletes every `sfp-`/`sfu-` mapping, including legitimate mappings created by new pods before the post-upgrade hook runs. This is more than losing a fast path: content-hash dedup only saves an identical payload. I reproduced creating a private score under `same-key`, running the purge, then retrying a changed payload under `same-key`; the second request returned `created=True` with a different score. That violates the documented same-key-replays-original contract during every affected upgrade. The rollout cleanup must distinguish legacy/old-replica rows from mappings created under the new scheme, or use a rollout strategy/namespace transition that does not delete live new mappings.

Verification: focused privacy/idempotency/migration suites 260 passed, 3 skipped; full Scoreboard 486 passed, 3 skipped, 3 deselected; Ruff check/format, Pyright, Helm lint/template, diff check, and all GitHub checks passed. `SCOREBOARD_TEST_DATABASE_URL` was unavailable, so the isolated PostgreSQL runtime test skipped locally.

FilipBolt:
Thanks — all six findings across your three reviews are fixed at `15efb127`, each reproduced before being touched and mutation-checked after. Two of them changed the design rather than patching it, and one of your reproductions pointed at something wider than the finding; details below.

## Review 1 (08-25)

**1. Forged writes readable through POST dedup.** Reproduced: the forged POST returned the victim's row with its `metadata`. A private board now refuses writes without a verified submitter, and the refusal happens *before* the dedup lookup — refusing after it would still confirm whether a matching row exists.

**2. The scoped key was impossible on PostgreSQL.** The NUL is gone; the stored value is a bounded, printable digest. Covered by a PostgreSQL-backed test.

**3. Identity-scoped responses were shared-cacheable.** `Cache-Control: private, no-store` plus `Vary` on every private exit, refusals included. While fixing that I found the two 404s in `get_score` were distinguishable by their headers alone, which defeated the matching body — both now carry the same policy.

**4. The concurrent-retry branch used the unscoped key.** It uses the scoped one. Separately, that branch's success return was uncovered when you flagged it and now has a test, so `store.py` is at 100% statements.

## Review 2 (08-26 10:26)

**5. Public raw keys could collide with private scoped keys.** Both `sfp-` and `sfu-` are reserved now. Rather than a prefix list, the property is asserted directly: no value the generator produces can be stored verbatim. Renaming a prefix and forgetting to reserve it fails that test.

**6. The PostgreSQL test deleted every idempotency row.** Isolated into a UUID schema, without editing your test. I also added a guard that the isolation fixture still matches it — renaming the test would otherwise re-arm the destructive cleanup silently.

## Review 3 (08-26 22:50)

**7. The route/store visibility race.** Reproduced exactly as you described: `201`, board private, attacker metadata under the claimed victim address. Fixed by deleting the route's read rather than adding a third — the store takes both decisions at one read and refuses before looking anything up.

**8. The purge broke `same key replays the original`.** Your reproduction used a private board, and on a private board it was **already broken before the purge ran** — by an earlier decision of mine to refuse every mapping pointing at a private row. So the finding was real and its cause was wider than the purge.

The purge is **deleted**, not fixed with a marker column. Once a private target is honoured for its verified owner, a stale mapping to someone else's row is refused at the lookup whatever the table holds — so the purge was not load-bearing for privacy and was only deleting live mappings. Migration `0009` stays: it is a `pre-upgrade` hook, so it runs before any new pod can have written a new-scheme row and can only ever delete legacy ones. Same SQL, opposite safety, decided by when it runs.

## Beyond your findings

Your review 3 prompted a sweep for the same shape elsewhere, and `visibility` turned out to be stale-readable at three more layers: every read path, a response body carrying a stale copy of it, and one exit of one branch. All closed. Reads **re-decide** rather than erroring, which a read can and a write cannot; writes answer **409** and take `select_for_update()` on the benchmark row inside the insert transaction.

Two things worth your scepticism specifically:

- ~~**The lock is invisible to the suite.** Tortoise no-ops `select_for_update()` on SQLite, so the tests demonstrate the revalidation and PostgreSQL is what the lock is for. Revalidation alone shrinks the window to the commit interval; it does not eliminate it. I would rather say that than imply the tests cover it.~~

  **Correction — this was false.** No lock was being taken on any dialect: `values_list()` dropped the `FOR UPDATE` that `select_for_update()` had set. And the SQL can be rendered without a PostgreSQL server, so the reason given for not testing it was wrong too. See [the reply below](https://github.com/ScreamingFace/screamingface/pull/719#issuecomment-5437565681); fixed at `4b5a8a82`.
- **Three of the concurrency tests are driven at a seam** — controlling what a re-resolve returns — because reaching those branches needs a racer to commit mid-transaction and the fixture is single-connection SQLite. `SCOREBOARD_TEST_DATABASE_URL` is still the only way to test them for real, which is OME-430's ground.

Since this class was found four times at four layers, it is now asserted rather than fixed instance by instance: `tests/unit/guards/test_visibility_exit_guard.py` checks, for every function deriving a decision from `visibility`, that each `return` and `raise` is preceded by a revalidation — with the exits that legitimately need none listed alongside the reason. Four regression shapes are proven caught.

The PR description is rewritten to match; the previous one described a response shape from several rounds ago.

HupBaHa (changes requested):
The two prior fixes moved in the right direction, but three concrete gaps remain at `15efb127`.

[P1] `apps/scoreboard/src/scoreboard/scores/store.py:628-633` loses the PostgreSQL row lock it relies on. With the pinned `tortoise-orm==1.1.8`, `select_for_update()` sets lock state on the current `QuerySet`, but `values_list()` constructs a new `ValuesListQuery` without copying that state. I generated SQL from the installed runtime:

```sql
-- rows.select_for_update().sql()
SELECT ... FROM "benchmarks" WHERE "id"=$1 FOR UPDATE

-- rows.select_for_update().values_list("visibility", flat=True).sql()
SELECT "visibility" "0" FROM "benchmarks" WHERE "id"=$1
```

So visibility can still flip after `_revalidate_visibility()` and before `_insert_new_score()`. Fetch the model row under `select_for_update()` (or otherwise issue a real locking query) rather than projecting through `values_list()`. The SQLite tests cannot hold this because SQLite does not implement this lock.

[P1] The replay revalidation checks the requested benchmark, not the benchmark owning the row being returned. `_resolve_owned()` may resolve a global public idempotency key to a score on another benchmark, but `store.py:721` and `store.py:768` revalidate only `submission.benchmark_id`. Executed schedule: key resolves victim score on public B; request targets public A; B flips private after resolution; A revalidates unchanged; response returns B with `metadata={"secret":"victim-only"}` and `created=False`. The normal replay and IntegrityError replay need to validate/read-gate the actual returned score after the final visibility decision.

[P2] Removing the destructive post-upgrade purge preserves legitimate mappings, but `_resolve_owned()` now trusts a rollout-window poisoned private token when its public target row carries the future owner email (`store.py:521-540`). Executed schedule: an old replica stores the predictable future `sfp-` token verbatim and binds it to an attacker-controlled public score with `submitted_by=alice`; after the target board becomes private, Alice makes a verified same-key submission; the new code returns `created=False`, the attacker row, and attacker metadata instead of creating Alice’s private submission. The assertion in `test_store.py:1694-1700` that only a wrong public replay remains is therefore too weak: that public row can become the result of a private request. New-format mappings need provenance/versioning, or the rollout path must otherwise prevent legacy rows from being honoured as private-owner mappings.

Verification: exact installed Tortoise 1.1.8 source and generated PostgreSQL SQL inspected; all three schedules executed; focused privacy/idempotency suites 273 passed, 3 skipped; full Scoreboard 499 passed, 3 skipped, 3 deselected; Ruff check/format, Pyright, Helm lint/template, diff check, and all GitHub checks passed. No live PostgreSQL URL was available; the missing `FOR UPDATE` is established by installed-source and generated-SQL evidence.

FilipBolt:
All three are fixed at `4b5a8a82`. Each was reproduced first, and each fix is mutation-checked. Before the findings themselves, a correction that is owed:

## A claim I published that was false

My last reply said "**the lock is invisible to the suite** — Tortoise no-ops `select_for_update()` on SQLite, so the tests demonstrate the revalidation and PostgreSQL is what the lock is for. I would rather say that than imply the tests cover it." The PR description said the same.

Both halves were wrong. The lock was not being taken **on any dialect**, so there was nothing for PostgreSQL to enforce — and the reason I gave for not testing it was wrong too, since rendering the SQL needs no server at all. You did exactly that from the installed runtime. I asserted a property about a dialect and never rendered it against that dialect, then presented the gap as a considered limitation. The description and this thread are both corrected.

## [P1] The row lock

Confirmed against `tortoise-orm==1.1.8`: `select_for_update()` sets lock state on the `QuerySet`, `values_list()` builds a fresh `ValuesListQuery` without copying it.

The revalidation now runs `.only("visibility").first()` — a model projection, which preserves the lock. `.first()` was checked separately not to strip it; it only adds a `LIMIT`.

The query is built by `ScoreStore.visibility_query()` rather than inline, so a test can render **the query production runs**. `test_the_persist_path_really_locks_the_row` initialises the asyncpg dialect and asserts `FOR UPDATE` in the rendered SQL. My first version of that test built its own equivalent query, which would have kept passing while `_revalidate_visibility` regressed underneath it — the same failure shape as the original bug, one level up.

## [P1] The replay revalidated the wrong benchmark

Your schedule reproduces exactly: `created=False`, `got_victim=True`, `metadata={'secret': 'victim-only'}`, board B private.

`_confirm_replayable()` now revalidates the **request's** board and then re-applies the read gate to the **returned row**, whose benchmark can differ. Both the normal replay and the `IntegrityError` replay go through it.

Worth reporting: the exit guard added last round caught this the moment revalidation moved into a helper it did not know about. That is the first edit it has had to judge, and it judged it correctly rather than needing to be taught the new shape.

## [P2] The poisoned legacy token

Reproduced: `created=False`, `got_attacker=True`, `metadata={'secret': 'attacker-controlled'}`.

`IdempotencyKey.scheme` records which code wrote a mapping. An old replica's `INSERT` leaves it `NULL`, and a reserved-namespace mapping is honoured only when it carries the current scheme. Migration `0010` adds the column, nullable and deliberately **not** backfilled — `NULL` is the meaningful value, so backfilling would erase the only signal.

**This reverses my own argument.** I wrote that a marker column "would fix that case and not the private one, and would carry a schema change forever to serve one rollout". That was right about the purge and wrong as a claim about provenance: the row is public, but the *request* is private, and with a forged `submitted_by` the ownership branch honoured it for the owner it named. Your reading of `test_store.py:1694-1700` was right that the assertion was too weak. Both halves of that comment are now corrected in place, with a pointer to the test that disproves them.

Two things came out of the fix that are worth your attention:

- **The provenance gate re-introduced an earlier denial.** A refused legacy row became invisible to the reclaim, so the caller's own write was blocked by a mapping we had just decided not to honour. Two of this PR's own tests caught it; the reclaim predicate now covers un-provenanced rows.
- **`_bind_idempotency_key` was writing unstamped mappings** — we would have refused our own fast path. It degrades rather than breaks, which is why nothing caught it; there is now a test that every write site stamps, not just the insert.

## On verification

My mutation harness silently produced four false "caught nothing" results — `timeout` does not exist on macOS, so the runs never started and absent output read as absent failures. Fixing that is what surfaced the `_bind_idempotency_key` gap above, since it let each write site be mutated separately. Harnesses now assert the run happened rather than inferring it from silence.

Gates green, 505 passed / 3 skipped, `makemigrations` reports no drift.

HupBaHa (approved):
Re-reviewed exact head `4b5a8a82417d2de8735609d03363e6d64f138c4d`. The three prior blockers are closed.

- The persist-path visibility query now renders real PostgreSQL locking SQL under the pinned Tortoise 1.1.8: `SELECT "visibility" ... LIMIT $2 FOR UPDATE`.
- Normal and IntegrityError replays both pass through `_confirm_replayable`, which revalidates the requested board and freshly read-gates the actual returned row. The previous cross-benchmark flip schedule now raises `BenchmarkVisibilityChanged` rather than returning victim metadata.
- Migration `0010` adds nullable, unbackfilled mapping provenance; old replicas leave `scheme=NULL`, while both current mapping write paths stamp `v2`. The previous rollout-poison schedule now creates Alice’s private submission, reclaims the legacy mapping, and stores a `v2` mapping instead of replaying the attacker row.

No new blocking privacy, migration, compatibility, or idempotency findings. Verification: targeted closure schedules all passed; focused privacy/idempotency suites 279 passed, 3 skipped; full Scoreboard 505 passed, 3 skipped, 3 deselected; Ruff check/format, Pyright, Helm lint/template, diff check, and all GitHub checks passed. No live PostgreSQL URL was available; locking was validated against exact installed 1.1.8 source and generated asyncpg-dialect SQL.

HupBaHa (changes requested):
A fresh rollback-focused pass found one release-blocking privacy gap at the unchanged head `4b5a8a82`, plus one broken opt-in test. I am revising the previous approval.

[P1] The documented `helm rollback` path can publish every private submission. Visibility is persisted by the seed Job, whose only hooks are `post-install,post-upgrade` (`apps/scoreboard/charts/scoreboard/templates/job-seed-benchmarks.yaml:9`); Helm rollback has distinct `pre-rollback`/`post-rollback` lifecycles, and this chart has neither. `apps/scoreboard/DEPLOYMENT.md:218-225` explicitly instructs operators to run `helm rollback` and notes that DB migrations/state are not reverted. Rolling back to base `454253da` therefore leaves `benchmarks.visibility=private` in the database while restoring code that never reads visibility and serves unscoped leaderboard/history/frontier data.

This was executed, not only inspected: current head created a private benchmark with a victim score; exact base `454253da` was then run against the same database, and `ScoreStore.leaderboard("private-challenge")` returned one row with `submitted_by=victim@example.test` and `url4=url4://private/victim`. The safe activation/rollback invariant must be explicit and enforceable before this capability is used: deploy a privacy-aware version as the rollback baseline before any board becomes private, and prohibit/guard rollback below that minimum revision once private data exists (or provide an equivalent fail-closed maintenance procedure). The current generic rollback command is unsafe for this feature.

[P2] The opt-in PostgreSQL regression no longer reaches PostgreSQL. `apps/scoreboard/tests/unit/scores/test_idempotency_postgres.py:54-59` submits to a private board without `identity_verified=True`; current `ScoreStore.submit()` rejects at `store.py:787-790` before opening the transaction. Because `SCOREBOARD_TEST_DATABASE_URL` is absent in default CI, this stays green as a skip and fails only in the environment it is supposed to validate. Pass verified identity for both calls (and ideally assert the stored `scheme`) so the production-only regression remains executable.

The prior three closure fixes remain valid: real `FOR UPDATE`, cross-benchmark replay gating, and `scheme=v2` provenance all rechecked cleanly. Existing same-head gates remain green (`505 passed, 3 skipped, 3 deselected`; Ruff/format/Pyright/Helm/GitHub CI). No live PostgreSQL server was available; the P2 failure is established by the unconditional private-write guard before any DB transaction.

FilipBolt:
Both fixed at `841a8336`. Each reproduced first. One of them changed my mind about where the fix belongs, and one mutation didn't bind on the first try — details below.

## [P1] `helm rollback` publishing a private board

Confirmed by execution, not inspection. Base `454253da`'s `src/scoreboard/` contains **zero** occurrences of `visibility`. Extracted it with `git archive`, ran it against a PostgreSQL holding one private board with three submissions:

```
leaderboard('healthbench-worst30') returned 3 rows:
    spec-0 | p0@example.test | url4://x/0
    spec-1 | p1@example.test | url4://x/1
    spec-2 | p2@example.test | url4://x/2
```

There is no configuration of the old code that filters, because it has nothing to filter on.

**I went looking for a `pre-rollback` hook and did not ship one.** `helm rollback` executes the *target* revision's hooks — `execHook(targetRelease, release.HookPreRollback, ...)` in `pkg/action/rollback.go`. The revision being rolled back *to* is a pre-privacy one whose stored manifest has no such hook, so nothing added to this chart can run in the dangerous direction. It would have looked like a fix and been theatre. I checked this rather than assuming it, having spent the previous round correcting a claim I made about a dialect without ever rendering against it.

So the control is the operator preflight you left room for:

```
$ python -m scoreboard.check_rollback_safety
REFUSED: 2 private benchmarks would be published by rolling back below scoreboard 0.1.1.

  empty-challenge      0 submissions  (Empty Challenge)
  healthbench-worst30  3 submissions  (HealthBench Worst-30% Challenge)

This code reads `Benchmark.visibility`; releases below 0.1.1 do not. A rollback leaves
`visibility=private` in the database and restores code that serves every row unscoped.

Run the fail-closed procedure in apps/scoreboard/DEPLOYMENT.md ("Private boards and rollback") first.
exit=1
```

Four decisions in it worth disagreeing with if you do:

- **No override flag.** Clearing the refusal by flipping a board public *is* the leak, performed deliberately.
- **An empty private board is still refused.** The check and the rollback are not one operation, and a board configured private is one that accepts submissions the moment the challenge opens. The configuration is the hazard, not the current row count.
- **The floor is the tool's own running version, not a constant.** The privacy-aware release did not exist when the module was written, so a hardcoded number would have been a guess that then rotted. This code demonstrably reads `visibility`; anything below the release shipping it demonstrably does not.
- **The exit code is the enforcement.** A printed warning is what "advisory" means.

`DEPLOYMENT.md` gains the invariant, the activation ordering rule (deploy the privacy-aware release *before* flipping any board private, so a safe rollback target already exists in `helm history`), and the fail-closed procedure: scale to 0 → export → delete → roll back. The generic rollback snippet at `:218` now points at it, since that is the block an operator actually reads.

One thing the tool cannot check, stated rather than papered over: it sees the database, not the release history. It cannot tell you whether a safe rollback target exists — only whether rolling back is currently dangerous.

## [P2] The opt-in regression that no longer reached PostgreSQL

Reproduced against a throwaway PostgreSQL 17: `PrivateBoardRequiresIdentity: private-pg`, exactly where you said, before any transaction opened. Fixed with verified identity on both calls, plus an assertion that `scheme` round-trips through a real `CharField` — a column that silently failed to persist would make every reserved-namespace mapping look legacy.

**The instance was the smaller half.** The test was green-by-skip everywhere anyone looked, which is the property worth removing. CI now runs it: a `postgres` job with a `postgres:17-alpine` service. The whole suite cannot run there — setting `SCOREBOARD_TEST_DATABASE_URL` routes the shared `tortoise_db` fixture through `postgres_schema_database_url`, whose `asyncio.run()` inside a running loop errors out **260 tests** (OME-430; measured this round rather than inferred from the docstring). So the job names the two self-contained modules, and a guard asserts that list still covers every module that skips itself without a database.

That guard took three attempts, and the failures are the useful part. Detecting "a PostgreSQL test" by filename matched the guard itself. Detecting it by searching the source for `skipif` also matched the guard — it writes those names down in order to look for them. The rule that works reads the **AST**: a `skipif` decorator or a `pytest.skip()` call is syntax; the same words in a string literal are not.

## Two mutations that did not bind

Reporting these because they were both in this round's own new code.

- **The exit code was untested where it is decided.** Mutating `0 if not boards else 1` to a bare `0` left the suite green: both `main` tests monkeypatch `_run`, so they proved the code was *propagated* while nothing proved it was *derived* — and that code is the whole enforcement. Closed by driving the real `_run` against the test connection.
- **The doc-pointer guard matched the wrong line.** It asserted the section name appeared in `DEPLOYMENT.md`, and the name also appears in the cross-link I had just added, so renaming the heading kept it green. It matches a heading line now.

Gates green, 518 passed / 3 skipped, the PostgreSQL pair passing against a real server, `helm lint` and `helm template` clean.

FilipBolt:
Following up on the rollback finding with the full picture, and then a proposal about landing this.

## Your P1, on all four read paths

I had only demonstrated `ScoreStore.leaderboard()`. It is every read path. Same PostgreSQL database throughout — one private board, three submissions, two participants — served by two code versions through the real app, anonymous caller, no auth:

| `GET` | base `454253da` | head `841a8336` |
| -- | -- | -- |
| `/v1/benchmarks` | `200` — listed, **with no `visibility` field at all** | `200` — `"visibility":"private"` |
| `/v1/leaderboard/{id}?top=50` | `200` — all three entries: ranks, scores, providers, spec ids | `200` — `entries:[]`, `my_submissions:[]`, `scoped_to_caller:true` |
| `/v1/leaderboard/{id}/{spec}/history` | `200` — the participant's submission | `404` |
| `/v1/leaderboard/{id}/frontier` | `200` — full trend across all three | `404` |

The `/v1/benchmarks` row is the one I would not have predicted: the base response has no `visibility` field, so a client could not tell the board had ever been marked private. There is no client-side mitigation available during that window.

I also ran the documented procedure against that database rather than only writing it down: the preflight refuses with `exit=1`, and `export_private_submissions` returns all three rows with full addresses, so step 2 has something to hand to step 3 before the delete.

## A proposal on scope

Six reviews in 54 hours, and every one of them found something real — I am not asking you to lower the bar. But the severity has been walking down: unauthenticated read of private submissions, then a missing lock and a cross-benchmark replay, then an operational hazard plus test hygiene. That is the shape of a change converging.

Two facts about what merging actually does:

- **The feature ships inert.** No benchmark is private, and the chart's `authMode: disabled` means a private board would be readable by nobody through the API even if one existed. Merging exposes nothing and changes no existing board.
- **The remaining risk is activation, not code.** It becomes live when someone sets `visibility: private` on a `seedBenchmarks` entry — a config change, in a separate deploy, by a person.

So I would like to propose: **merge unless the next review finds a P1.** Anything P2-and-below goes to a follow-up issue on an activation checklist — the same checklist that already has to carry `authMode` and the rollback floor before the entry challenge can open. Those gates want to be tracked where an operator will see them, not in a PR thread that closes.

**The first follow-up is filed, so this is a pointer and not a promise.** `OME-1027` (High, Week 4) carries the rollback gate: the activation ordering rule, the four-path evidence above, and three concrete done-when conditions — a privacy-aware release deployed and selected by `helm history`, the preflight run once in a production pod, and the person flipping the first board having read the procedure that destroys participant data by design.

If you do not have Linear access, the operative half is already in this PR: `apps/scoreboard/DEPLOYMENT.md`, "Private boards and rollback". The issue adds only the thing the preflight structurally cannot check — whether a safe rollback target exists at all, which is release history rather than database state.

If you would rather keep something in this PR, say which and I will do it here — I would just rather that be a deliberate choice than the default.

CI is green on `841a8336`, including the new PostgreSQL job running against a real server for the first time.

HupBaHa (approved):
Re-reviewed exact head `841a833656d12707a86763898858c6c5309c9b5c`. The prior PostgreSQL regression gap is fully closed: verified identity reaches the database, scheme provenance is asserted, a dedicated PostgreSQL 17 CI job executes the opt-in modules, and I independently ran the pair against disposable PostgreSQL 17 (3 passed). The rollback finding no longer leaves a merge-time P1: shipped chart/production configuration contains no private board, the feature is inert, and this head adds the refusing preflight, activation ordering rule, and fail-closed procedure. No new privacy blocker was found.

Two activation-only P2 details are recorded on OME-1027 rather than blocking this dark merge: package semver cannot identify the safe Helm/image rollback floor because both old and new code are `0.1.1`, and the destructive fallback still needs executable wait/export/delete/verification steps before first activation. Do not set `visibility: private` until that gate is closed or explicitly accepted.

Verification: 518 passed / 3 skipped / 3 deselected; focused remediation tests 13 passed / 1 expected PostgreSQL skip; disposable PostgreSQL regressions 3 passed; portal 25 passed; Ruff check/format, Pyright, lock check, Helm lint/template, diff check, all visible GitHub checks green; merge state CLEAN.

PR #726, opened 2026-08-25
feat(scoreboard): retire the legacy news demo benchmarks

FilipBolt:
Closes OME-986.

`hle`, `livetruth` and `livetruth-latest` are leftovers from the previous SF project and were still advertised on the public catalogue — on the board being handed to internal testers this week. Requested by @Irina in `#scream-dev`.

## The part that isn't obvious

**Removing them from the chart's seed list does not remove them.** `seed_benchmarks` only registers and updates; there was no deletion path anywhere in the codebase. Emptying the list stops them being *recreated* and changes nothing a reader of the board sees — the rows persist and `GET /v1/benchmarks` keeps serving them.

So this adds the missing half, and the config change alone is deliberately not the deliverable.

## Why now

All three are empty today — 0 entries, 0 baselines. Both `Score.benchmark` and `Baseline.benchmark` are `on_delete=RESTRICT`, so deletion is refused the moment anything references them. One tester submitting against `hle` turns a clean delete into a data conversation.

## Design

This is the **first operator module here that destroys anything** — every other one is additive (`seed`, `import_baselines`) or read-only (`export_private_submissions`). So:

- Deletion is opt-in behind `--yes`. The default outcome of a correct-looking command is a report.
- A benchmark that any score or baseline references is **refused and survives**, and `--yes` does not override that. It confirms intent to delete something unreferenced; it is not a force flag.
- An unknown id is refused with a non-zero exit, because "already gone" and "you typed it wrong" must not look identical to someone cleaning up a live board.

`RESTRICT` means the database would refuse regardless. The module's job is naming *which* rows stand in the way rather than surfacing an `IntegrityError` traceback at an operator. `Score` and `Baseline` are the only two models referencing `Benchmark` (`IdempotencyKey` links to `Score`), so counting those two is a complete check.

**Rejected:** teaching the seed job to prune anything absent from its list. During an Engine catalogue outage `published_rows` is empty, so it would delete every Engine-published benchmark — the same failure shape review caught in OME-894, and worse here because re-running would not recover it. Also rejected: a data migration, which would re-run against every fresh database including local dev.

## Test plan

The refusal path was built and tested **before** the deletion path existed, since destroying submissions is this module's only real risk.

Both guards are **mutation-proven**, not merely passing:
- remove the `--yes` gate → `test_without_confirmation_nothing_is_deleted` fails
- remove the refusal → 4 tests fail, one with `IntegrityError: FOREIGN KEY constraint failed`, which also shows `RESTRICT` working as the second line of defence rather than the first

End-to-end against a real database: dry run changed nothing; the three legacy ids retired; `draco`, holding one score, was refused **even with `--yes`** and survived. Exit codes verified: refusal `2`, unknown id `2`, dry run `0`, confirmed delete `0`.

## Before this has any effect on the live board

**Two things, in this order.**

1. **Confirm with @Stephen which values file feeds the deployed seed job.** The chart here lists all three and sets `engineUrl: ""`, yet the live board serves five Engine-published benchmarks — so the deployed values are not this repo's defaults (OME-730). Without that, the config half lands nowhere.
2. **Then run the module against the deployed database.** Emptying the list does not touch rows that already exist:
   ```
   python -m scoreboard.retire_benchmark --benchmark hle --yes
   ```
   Order matters — run it *after* the config change is deployed, or the next deploy recreates them.

The five Engine-published benchmarks are untouched and were never in the chart list.

### DIFF
--- a/apps/scoreboard/DEPLOYMENT.md
+++ b/apps/scoreboard/DEPLOYMENT.md
@@ -118,27 +118,41 @@ Leaving `config.authMode` at its default (`disabled`) keeps today's behavior —
 
 The app chart runs `python -m scoreboard.seed` after install and upgrade. Seed data comes from `.Values.seedBenchmarks.benchmarks` and is passed through `SCOREBOARD_SEED_BENCHMARKS_JSON`.
 
-The default seed registers HLE plus the Livetruth public datasets:
+**The default seeds nothing.** `benchmarks` is `[]` and `engineUrl` is empty, so a fresh install registers no benchmark at all and `GET /v1/benchmarks` returns an empty list. The legacy `hle` / `livetruth` / `livetruth-latest` demo entries were removed in OME-986; they were leftovers from the previous SF project.
+
+A real deployment gets its benchmarks from the Engine, by pointing `engineUrl` at the Engine's **in-cluster** address:
+
+```yaml
+seedBenchmarks:
+  enabled: true
+  engineUrl: http://screamingface-engine.screamingface.svc.cluster.local:8000
+  benchmarks: []
+```
+
+To register a benchmark the Engine does not publish — a local smoke target, say — list it explicitly:
 
 ```yaml
 seedBenchmarks:
   enabled: true
   benchmarks:
-    - id: livetruth-latest
-      display_name: News Livetruth Latest
-      description: OpenMined Livetruth latest demo dataset
-      dataset_url: https://scoreboard.screamingface.ai/livetruth-latest.jsonl
-    - id: hle
-      display_name: News Hallucinations
-      description: OpenMined HLE benchmark
-      dataset_url: https://github.com/openmined/HLE.jsonl
-    - id: livetruth
-      display_name: News Livetruth
-      description: OpenMined Livetruth benchmark
-      dataset_url: https://scoreboard.screamingface.ai/livetruth-masking.dataset.jsonl
+    - id: smoke
+      display_name: Smoke
+      description: Local smoke target, not an Engine benchmark
 ```
 
-Re-running the Job is safe because benchmark registration is an upsert. Disable seeding with `--set seedBenchmarks.enabled=false`.
+An entry whose id the Engine also publishes is ignored and named in the Job's output; the Engine is the only place a published benchmark's text is written (OME-904).
+
+Re-running the Job is safe because benchmark registration is an upsert.
+
+**Seeding never deletes.** Removing an entry from this list stops it being recreated; it does not remove a row that already exists, which stays and is still served. To remove one:
+
+```bash
+python -m scoreboard.retire_benchmark --benchmark <id> --yes
+```
+
+It refuses while any score or baseline references the benchmark, and refuses an Engine-published one unless you also pass `--include-engine-owned` — which is only correct once the Engine has stopped publishing it, since otherwise the next seed recreates it.
+
+Disable seeding with `--set seedBenchmarks.enabled=false`.
 
 ## Smoke Checks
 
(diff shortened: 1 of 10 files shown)


HupBaHa (changes requested):
Thank you for the thoughtful implementation and the additional race/Engine-ownership hardening. The core safety properties look good: referenced benchmarks remain protected by both the explicit checks and the RESTRICT foreign keys, dry-run is the default, and `--yes` cannot override submitted data.

Before merge, could you please address three small items?

1. Reconcile the documented scope.

Owner decision: `retire_benchmark` may be a reusable operator command for any unreferenced benchmark, not only the three OME-986 legacy IDs. Please keep `--include-engine-owned` for the stranded case where the Engine no longer publishes a benchmark, and update the Linear/spec/plan/ledger wording to record that scope. The reference guard must remain unconditional, and the documentation should continue to warn that an actively published Engine benchmark will be recreated.

2. Add tests for the actual CLI boundary.

Please cover the default dry run, `--yes`, unknown and referenced benchmark exit code 2 without a traceback, `--include-engine-owned` wiring, and database initialization/cleanup on success and refusal. The current tests cover `retire_benchmark()` well, but the parser, `_run()`, and `main()` paths are currently uncovered.

3. Check the delete row count.

`QuerySet.delete()` returns the number of deleted rows. Please require exactly one deleted row; if it returns zero, report that the benchmark disappeared concurrently rather than printing `retired`. Please add a regression test for this path. If practical, include the revision condition in the final delete predicate so the Engine-owned guard remains atomic.

After these corrections I expect the PR to be merge-ready. I found no issue with the FK protection, dry-run default, reference refusal, or the score/baseline check-delete race handling.

FilipBolt:
All three addressed in `450f917e`. Thanks — the parenthetical on item 3 was the one worth pulling.

**1. Scope reconciled.** `retire_benchmark` is recorded as a general operator command for any unreferenced benchmark, not just the three legacy ids. Reference guard stays unconditional, `--include-engine-owned` stays for the stranded case, and the docs keep warning that an actively published Engine benchmark is recreated by the next seed. Updated in the spec (D8), the plan and the ledger — the plan's out-of-scope line still excluded "the five Engine-published benchmarks", which directly contradicted the decision. OME-986's description updated to match.

**2. CLI boundary covered.** New `tests/unit/test_retire_benchmark_cli.py`: dry-run default, `--yes`, exit code 2 with no traceback for both the unknown and the referenced case, `--include-engine-owned` wiring, and that the override alone does not delete without `--yes`. It runs against a real on-disk SQLite database rather than the `tortoise_db` fixture, deliberately — `main()` goes through `Settings()`, `init_db()` and `close_db()`, and a fixture-provided connection would hide whether the CLI opens and closes its own on both the success and the refusal path. One test reads the database back after a refusal to prove exactly that.

**3. The delete now confirms it deleted — and following your atomic-predicate note found two more paths.**

- The row count is required to be exactly `1`; anything else is a refusal naming what it matched.
- `Benchmark.get()` raised `DoesNotExist` when the row disappeared between the existence check and the revision read — a third unhandled path `main()` does not catch. Now `get_or_none()` and a refusal.
- The Engine-ownership guard is repeated in the `DELETE` predicate, so a benchmark that gains a revision between the read and the write is not removed by a call that refused to touch Engine-owned rows.

All three have tests that fail without the fix. Both new guards are mutation-proven: dropping the `--include-engine-owned` wiring fails two CLI tests, ignoring the row count fails two decision tests.

`retire_benchmark` exceeded the complexity gate after this, so the delete-and-verify block is extracted into `_delete_exactly_one`.

Worth noting the pattern: this is the second round where the finding sat directly beside something I had just hardened — the discarded row count is the same defect family as the check/delete race fixed last round.

HupBaHa (approved):
Re-reviewed at 450f917e. All three requested changes are resolved: the reusable scope is reconciled across the spec/plan/ledger; the real CLI boundary covers dry-run, confirmation, refusal exits, lifecycle, and Engine override wiring; and deletion now requires exactly one affected row while keeping the Engine-ownership condition in the DELETE predicate. The additional get_or_none race handling is appropriate.\n\nVerified locally on the PR head: 26 focused retirement tests passed; full Scoreboard suite passed (384 passed, 2 skipped, 3 deselected); Ruff check, Ruff format check, and Pyright passed. CI is green. No remaining findings.

Issue #740, opened 2026-08-25
DRACO grading fails with 'invalid Criterion envelope' on all cases of a run; candidates answered normally (billed, no score)


peteride-kostic-lgtm:
## Summary

All eight case-gradings in a 4-candidate DRACO run failed at the **grading stage** with `draco_grading_failed: "DRACO Case evaluation contains an invalid Criterion envelope"` (`error_kind: ResolutionError`), even though the candidates answered normally (`finish_reason=stop`, real token usage billed). Earlier the same day, the identical grading path scored five candidates on the same benchmark revision without issue — this looks like a judge-pipeline regression or an upstream judge-model hiccup rather than anything candidate-side.

**Environment:** screamingface 0.1.1.post5, hosted engine `https://fusion.dev.screamingface.ai`, benchmark `draco` revision `66a463248586b277`, `limit=2`, 2026-08-26 (~18:3x UTC+... server time in run records).

## Repro

```python
def solo(route):
    return sf.Model(route, params={"max_tokens": 24000})

solos = [solo("openrouter/deepseek/deepseek-v4-pro"), solo("openrouter/qwen/qwen3.8-max"),
         solo("openrouter/moonshotai/kimi-k3"), solo("openrouter/z-ai/glm-5.2")]
report = sf.evaluate(solos, benchmark="draco", limit=2, progress=False)
```

Result: every candidate `score=None, coverage=0.0`; 7 of 8 cases show `stage: 'grading'` / `invalid Criterion envelope` (the 8th failed earlier at the candidate stage with `finish_reason=length`, unrelated — see #737).

## run_ids (for server-log correlation)

- deepseek-v4-pro: `RyK3NFwyK6GLDTtoWjxUbYiv75O7y74EL69cS7XxG9K6u597jYL7qTCVkhfXc9Sw`
- qwen3.8-max: `mmJc6W6soajvRQecFJRbNBk6Oq3Rv2uY6Qd1b9gwCzLsRdjNa3NPebVXDFcBTGpm`
- kimi-k3: `bvSJlUsfsi9SWcTlpopwtDqdUutmwE5OprbHye18f86TBloZUfKJvxFcQabaU0WX`
- glm-5.2: `H8389sb4mUD7WAwsOm3LrTVekCqsgEvVmA4mpctvd4giIUZf6fKAF7ZX3gPdKWhZ`

## Details that may help

[...]

PR #742, opened 2026-08-26
fix (screamingface/benchmarks): Surface real judge failures, throttle the DRACO judge, and stop billing runs into an envelope error

khoaguin:
Fixes the DRACO grading incident where every case failed with `invalid Criterion envelope` while the run still billed (GH #740, `OME-993`).

### Before
- Trigger: a researcher runs a benchmark; their models all answer normally.
- Today: when a judge call fails mid-grading (rate limit, or the reasoning judge burning its whole token budget thinking), the run reports a cryptic internal format error — `invalid Criterion envelope` — on every case. The run is billed in full, scores never arrive, and the error points at the wrong component.
- Cost: paid runs with nothing to show; the failure looks like the researcher's models when it is the engine's judge pipeline.

### After
- Same trigger.
- Now: the judge is throttled to its official low-reasoning effort with a doubled token budget, transient judge failures retry automatically, and if grading still fails the case names the real cause (`model_token_cap`, `rate_limited`, …) at `stage=grading` with the candidate's answer preserved and a retryability flag.
- Win: a billed run always ends in either scores or an honest, actionable failure.

### Don't regress
- Normal grading output and scoring stay byte-for-byte as today (template re-pinned deliberately, once).
- Candidate-stage failures keep their own distinct stage and now carry their original gateway code too.

## Rebased on OME-924 (PR #741)

A sibling fix landed mid-flight: url4 error rows now carry `code`/`retryable`, and the benchmark fan-outs are **fail-fast** (`on_error="fail"`), so the original judge error propagates to the per-Case boundary directly. This branch is rebased on it: the case-evaluation seam stays as the belt-and-braces guard for any other collect boundary, and the DRACO template pin is recomputed over fail-fast + retry + the new judge params.

## Root cause (two bugs stacked)

1. A judge call failed at transport level (429/5xx, or an all-reasoning `length` turn at `max_tokens=4096` with no reasoning throttle).
2. url4's default `iterate(on_error="collect")` turned the failed criterion into an `{"error": ...}` row inside the criteria array; the case-evaluation route then rejected the row's **shape** ("invalid Criterion envelope"), destroying the real message. Full investigation on `OME-993`.

## What this PR does (one revision bump, four stacks)

- **url4**: collected error rows also carry the failure's `code` and `permanent` flag.
- **engine**: the shared case-evaluation seam re-raises a collected row's original failure (DRACO + HealthBench); judge sources gain bounded `retry=2` (url4 never retries permanent errors); `on_error` is stated explicitly; judge pins `reasoning_effort=low` + `max_tokens=8192`. Judge params are hashed into board identity, so the canonical DRACO revision moves deliberately: `66a463248586b277` → `fe291f4cdc670208`.
- **aigateway**: validated `reasoning_effort` for the OpenRouter provider (direct keyed rule, proven low/medium/high ladder, litellm transform tripwire, cache key-difference proof).
- **client e2e**: two authored failure tapes (`judge_token_cap`, `judge_429`) rehearse the incident against a real engine + FakeGateway and pin the post-fix landings.

## Cross-service contract / deploy order

The engine now sends `reasoning_effort` on judge calls and the gateway validates it. **Deploy aigateway before the engine** — an old gateway fails the unknown field closed and every judge call would error.

## Owner follow-ups (deliberately not in this PR)
- Re-seed production scoreboard benchmarks at the new revision.
- Re-record judge cache seeds and re-bless goldens (paid runs; candidate cache replays free).
- Correlate the 4 run_ids in GH #740 against engine span logs to name which trigger fired that day.

## Test plan
- `run_gates.py` ALL GATES GREEN: url4 (1166), screamingface-engine (2041), aigateway (3929+), screamingface (1127+), scoreboard (358).
- Gated e2e failure lane run locally (docker + `SCREAMINGFACE_TEST_E2E=1`): 43 passed — both new scenarios land as `stage=grading` with the original cause, candidate answer preserved.

Refs: OME-993


IonesioJunior (approved):
LGTM!

PR #750, opened 2026-08-26
feat(engine): fair-share concurrent run admission (OME-908)

IonesioJunior:
## Summary

Fair scheduling of concurrent Engine runs so one large benchmark run cannot starve the others (OME-908), layered as the spec recommends:

- **Layer 1 (local mode, dynamic)** — one shared `FairShareGate` per process (`Settings.local_io_capacity`, default 32): a work-conserving fewest-in-flight scheduler over per-run FIFO queues. Each run's executor wraps its world io in `FairShareIOLayer`, which claims one permit per `fetch`/`fetch_ex`/`fetch_holdings`. A solo run still gets the whole capacity; two contending runs converge to near-equal shares; a finished/stopped run's share reverts immediately.
- **Layer 1 (deployed mode, static)** — a per-run in-flight budget written onto the Job as `URL4_CLOUD_IO_CONCURRENCY` and enforced by URL4's own `BoundedIOLayer`; the `url4_run` kwarg is tri-state (omitted / `N` / explicit `None` when the gate is bound).
- Spec, plan, task mirror, and work ledger under `docs/`; diagram under `docs/diagrams/`.

## Review response (commit 4e940701)

Both blocking findings from the CHANGES_REQUESTED review were reproduced deterministically, then fixed:

1. **Grant/cancel race permit leak (P1)** — a waiter cancelled after `_pump` granted into its future unwound into `CancelledError` without returning the permit (asyncio defers the cancellation via `_must_cancel`; a done future refuses `cancel()`). `_abandon` now returns the permit for a granted-but-never-owned future. Regression test: `test_a_waiter_cancelled_after_grant_returns_the_permit`.
2. **Fixed arrival tie-break starvation (P1)** — a run whose wait queue never emptied (benchmark backlog) kept the oldest tie stamp and won every tie at equal holdings. The tie line now rotates on every grant, making capacity 1 strict round-robin under continuous demand (plan invariant 6). Regression test: `test_unbroken_demand_from_more_runs_than_capacity_cannot_starve`.

Both tests fail on the pre-fix code. PR title/body updated from the stale documentation-only text.

## Test plan

- `tests/unit/test_fair_share_gate.py` — the gate contract (26 tests incl. both new regression tests)
- `tests/unit/test_runner_io_concurrency.py` — env contract, executor call shape, local wiring, two-run interleave end to end
- Full engine unit suite: **2061 passed, 5 skipped**

Refs: OME-908


IonesioJunior:
Post-review documentation alignment (772a6710): spec/plan amended to the implemented design (unconditional Job env write, tri-state `url4_run` kwarg, fewest-in-flight naming, executable rollback recipe); D0–D5 resolutions and per-criterion acceptance status recorded; task mirror reflects the session approval; work ledger enumerates the six deviations; companion gateway ticket text drafted in the ledger (filing pending the owner); chart runner block gains the Layer-3 co-scheduling note. Gate suite ALL GREEN.

Refs: OME-908

HupBaHa (changes requested):
## Blocking findings

1. **[P1] A waiter cancelled after grant leaks the permit** (`apps/screamingface-engine/src/screamingface_engine/runner/fair_share.py:209-218`). The stated asyncio assumption is incorrect. If `_pump()` completes the future and the owning task is cancelled before it resumes, `Task.cancel()` cannot cancel the done future but still schedules cancellation into the task. `await fut` raises `CancelledError`, `granted` remains false, and `_abandon()` cannot find the already-popped future, so the `_active` and `_in_flight` increments are never reversed. Each occurrence permanently reduces local capacity and enough stopped/cancelled runs deadlock all later local fetches. I reproduced this on Python 3.12 and 3.13 with capacity 1: acquire `holder`, queue `waiter`, synchronously release `holder`, cancel `waiter` before yielding; the waiter exits cancelled while the snapshot remains `in_flight=1` for `waiter`. Please account for a future that was granted but whose task did not take ownership, and add this exact deterministic race test.

2. **[P1] Fixed arrival tie-breaking can starve later runs when demanders outnumber capacity** (`apps/screamingface-engine/src/screamingface_engine/runner/fair_share.py:192-207`). With capacity 1, run A holding a permit, and both A and B already queued, releasing A makes both active counts zero. `_arrival` always favors A, so A is immediately re-granted and B remains queued; repeated A demand can starve B indefinitely. The same issue generalizes whenever active runs exceed capacity. This contradicts the core no-starvation promise and the plan requirement that capacity 1 becomes strict round-robin; the existing capacity-1 test never queues the A successor before releasing A, so it misses the real refill shape. I reproduced `a_granted=True`, `b_granted=False`, with B still waiting after the release. Please use a rotating tie-break/fair queue among equal holdings and cover continuous demand with `N > capacity`.

## Additional review notes

- PR source changes are confined to `apps/screamingface-engine`; there are no direct `apps/aigateway` changes. The deployed per-run bound does affect AIGateway operationally by limiting arrivals into its provider FIFO, but I found no separate contract or credential regression in that plumbing.
- The PR title/body still describe documentation-only work even though this is a 26-file application-code change. Please update the metadata before merge.

Verification: the 24 focused fair-share/concurrency tests pass, `git diff --check` passes, and all visible GitHub checks are green. The two deterministic runtime probes above demonstrate missing scheduler coverage rather than existing test failures.

IonesioJunior:
Review response (4e940701): both blocking findings reproduced deterministically, then fixed.

1. **Grant/cancel race (P1) — confirmed.** Exactly your trace: `Task.cancel()` on a task parked on a done future defers the cancellation (`_must_cancel`), the resume throws `CancelledError` at `await fut`, `granted` stays false, `_abandon` finds nothing to remove. Reproduced with capacity 1: `in_flight=1` booked to the cancelled run, fresh waiter deadlocked. Fix: `_abandon` now recognizes a granted future (done, not cancelled, no exception) and returns the permit through `release()` (guarded on `_closed`). The added regression test is your exact recipe and fails on the pre-fix code.

2. **Arrival tie-break starvation (P1) — confirmed, mechanism refined.** `_arrival` stamps are (re)set when a run's wait queue is *created*, so a run whose queue never empties — the benchmark-backlog shape — keeps the oldest stamp and wins every tie at equal holdings (10/10 re-grants to A, B never served). A demand gap re-creates the queue with a fresh stamp, which is why the existing capacity-1 test passed. Fix: the tie line now rotates on every grant (the granted run's stamp is bumped), so no run wins two consecutive ties while an equal-holding competitor waits — capacity 1 under unbroken demand is strict round-robin, plan invariant 6. Added the demanded N>capacity continuous-demand test (3 runs, capacity 1, replenished demand); it fails on the pre-fix code.

Also: the impossible-race docstrings corrected, plan/spec amended to the rotating tie-break and grant-unwind permit return, PR title/body rewritten from the stale documentation-only text. Full unit suite green (2061 passed, 5 skipped); both CI matrix jobs (3.12/3.13) green.

Refs: OME-908

apps/screamingface-engine/tests/unit/test_fair_share_gate.py
```diff
@@ -0,0 +1,494 @@
...
+    await _settle()
+
+    snapshot = gate.snapshot()
+    assert _share(snapshot, "run-b").in_flight == 2  # the freed permits served B
+    assert snapshot.in_flight == 2
+    await waiter.finish()
+    assert gate.snapshot().runs == ()
+
+
+# --- the IOLayer wrapper -------------------------------------------------------------------------
```
HupBaHa: [P2] Split this new test module at this responsibility boundary. The two required regressions are correct, but their addition takes this hand-maintained new file to 494 physical lines, over the 450-line limit. Moving the `FairShareIOLayer` wrapper/capability tests from this heading into a focused `test_fair_share_io_layer.py` keeps the scheduler fairness/cancellation evidence together, preserves every existing test, and avoids an artificial line-count-only split because the responsibilities are already distinct.

HupBaHa (changes requested):
Both original P1 findings are fixed on this head. The cancellation-after-grant path now returns the booked permit before pumping the next waiter, and the rotating least-recently-served stamp removes the fixed-arrival starvation mechanism. The exact regression tests pass on Python 3.12.9 and 3.13.11, an independent falsification pass found no material race in the delta, and the full Python 3.13 suite passes (2078 passed, 5 skipped). Lock check, Ruff lint/format, Pyright, layering, Helm lint, git diff --check, and all visible CI checks are green.

One structural change remains before approval: the newly added fair-share test module is now 494 physical lines, over the 450-line boundary. The existing responsibility heading gives a minimal split point without changing behavior or tests.

HupBaHa (approved):
The two original P1 correctness findings are fixed and verified on this unchanged head. The test-module size note is non-blocking; the inline comment remains as an optional cleanup suggestion.

PR #752, opened 2026-08-26
feat(aigateway): weekly response-cache snapshot to Garage (OME-1021)

IonesioJunior:
## OME-1021 — Snapshot the aigateway response cache to Garage every Friday 05:00 UTC

Weekly archive of `request_cache_entries` in the existing OME-952 snapshot format (gzip'd COPY dump + manifest sidecar), PUT to a Garage bundled in the aigateway chart under a dedicated bucket/prefix (`screamingface-cache-snapshots` / `cache-snapshots/<stamp>`).

**Locked decisions** (owner-confirmed): bundled Garage with external-endpoint override · strictly next-Friday 05:00 UTC (no catch-up) · keep-all retention (bucket lifecycle rule) · schedule-only, no admin trigger.

### What lands

- **Exporter** (`core/request_cache/snapshot_export.py`): streaming `COPY (SELECT …) TO STDOUT` on a dedicated asyncpg connection → gzip off-loop through a hashing tee → exact loader contract (canonical header, verbatim rows, `\.`) + honest `SnapshotManifest` (`row_count` from the COPY status, sha256 of the gzip file, `revisions` from `active_cache_revisions()`).
- **SigV4 + store** (`core/sigv4.py`, `core/object_store.py`): port of the Engine's proven slice, PUT-only, streaming body with Content-Length + full-payload hash, sanitized errors.
- **Scheduler** (`core/snapshot_scheduler.py`): in-process weekly loop, single-flight skip, jittered exponential backoff, owned task (started/stopped by the lifespan).
- **Settings + wiring**: `AIGW_CACHE_SNAPSHOT_*` (opt-in, fail-fast when enabled without storage); `_lifespan` arms/stops the scheduler; containment proven — a cache hit completes while an export is in flight.
- **Chart**: bundled self-configuring Garage StatefulSet, stable chart-owned key-pair Secret, snapshot env keys, Deployment envFrom.

## Tests

- P0 round-trip through the **existing** loader (byte-identical `response_json`) + admin-runner acceptance against real Postgres; SigV4 vs AWS published vectors; schedule boundary math; single-flight; settings fail-fast; lifespan containment.
- MVCC: a mid-COPY committed insert is invisible (one point-in-time view).

## Verification

- `uv run .claude/scripts/run_gates.py aigateway --base origin/main` — ALL GATES GREEN
- `helm lint` + `helm template` (disabled / bundled garage / external endpoint / values-prod) + `verify_chart_wiring.py` 31/31

Refs: OME-1021


apps/aigateway/charts/aigateway/templates/garage.yaml
```diff
@@ -0,0 +1,159 @@
...
+  labels:
+    {{- include "aigateway.labels" . | nindent 4 }}
+spec:
+  # INVARIANT: 1. Two Garage processes on one metadata store is corruption, not contention.
+  replicas: {{ .Values.snapshot.garage.replicaCount }}
+  serviceName: {{ $name }}
+  selector:
+    matchLabels:
+      {{- include "aigateway.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: garage
```
HupBaHa, line 79: [P1] Keep Garage out of the gateway NetworkPolicy selector. This Pod receives the same `app.kubernetes.io/name` and `instance` labels selected by `templates/networkpolicy.yaml`; with the chart's default `networkPolicy.enabled=true`, that policy therefore applies to Garage too, but its only ingress rule permits TCP 9105. The gateway's PUTs to this Pod on 3900 are denied, so the standard `snapshot.enabled=true` + bundled Garage deployment can never publish a snapshot. Give gateway and Garage distinct component selectors (and select the gateway component in the existing Service/NetworkPolicy), or add an explicit correctly-scoped Garage policy.

apps/aigateway/src/aigateway/core/object_store.py
```diff
@@ -0,0 +1,130 @@
...
+        headers = self._signed_headers(key, sha256_hex=sha256_hex)
+        headers["Content-Length"] = str(size)
+        url = f"{self._config.endpoint_url.rstrip('/')}/{self._config.bucket}/{key}"
+        body = _FileStream(path)
+        try:
+            async with self._client_factory() as client:
+                response = await client.put(url, headers=headers, content=body)
+        except httpx.HTTPError as exc:
+            raise S3StorageError(f"PUT {key} could not reach object storage: {exc}") from exc
+        if response.status_code >= 400:
```
HupBaHa: [P1] Treat every non-2xx response as a failed PUT. HTTPX 0.28 defaults to `follow_redirects=False`, while this check accepts 3xx; a 301/307/308 therefore returns from `put()`, the caller deletes the local spool, and the scheduler logs `published` even though the target did not store the object. I reproduced this with `MockTransport` returning 307. Redirects cannot simply reuse this Authorization header because a changed host/path must be re-signed, so reject them unless explicit re-signing is implemented.

apps/aigateway/src/aigateway/main.py
```diff
@@ -156,6 +162,58 @@ def _factory() -> httpx.AsyncClient:
...
+                secret_key=secret_key.get_secret_value(),
+                region=settings.cache_snapshot_s3_region,
+            ),
+            timeout_s=settings.cache_snapshot_timeout_s,
+        )
+    )
+
+    async def _run() -> None:
+        export = await exporter.export()
+        prefix = f"cache-snapshots/{export.stamp}"
```
HupBaHa: [P1] Serialize publication across gateway replicas before using this shared prefix. Every replica starts its own scheduler, and exports started in the same second receive the same stamp. Two valid replicas can then interleave A-archive, B-archive, B-manifest, A-manifest, leaving B's bytes paired with A's checksum/row count. `replicaCount` is configurable and there is no startup guard or distributed lock. Hold a Postgres advisory lock for export plus both PUTs (or otherwise guarantee one writer) so a valid multi-replica deployment cannot corrupt the backup pair.

apps/aigateway/src/aigateway/config.py
```diff
@@ -150,6 +150,45 @@ class Settings(BaseSettings):
...
+    )
+    cache_snapshot_s3_secret_key: SecretStr | None = Field(
+        default=None, validation_alias="AIGW_CACHE_SNAPSHOT_S3_SECRET_KEY"
+    )
+    cache_snapshot_timeout_s: float = Field(
+        default=600.0, gt=0, allow_inf_nan=False, validation_alias="AIGW_CACHE_SNAPSHOT_TIMEOUT_S"
+    )
+    # Spool cap for one archive — protects the pod temp directory shared with the database.
+    cache_snapshot_max_bytes: int = Field(
+        default=512 * 1024 * 1024, gt=0, validation_alias="AIGW_CACHE_SNAPSHOT_MAX_BYTES"
```
HupBaHa: [P1] Keep successful exports within the default restore contract. This permits up to 512 MiB of COPY data, but `cache_upload_max_bytes` defaults to 256 MiB and `CacheUploadRunner` rejects larger compressed archives before loading them. A 300 MiB gzip can therefore be published successfully under all defaults and then be unrestorable through OME-952. Align the limits and validate the final archive size against the restore cap (the current counter measures uncompressed COPY bytes, so matching numeric defaults alone is not a complete invariant).

apps/aigateway/charts/aigateway/templates/snapshot-secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+{{- $secretKey := .Values.snapshot.storage.secretKey -}}
+{{- if and $existing $existing.data -}}
+{{- if and (not $accessKey) (index $existing.data "AIGW_CACHE_SNAPSHOT_S3_ACCESS_KEY") -}}
+{{- $accessKey = (index $existing.data "AIGW_CACHE_SNAPSHOT_S3_ACCESS_KEY" | b64dec) -}}
+{{- end -}}
+{{- if and (not $secretKey) (index $existing.data "AIGW_CACHE_SNAPSHOT_S3_SECRET_KEY") -}}
+{{- $secretKey = (index $existing.data "AIGW_CACHE_SNAPSHOT_S3_SECRET_KEY" | b64dec) -}}
+{{- end -}}
+{{- end -}}
+{{- if not $accessKey -}}
```
HupBaHa, line 37: [P2] Fail rendering instead of generating credentials for an external store. When `snapshot.garage.enabled=false`, `snapshot.enabled=true`, and only `endpointUrl` is supplied, these branches mint a random key pair that the external S3 service cannot know. The app then passes startup validation and discovers the permanent auth failure only on Friday. Restrict generation to bundled Garage; external mode should require both values or `existingSecret`.

apps/aigateway/src/aigateway/config.py
```diff
@@ -307,3 +346,41 @@ def _empty_secret_key_is_unset(cls, value: object) -> object:
...
+        that silently does nothing every Friday, and an unknown cron must never become no
+        schedule at all.
+        """
+        cron = self.cache_snapshot_cron.strip()
+        if cron != "0 5 * * 5":
+            raise ValueError(
+                "AIGW_CACHE_SNAPSHOT_CRON must be '0 5 * * 5' (every Friday 05:00 UTC); "
+                f"got {cron!r}"
+            )
+        if not self.cache_snapshot_enabled:
```
HupBaHa, line 368: [P2] Fail fast when snapshots are enabled on a non-Postgres database. The exporter supports only PostgreSQL COPY, but this validator accepts the application's default SQLite URL as long as S3 fields are present. Such a deployment boots as `scheduler armed`, waits a week, then retries the same deterministic unsupported-database error three times. Validate the database scheme here or in scheduler construction so an impossible schedule cannot look armed.

apps/aigateway/src/aigateway/core/object_store.py
```diff
@@ -0,0 +1,130 @@
...
+        """Upload `path` as object `key` under a full-payload signature.
+
+        ``sha256_hex`` is the sha256 of the file's bytes — the exporter computed it while
+        writing, so signing costs nothing extra. The PUT either completes or raises
+        :class:`S3StorageError`; no partial object is ever reported as stored.
+        """
+        size = path.stat().st_size
+        headers = self._signed_headers(key, sha256_hex=sha256_hex)
+        headers["Content-Length"] = str(size)
+        url = f"{self._config.endpoint_url.rstrip('/')}/{self._config.bucket}/{key}"
```
HupBaHa, line 130: [P2] Sign the same path/query that HTTPX transmits, or reject non-origin endpoints. For `endpoint_url=http://host/base`, this URL sends `/base/<bucket>/<key>`, while `_url_path()` signs only `/<bucket>/<key>` and `_signed_headers()` always signs an empty query, producing `SignatureDoesNotMatch`. The external endpoint override currently accepts these values without validation. Either derive the canonical target from the final `httpx.URL`, or fail fast unless the endpoint has path `/` and no query.

apps/aigateway/src/aigateway/main.py
```diff
@@ -156,6 +162,58 @@ def _factory() -> httpx.AsyncClient:
     app.state.codex_http_factory = _factory
 
 
+def _build_snapshot_scheduler(settings: Settings) -> CacheSnapshotScheduler:
```
HupBaHa: [P2] Move this independent snapshot composition into a cohesive module instead of growing the app composition root further. This PR takes `main.py` from 417 to 489 lines, crossing the 450-line review boundary with 72 lines of exporter/S3/scheduler construction. Keeping the builder beside `snapshot_scheduler.py` (and leaving lifespan start/stop ownership here) preserves the wiring while restoring a focused, reviewable main module.

HupBaHa (changes requested):
Requesting changes for four correctness blockers: the bundled Garage is unreachable under the chart's default NetworkPolicy, HTTP redirects are reported as successful uploads, multiple gateway replicas can overwrite one archive/manifest pair, and a default-valid export can exceed the default restore limit. I also found fail-fast gaps for external credentials and SQLite, endpoint canonicalization drift, and the snapshot wiring pushes main.py past the agreed size boundary.

Verification on exact head feacc6be: 35 focused tests passed; 3 real-Postgres snapshot integration tests passed; Ruff, Pyright, no-enterprise, Helm lint, and Helm rendering passed. The full non-live suite completed with 3963 passed and one unrelated timing-threshold failure that passed when rerun alone. GitHub CI is green, but the PR currently conflicts with origin/main in apps/aigateway/src/aigateway/main.py.

HupBaHa:
Review correction/refinement: the original multi-replica archive/manifest collision finding was overstated as a P1 relative to the approved OME-1021 v1 contract, which explicitly says `single replica, logged invariant`. However, the chart still exposes configurable `replicaCount` and does not enforce that invariant when snapshots are enabled, so the risk should remain as a P2 configuration/invariant gap rather than disappear. The minimal v1 fix is to fail Helm rendering or startup when snapshots are enabled with more than one AIGateway replica; alternatively, add cross-replica serialization such as a PostgreSQL advisory lock around export plus both PUTs. It is not an active collision in the declared one-replica deployment, but an unenforced invariant is worth fixing before operators can accidentally enter that state. The other CHANGES_REQUESTED blockers remain independent: bundled-Garage NetworkPolicy failure, false-success HTTP redirects, and export/restore size-contract mismatch.

IonesioJunior:
Thank you for an exceptional review — every failure chain checked out against the code, and the self-correction on the replica finding against the approved contract was exactly the right call. All findings are addressed in two commits: **`c3eb0a67`** (the four blockers) and **`50746798`** (the P2 gaps). Detail per finding:

## The four blockers

**1. NetworkPolicy selecting the bundled Garage (`garage.yaml:79`)**
Gateway Pods now carry `app.kubernetes.io/component: gateway` (Deployment selector + pod labels — the same pattern the migrate Job already used), and both the gateway NetworkPolicy and the Service selector are tightened to it. The Service was previously only *accidentally* safe via its named `targetPort: http`, which Garage's ports don't carry — that's now design, not luck. A new Garage-scoped NetworkPolicy in `garage.yaml` (renders with `snapshot.enabled && garage.enabled && networkPolicy.enabled`) admits exactly this release's gateway Pods on 3900, ingress-only. `verify_chart_wiring.py` gained 11 checks pinning the label/policy pair, including "the peer Garage admits IS the label set the Deployment renders".

**2. Redirects accepted as success (`object_store.py:96`)**
`put()` now rejects **every non-2xx**. Redirects get a dedicated error naming the `Location` target and `AIGW_CACHE_SNAPSHOT_S3_ENDPOINT_URL`, with your re-signing rationale inlined: the Authorization is signed for this host and path, so redirects are rejected, never followed. Tests: 301/307/308 assert failure, actionable message, no credential material, and exactly one request (never followed, never re-signed).

**3. Multi-replica stamp collision (`main.py:198`, per your correction comment)**
Handled as the configuration/invariant gap, not the advisory lock: the spec's single-replica assumption is now **enforced** — `aigateway.validateSnapshot` (`_helpers.tpl`, included from `configmap.yaml` beside `validateAuth`) fails the render for `snapshot.enabled=true` + `replicaCount>1`, naming the stamp collision and pointing at the advisory-lock/CronJob future work. The chart is the only place that knows `replicaCount`; multi-replica with snapshots off still renders (pinned in the verify script).

**4. Export cap vs restore cap (`config.py:189`)**
`cache_snapshot_max_bytes` default lowered 512→**256 MiB** (= `cache_upload_max_bytes`), plus a Settings validator refusing `snapshot_max > upload_max` when enabled ("raise both together or lower the export cap"). One correction to the review's parenthetical, which changed the fix's shape: the export counter measures the **compressed archive bytes**, not uncompressed COPY bytes — `_HashingSink` sits *beneath* the `GzipFile` wrapper and `_write()` flushes before checking `sink.total`. Both caps therefore count the same artifact, so equal numeric defaults *are* a complete invariant; the validator is the belt-and-braces. Spec §4.1/§4.4 updated to match.

## The P2 gaps

**5. Minted credentials for an external store (`snapshot-secret.yaml:25`)**
Generation is now bundled-Garage-only: external mode (`garage.enabled=false`, no `existingSecret`) without **both** keys fails the render, stating why — an external S3 cannot adopt a chart-minted pair, so the first signal would otherwise be a 403 a week late. The `lookup` upgrade-stability for the bundled case is preserved. Verify script: refusal pinned, operator pair rendered verbatim, `GK…` generation still works for bundled.

**6. SQLite passing validation (`config.py:365`)**
`_validate_cache_snapshot` now requires a `postgres://` DSN when enabled (checked after storage presence so the more specific error wins). The DSN itself is never echoed — it carries credentials; a test pins that. The wiring tests were adapted through a documented seam: Settings carry a postgres-form DSN that is never dialed (every test replaces `_run` before any fire) while the ORM runs on sqlite via `init_db`.

**7. Signed path ≠ transmitted path (`object_store.py:89`)**
Took your second option: `S3ObjectStoreConfig.__post_init__` fails fast at construction on base path / query / fragment / userinfo / missing host / non-http(s), each message naming the env var. Trailing-slash origins are accepted and a test pins the PUT still going to `/<bucket>/<key>`. Deriving the canonical target from the final `httpx.URL` was the fragile alternative; if base-path endpoints ever become a real requirement, that is the module's own stated signal to take a real S3 client.

**8. `main.py` size (`main.py:165`)**
The builder and publish protocol moved **verbatim** to `core/snapshot_publish.py::build_snapshot_scheduler` — `main.py` 495→**441** lines, back inside the ≤450 boundary. The lifespan keeps only start/stop ownership; the new module documents the publish protocol (archive before manifest, `finally` rmtree, `published` reachable only after both PUTs returned 2xx). Spec §4.5 updated.

## Gates

Per commit: focused tests 41 / 44 passed; full non-live suite **4094 / 4103 passed** (0 failures); ruff + format, pyright 0 errors, `check_no_enterprise`, helm lint, prod-values render, and `verify_chart_wiring.py` **42/42 → 47/47** (5 new refusal/generation checks). CI on `50746798`: 17 passed, 0 failed, 2 path-filtered skips at the time of writing, with the remaining matrix lanes finishing. Both rounds are documented in the work ledger (`docs/work/aigw/2026-08-26-OME-1021-cache-snapshot-schedule.md`).

One deliberate non-change to flag: `garage.replicaCount > 1` is still only comment-documented (pre-existing, outside this review's scope) — happy to add a render guard for it too if you'd like it in the same invariant family.

apps/aigateway/charts/aigateway/templates/deployment.yaml
```diff
@@ -9,10 +9,17 @@ spec:
   selector:
     matchLabels:
       {{- include "aigateway.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: gateway
```
HupBaHa: [P1] Preserve the existing Deployment selector. This chart already ships an `apps/v1` Deployment selected by `app.kubernetes.io/name` + `app.kubernetes.io/instance`. Adding `app.kubernetes.io/component: gateway` to `spec.selector` changes that selector, but Kubernetes makes a Deployment selector immutable after creation, so `helm upgrade` of an existing release fails with `field is immutable`. Keep `component: gateway` on `template.metadata.labels` so the Service and NetworkPolicy can select gateway Pods, but remove it from `spec.selector`. Fresh installs and render-only CI do not expose this upgrade failure. Official contract: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

HupBaHa (changes requested):
One merge-blocking upgrade regression remains at the current head. The earlier review findings are otherwise resolved; the redirect-target sanitization and Garage replica hardening observations are non-blocking follow-ups and are not part of this request for changes.

HupBaHa (approved):

PR #762, opened 2026-08-27
feat(screamingface-engine): retain operation accounting

keelancj:
Refs: OME-1030
Parent: OME-901
Design and lineage: #763 (merged)

## Summary

Retains the exact Gateway accounting already present during an evaluation on the semantic records that explain the work: Candidate `CaseOperation`s and rubric-grading `Evidence`.

This is the Engine producer for the completed-Report cost breakdown. It does not change scoring or add model calls. The strict Client decoder is stacked in #764 so both halves land atomically.

## Before / After

**Before**

- the final Report retained one authoritative total for the whole Candidate run
- Candidate operations kept outputs but not their cache outcomes, tokens, cost, model identity, provider latency, or provider-attempt count
- rubric Evidence kept the verdict but not the accounting for the judge calls that produced it
- generation versus grading, member versus synthesizer, model, and uniquely attributable Case spend could not be explained from the completed Report

**After**

- uniquely attributable Model/Fusion/Pipeline operations retain exact current-run accounting
- DRACO, HealthBench, and GDPval grading Evidence retain the exact accounting for their authored judge requests; deterministic IFEval Evidence remains explicitly null
- provider latency is paired with the exact validated attempt count, so a slow model can be distinguished from a retry-heavy route
- ambiguous work and nested CorrectiveLoop work remain unattributed rather than guessed
- the existing root total remains authoritative and unchanged

## How it works

Two deliberately different capture scopes share the connector's existing accounting facts:

1. **Candidate-local operation capture** retains the model output already required to attribute member and synthesis operations.
2. **Run-level grading capture** retains only an in-memory request key and accounting. It never retains prompts or model outputs.

A generic composition-root `Executor` decorator owns the run scope around the unchanged `Url4Executor`. Each rubric benchmark registers the exact judge request it authored and joins that request to its own Evidence only when ownership is unique.

The retained value is strict and exact-only: provider/request/response model identity, input/output/cache-read/cache-creation/reasoning tokens, fixed-point USD cost, summed provider latency, exact validated provider-attempt count, and cache hit/miss/bypass/unknown counts. Missing or inconsistent facts remain null. The attempt count reuses the Gateway attempt ledger the Engine already consumes; it adds no timer, retry mechanism, or transport.

## Review hardening

- provider-refused terminal calls retain accounting before the refusal leaves the connector
- provider and served-model identity are retained only when every contributing Gateway attempt agrees; usage still reflects all contributing attempts
- malformed accounting metadata is fail-open: the successful model result survives and accounting becomes null
- fixed-point cost sums are independent of ambient `Decimal` precision
- an unavailable tool round poisons the whole operation rather than publishing a false subtotal
- late duplicate grading ownership is reconciled retroactively to null
- repeated grading keys are indexed rather than rescanned per verdict
- Candidate payload capture is isolated from the payload-free run ledger
- solo nested Recipes do not create meaningless empty operations, while multi-operation Candidates preserve explicit null entries when attribution was attempted but unavailable
- live/root Usage keeps its existing terminal-attempt identity; unanimous identity applies only to the new retained breakdown

Diagnostics name only the accounting phase and exception type. They never interpolate prompts, URL4, parameters, outputs, request keys, Gateway bodies, call ids, or attempt ids.

## Integrity boundaries

- no `packages/url4`, AI Gateway, URL4 expression, score, retry, cache-key, model-call, or live Event changes
- no new timer, retry mechanism, serialized request identity, or failed-root partial Report
- partial, omitted, malformed, duplicate-owner, or otherwise ambiguous accounting becomes null rather than a guess
- Candidate calls stay isolated from the grading ledger; all scopes unwind on cancellation and early close
- CorrectiveLoop nested detail and failed-path accounting remain explicitly deferred

## Rollout order

This directly evolves the pre-1.0 Candidate Invocation/Result v1 contract by adding required-null `accounting` fields to `CaseOperation` and `Evidence`.

#764 targets this branch and supplies the strict Client decoder. Merge #764 into this branch, let the combined checks rerun green, then merge this PR into `main`. Neither half lands alone; no legacy compatibility fallback is added.

## Test plan

- complete/partial/omitted/malformed Gateway accounting, cache hits, retries/tool rounds, identity disagreement, refusal, and strict sums
- exact provider-attempt counts for complete responses and cache hits; null on partial/malformed/absent attempt evidence
- unique and ambiguous Candidate projection, solo Model, nested Recipe, CorrectiveLoop non-attribution, and strict old/new wire shapes
- exact, redrawn, and colliding grading ownership plus DRACO, HealthBench, GDPval, and IFEval vertical slices
- payload-free run capture, Candidate isolation, concurrent runs, nested tasks, cancellation, and early iterator close
- unchanged root Usage, scores, Evidence meaning/raw output, URL4 rendering, cache keys, model-call cardinality, failures, and stream Events
- Engine gates green: Ruff check, Ruff format, Pyright, layering, and full pytest/coverage
- combined Client + Engine golden replays are green on #764 without re-blessing

## Screenshots

Not applicable — retained Engine contract and accounting projection only.


khoaguin:
Review note:

**Core idea**: Capture accounting at the single point where a call finishes, then join it back to the record that named the work.
pr762-ome1030-operation-accounting

`_chat_completion_loop` normalizes the Gateway `_aigw` envelope (fail-open), combines tool rounds, and calls `record_operation_call` — one publication point that fans out to two ledgers with two different join keys

Screenshot 2026-08-28 at 10 20 58

**Manual review order**

1. Vocabulary — `operation_accounting.py` (new), `request_identity.py` (new, 20 lines). Learn the shape of a retained fact and the hash before anything uses them.
2. Producer — `runner/accounting.py` → `operation_calls.py` → `runner/connector.py`. This is where facts are born and forked.
3. The two joins — `operation_outputs.py` (Candidate), `grading_accounting.py` (grading), `invocation.py` (isolation switch).
4. Root + contract — `operation_capture.py`, `runner/main.py`, then `contract.py` (the breaking change).
5. Benchmark wiring (20 min) — DRACO closely (`prompts.py` + `exam.py` + `runtime.py`), skim GDPval/HealthBench/IFEval, then `aggregation.py`.
6. Tests (10 min) — failure boundaries, capture executor, grading collisions.

khoaguin (approved):

khoaguin:
Please review and merge https://github.com/ScreamingFace/screamingface/pull/764 into this first

PR #764, opened 2026-08-28
feat(py-screamingface): decode the Engine's retained per-operation accounting

khoaguin:
Unblocks #762. The Engine there emits per-operation `accounting`; this teaches the Client to decode it, so an evaluation completes instead of dying on `Case Operation contains unsupported field 'accounting'` and the combined golden-replay lane stays green.

> **Base is `OME-901-runtime-accounting-lineage`, not `main`.** Merge this INTO #762, which then lands both halves in one commit. Either side alone turns CI red in the mirror-image way.

### Before

- Trigger: a dev runs `sf.evaluate(...)` against an Engine that retains per-operation accounting
- Today: the Client's strict decoder refuses the reply, so the whole evaluation fails
- Cost: the Engine change cannot land alone

### After

- Same trigger
- Now: the Client decodes retained accounting on Candidate operations and grading Evidence and round-trips it through the public API and Report JSON
- Cost, token, cache, provider-time, and provider-attempt facts remain attached to the operation they explain
- Ambiguous or absent accounting stays explicitly null — never a guess or false zero

## Why the two halves land together

The Engine emits `"accounting"` unconditionally (`null` when empty), so it is a required-nullable field and the decoder is strict with no compatibility fallback:

| Merged alone | CI result |
|---|---|
| #762 (Engine only) | current Client rejects unsupported `accounting` |
| this branch into `main` | new Client rejects missing `accounting` from the old Engine |

The e2e harness boots both from the same checkout, and the combined tree is green.

## What's here

- focused public `OperationCache` and `OperationAccounting` values; the six usage fields reuse the existing `sf.Usage`
- focused strict accounting wire decoder, separate from the general result decoder
- `accounting` on `CaseOperation` and `Evidence`, emitted by `to_dict` unconditionally
- exact validated `provider_attempts` beside summed provider latency
- canonical fixed-point `cost_usd` enforcement: exponent notation, leading-zero/plus spellings, floats, and `Decimal` objects refuse instead of being silently normalized
- typed token/cache boundaries with no `# type: ignore` escapes
- deliberately out of scope: computed views, `MemberResult.usage`, and the completed-Report SFDS table (OME-1031)

## Test plan

- **29 accounting decoder tests** covering both owners, exact round-trip, null preservation, version mismatch, malformed fields, fixed-point cost spellings, cache boundaries, and provider-attempt count
- full Client gate green: Ruff check, Ruff format, Pyright, pytest with coverage ≥95%, deterministic notebooks, wheel build, and distribution verification
- combined Docker e2e lane: both golden replays green without re-blessing

## Reviewer notes

- Prior hand-authored old wire fixtures gained the required `"accounting": null` field; assertions were extended, not weakened.
- The public-surface snapshot was regenerated for the two new public values.
- Review extraction moved the new values and decoder into focused 97- and 129-line modules, removing the feature's material growth from the two already-large result modules.
- Scores, Evidence meaning, outputs, failures, URL4 rendering, cache keys, root usage, model-call cardinality, and live events remain unchanged.

Refs: OME-1032

### DIFF
--- a/docs/tasks/2026-08-28-OME-1032-decode-operation-accounting.md
+++ b/docs/tasks/2026-08-28-OME-1032-decode-operation-accounting.md
@@ -1,0 +1,28 @@
+---
+id: OME-1032
+linear_url: https://linear.app/openmined/issue/OME-1032/keep-each-operations-cost-and-token-facts-when-a-report-is-read-back
+status: in_progress
+type: feature
+priority: 2
+labels: [py-screamingface, agentic, autonomous]
+created: 2026-08-28
+closed:
+---
+
+# Keep each operation's cost and token facts when a report is read back
+
+Parent: `OME-1031` (which keeps the computed views, `MemberResult.usage`, and the
+completed-Report table). Grandparent: `OME-901`.
+
+The Engine (`OME-1030`, PR #762) retains `accounting` on Candidate `CaseOperation`
+and grading `Evidence`, emitted unconditionally and `null` when empty. The Client's
+strict decoder rejected the unknown field, so every evaluation against that Engine
+failed and #762's golden-replay lane was red. This slice decodes the retained value
+and round-trips it through the public API and Report JSON — nothing more.
+
+**Lands atomically with the Engine.** The field is required-nullable and the decoder
+has no compatibility fallback, so either side alone turns CI red in the mirror-image
+way. This branch is cut from `OME-901-runtime-accounting-lineage` and its PR targets
+that branch, not `main`.
+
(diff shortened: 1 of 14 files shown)


keelancj (approved):
@khoaguin added just a few small things, fixed the Client accepting/ normalizing non-fixed-point cost_usd, removed the new type: ignores by extracting the  accounting value and decoder into focused modules, and aligned the spec/Linear/PR descriptions with provider_attempts. No accounting behavior or benchmark logic was redesigned. Combined tests and golden replays are green and rest of review looks good, so I’m approving the merge!

PR #765, opened 2026-08-28
fix(py-screamingface): make the README quickstart runnable and correct the first-run docs

khoaguin:
Closes the gap an external tester found on 2026-08-27: the top-level README's Quickstart does not run, and `sf.__version__` — the first thing anyone types to check what they installed — raises. Fifteen fixes across the client README and the docs site, all from one tester report, all verified against `main` before being written up.

Linear: [OME-1034](https://linear.app/openmined/issue/OME-1034/fix-the-broken-readme-quickstart-and-the-docs-errors-a-first-time-user)

### Before
- Trigger: someone installs `screamingface`, opens the README, and copy-pastes the Quickstart.
- Today: **line 1 of the ensemble example raises `TypeError: missing 1 required keyword-only argument: 'synthesizer'`.** The README also promises an `sf.from_url4(...)` entry point that does not exist, and `sf.__version__` raises `AttributeError`. On the docs site the installation FAQ tells them to go kill a process when the CLI would have moved the port for them, and the Pipelines page opens with a diagram of a Fusion.
- Cost: the first thing a stranger runs fails, and the one command that would tell them what they installed fails too — so they cannot tell a bad install from stale docs from a broken product.

### After
- Same trigger.
- Now: the Quickstart runs top-to-bottom as written, every documented symbol exists, `sf.__version__` reports the installed version, the ports FAQ names the override flags, and the Pipelines diagram shows stages.
- Win: a first-run session ends in a score instead of a traceback.

### Don't regress
- `packages/screamingface/README.md` examples were already correct and are untouched — the root README was the sole code outlier.
- No `github.com/OpenMined` link is touched here; that is OME-914 / OME-945 on a sibling branch.
- Scores, evaluation behaviour, and the live widget are unchanged.

## Root cause

Three independent kinds of drift, none of them caught by CI because none of them is executable:

1. **The README examples are not tested.** `Fusion.__init__` takes `synthesizer` as keyword-only with no default (`fusion.py:20-26`). Every docs-site page and the package README pass it. The root README never did, and nothing runs it.
2. **`sf.from_url4` was documented but never built.** The real replay path is `sf.evaluate(url4_string)`; the editable-copy path is `sf.Url4(v).to_python()`, which returns a `str` of source rather than an object — also mis-described in the package README.
3. **The docs site drifted from the CLI.** `--gateway-port` / `--scoreboard-port` / `--engine-port` and the `SCREAMINGFACE_*_PORT` env vars have worked since `_runtime/config.py:23-35`, but the FAQ still told users to free the port by hand. Separately, the Pipelines page's lead diagram was a copy of the Fusions one — boxes, `aria-label` and `figcaption` all described a synthesizer combining members.

## What's here

**Client (`packages/screamingface`)**
- New `_version.py` exporting `sf.__version__`, read from `importlib.metadata` so it cannot drift from `pyproject.toml`. A bare source tree resolves to `0.0.0+source` rather than raising. Five tests, written RED-first.
- README: the Fusion example passes `synthesizer=`; the `from_url4` claim replaced with the two real entry points; `to_python()` correctly described as returning source text.

**Docs site (`public-docs`)**
- Ports FAQ rewritten around the override flags and env vars, including the real caveat that `down` and `logs` ignore them.
- `Note.vue` callout links get a visible affordance (`--accent-text-low` + underline) instead of rendering as plain body text.
- The Pipelines lead diagram is now three stages in series inside a labelled Pipeline frame, with `aria-label` and `figcaption` rewritten to match.
- Six prose corrections adopted verbatim from the tester, the LANL paper (*Beyond Leaderboards: Tokenomics of Agentic Small Language Model Ensembles*) linked where it was cited bare, and the benchmark feedback CTA given a destination.

## Test plan

- `run_gates.py screamingface`: **ALL GATES GREEN** — 1240 passed, 17 skipped, coverage 95.09% against a 95% floor.
- `public-docs`: `vue-tsc`, `vite build`, `oxlint`, `eslint` all exit 0.
- Acceptance: `python -c "import screamingface as sf; print(sf.__version__)"` → `0.1.1.post5`.

## Reviewer notes

- **One gate was skipped and it needs a call.** Adding a public export necessarily moves `tests/public_surface_snapshot.json`, and the append-only stage cannot parse JSON, so it rejects any change to that file. Regeneration used the documented `UPDATE_SURFACE_SNAPSHOT=1` path and the run used the runner's own `--skip-append-only` flag; no gate config was edited. Hand-audited and independently re-checked: the snapshot diff is **+5/−0**, all five lines being `__version__`, and no pre-existing test file or assertion was modified.
- **Two items need eyes, not CI.** The Pipelines diagram and the Note-callout link styling type-check and build but are not machine-verifiable: `npm run dev` → `/sf-client/guides/pipelines`, both themes.
- Item 13's source read `DRACO state-of-art` (uppercase), not `Draco`; the missing "the" was added and the existing casing left alone.
- Item 14 links the OpenReview **forum** URL (HTTP 200 verified); the PDF attachment URL 403s to non-browser clients.
- Unplanned but necessary: `InstallationPage.vue` asserted `len(sf.__all__)  # 55`, which went stale the moment the new export landed. Corrected to 56.

## Not fixed here

The tester also reported per-member `usage` and `duration_ms` always reading `None`. That is correct-as-designed on this branch — hard-coded null at `_evaluation/results.py:137-147` — and belongs to OME-699 and the OME-1030/1031/1032 stack. Worth knowing: OME-1031 keeps `duration_ms` null deliberately, because Gateway provider time is not wall duration, so that half of the report has no ticket that closes it.

## Follow-ups spotted, not actioned

- `_runtime/cli.py`'s `--version` still calls `importlib.metadata.version(...)` directly rather than the new `resolve_version()` — a second copy of the same lookup.
- Three `public-docs` files already fail `prettier --check` on `main` (`learn/LeaderboardPage.vue`, `sf-client/guides/LeaderboardsPage.vue`, `sf-client/guides/Url4Page.vue`). Pre-existing; Prettier is not a gate.

Refs: OME-1034

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,9 @@ ScreamingFace is a toolkit for composing **fusions**, several models answering t
 reduced to a single answer, and scoring them against real benchmarks. Concretely, it is three pieces:
 
 - **An open protocol, `url4`.** Every fusion and its benchmark run compile to one short,
-  human-readable line. Sharing a result is sharing that string, and `sf.from_url4(...)` reproduces
-  the exact run in a single call. **A result is only worth as much as your ability to rerun it.**
+  human-readable line. Sharing a result is sharing that string: `sf.evaluate(url4_string)` replays
+  the exact run, and `sf.Url4(url4_string).to_python()` hands you the same recipe as Python source
+  you can edit. **A result is only worth as much as your ability to rerun it.**
 - **A shared cache.** Every model call is cached and shared across the community, so reproducing a
   run is both **faithful and nearly free**. Verifying someone's work costs minutes, not budgets, and
   the more people run, the cheaper it gets for everyone.
@@ -60,7 +61,7 @@ sf.configure(engine_url="http://127.0.0.1:9108")   # your own engine, or a hoste
 
 gpt = sf.Model("openrouter/openai/gpt-5.5")
 flash = sf.Model("openrouter/google/gemini-3-flash-preview")
-fusion = sf.Fusion([gpt, flash])
+fusion = sf.Fusion([gpt, flash], synthesizer="openrouter/openai/gpt-5.5")
 
 # score the solo model beside the fusion, on the same cases
 report = sf.evaluate([gpt, fusion], benchmark="ifeval", limit=3)
--- a/docs/tasks/2026-08-28-OME-1034-readme-and-docs-fixes.md
+++ b/docs/tasks/2026-08-28-OME-1034-readme-and-docs-fixes.md
@@ -1,12 +1,12 @@
 ---
 id: OME-1034
 linear_url: https://linear.app/openmined/issue/OME-1034/fix-the-broken-readme-quickstart-and-the-docs-errors-a-first-time-user
-status: in_progress
+status: done
 type: task
 priority: P2
 labels: [py-screamingface, autonomous, agentic]
 created: 2026-08-28
-closed:
+closed: 2026-08-28
 ---
 
 # Fix the broken README quickstart and the docs errors a first-time user hits
--- a/docs/tasks/2026-08-28-OME-1034-readme-and-docs-fixes.md
+++ b/docs/tasks/2026-08-28-OME-1034-readme-and-docs-fixes.md
@@ -1,0 +1,41 @@
+---
+id: OME-1034
+linear_url: https://linear.app/openmined/issue/OME-1034/fix-the-broken-readme-quickstart-and-the-docs-errors-a-first-time-user
+status: in_progress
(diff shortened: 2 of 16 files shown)


keelancj (approved):
No blocking issues, the two open items (MemberResult.usage/duration_ms still null, cli.py --version not yet using the new resolve_version()) are already disclosed in the PR body as out-of-scope/follow-up and I agree with leaving both for later rather than expanding this PR's scope. Approved

PR #766, opened 2026-08-28
docs(repo): repoint the remaining OpenMined org references to ScreamingFace

khoaguin:
Finishes the org move: every `github.com/OpenMined/screamingface` link in docs, portal and public-docs now points at `ScreamingFace/screamingface`, and the README's "just open an issue" sentence becomes a real link. Closes the two halves of the same sweep.

Linear: [OME-914](https://linear.app/openmined/issue/OME-914/repoint-the-remaining-openminedscreamingface-references-in-docs-portal) · [OME-945](https://linear.app/openmined/issue/OME-945/finish-the-org-repoint-pypi-issue-urls-and-readme-issue-link)

### Before
- Trigger: a user hits a bug and looks for where to report it — or any reader clicks a source link in the docs.
- Today: PyPI metadata sends them to `ScreamingFace/screamingface/issues` while every doc, portal page and docs-site footer sends them to `OpenMined/screamingface`. Two different trackers for the same project. The README's invitation to "open an issue" is plain text with nothing to click.
- Cost: reporters split across two destinations, and the docs quietly teach the wrong org name to everyone reading them.

### After
- Same trigger.
- Now: every link resolves to the canonical org, and the README invitation is clickable.
- Win: one tracker, one org name, and a reader who wants to file something can actually get there.

### Don't regress
- **Release history stays honest.** No `CHANGELOG.md` is rewritten — those links were correct when written.
- **Sibling repos stay pointed at OpenMined.** `screamingface-brand`, `screamingface-benchmarks` and `screamingface-stack` are genuinely still OpenMined's, as is `OpenMined/sf-installer` in the release workflows.
- Prose that is *about* the old org keeps naming it (see below).

## Root cause

The org move landed in two passes and neither finished the job. `OME-910` deliberately fixed only what a GitHub redirect cannot cover — published PyPI metadata and the Trusted Publishing comments — leaving everything else to resolve via redirect. That was a reasonable call, but it left ~130 stale references, and the half that *was* fixed made the two halves disagree with each other.

Notably, **the divergence OME-945 describes was one-sided.** Both `pyproject.toml` files already declared `ScreamingFace/…` on `main`; the docs were the only stale side. OME-945 reduced to the README link alone.

## The trap in this change

The obvious pattern `github.com/OpenMined/screamingface` **prefix-matches three sibling repos** — `-brand` (9 places), `-benchmarks` (3, including a private repo referenced in `docs/PROJECT-OVERVIEW.md`), and `-stack`. A blanket find-and-replace would have silently repointed a dozen links at repos that do not exist under the new org, and nothing in CI would have caught it. The rewrite guards on the trailing hyphen; all 12 were verified intact afterwards.

## What's here

- 40 files, +244 / −43, across `README.md`, `CONTRIBUTING.md`, `.claude/README.md`, `docs/**`, `public-docs/src/**`, `apps/scoreboard/portal/*.html`, `apps/screamingface-studio/src-tauri/tauri.conf.json`, and the notebook build script.
- OME-945's README issue link.

## Test plan

- `run_gates.py screamingface` (8 gates) and `run_gates.py url4` (5 gates): green.
- `public-docs`: `vue-tsc`, `oxlint`, `eslint`, `vite build` all exit 0. Linters were run **without** `--fix` (the package scripts include it) so they would report rather than silently rewrite files owned by the sibling branch.
- Final sweep: 110 occurrences remain, every one deliberate — 103 in CHANGELOGs, 3 sibling-repo, 4 in this branch's own ledger and mirror quoting the string to document the rule.

## Reviewer notes

- **One judgment call worth a glance.** Two dated `2026-07-15` url4-PyPI artifacts (`docs/plan/…`, `docs/spec/…`) were rewritten despite being historical documents. The reference sits under *"Owner actions (post-merge, pre-publish)"* — a checklist `OME-910` records as **still open** — so it is a live instruction rather than release history, and leaving it stale would misdirect whoever executes it. Two-line revert if you read it the other way.
- **Deliberately left alone**, applying the tickets' rationale rather than just their letter: the `AIDEV-NOTE (OME-910)` blocks in `release-screamingface.yml` / `release-url4.yml` say an entry registered against `OpenMined/screamingface` no longer matches the OIDC claim — rewriting those makes them self-contradicting. Likewise `docs/work/2026-08-20-OME-910-repoint-org-refs.md`, which makes assertions *about* the old string.
- **Expect a rebase.** #765 (OME-1034) touches `README.md` and several `public-docs/src/pages/sf-client/` files on a sibling branch; a test-merge shows they conflict. That PR owns those files' content, this one owns only their URLs. This should land second.

## Still open after this

The PyPI Trusted Publisher entries **still name the old org in the PyPI console**. Repointing docs does not repoint PyPI — that stays an owner action, carried over from `OME-910`.

Refs: OME-914
Refs: OME-945


keelancj (approved):
Changed a couple things but other than this, approved
- Fixed six lowercase openmined/screamingface OCI/Helm source URLs.
- Repointed public-docs’ generic GitHub buttons to github.com/ScreamingFace, preserving their organization-level meaning.
- Replaced the case-sensitive audit with a case-insensitive one and updated the ledger
- Verified public-docs typecheck/build, oxlint, eslint, git diff --check, and the rendered diagram. Remaining OM references describe OpenMined Compute or historicalm decisions and are intentional

itstauq (approved):

PR #770, opened 2026-08-28
feat(screamingface): send the run cost on leaderboard submissions

FilipBolt:
Closes OME-1029.

The last missing link in the run-cost chain. Scoreboard has accepted `run_cost_usd` since OME-770, the Engine produces a run total, and the SDK already held it on `CandidateResult.usage.cost_usd`. The submission payload just omitted it.

The consequence is measurable: `GET /v1/leaderboard/draco` returns `entries=0, with a cost=0`, and `git grep -l run_cost_usd` over `packages/` and `apps/screamingface-engine` returned nothing.

## The change

One helper and one payload key in `_submission()`. Both the sync and async submit paths share it.

## Two decisions worth review

**A decimal string, never a float or a raw `Decimal`.** The payload is handed to `json=`, whose `json.dumps` raises `TypeError` on a `Decimal`; a float would silently lose precision on what Scoreboard stores as `DECIMAL(12, 6)`. `str()` is already this SDK's idiom for the same value — see `_report_primitives.Usage`.

**`None` stays `None`, never `0`.** Absent means "no cost was reported". `0` means "this run genuinely cost nothing", which a fully cache-served run legitimately does. OME-770's D10 and OME-923's frontier exclusion rule both depend on the distinction — a null read as zero would place an unpriced run at the cheapest end of the Pareto frontier, asserting something about money nobody measured.

Normalisation stays Scoreboard's. Quantization, sub-quantum rounding and sign-zero are not re-implemented client-side, or the two contracts would drift.

## Test plan

The absent/zero boundary is pinned first and separately, since that is the whole risk. 11 new tests; the file goes 47 → 58.

All eight gates green, including the **95% coverage floor**, notebook determinism, `uv build` and the distribution check.

Beyond the suite, every cost state round-trips through `json.dumps` and back to the identical `Decimal`: priced, cache-served `0`, unpriced `None`, sub-quantum `4E-7`, and the column ceiling `999999.999999`.

**Note on one test:** Step 4 was meant to validate against Scoreboard's real `ScoreSubmission`. That is not importable from this venv (`ModuleNotFoundError: pypika_tortoise`) because the SDK and Scoreboard are separate deployables that correctly do not share dependencies. Substituted the property that field relies on — what we emit parses back to exactly the `Decimal` we started with, parametrised across the column's range.

**No prior test modified.** An earlier attempt added a parameter to the shared `_candidate_result` fixture and the append-only gate caught it; the helper was rebuilt to construct from the fixture's own attributes rather than request a rule-5 exception. `dataclasses.replace` is not usable here — `CandidateResult` accepts `metrics=` but stores `_metric_items`. The diff against `main` in that file is purely additive.

## What this unblocks, and what it does not

**Unblocks** OME-822's real gate — "a client can produce a run total end to end" — which was never met despite the ticket leaving Blocked on 2026-08-19.

**Does not** populate existing rows. No backfill, and `content_hash` dedup excludes cost (OME-391, OME-770 D8), so a recipe already submitted without a cost cannot gain one later.

**OME-923 part B stays blocked** until real cost data is actually flowing, not merely until this merges — the Pareto marks need populated rows, and that needs a released client carrying this.

### DIFF
--- a/docs/plan/2026-08-28-OME-1029-sdk-run-cost.md
+++ b/docs/plan/2026-08-28-OME-1029-sdk-run-cost.md
@@ -1,0 +1,85 @@
+# OME-1029 — Implementation plan
+
+**Spec:** `docs/spec/2026-08-28-OME-1029-sdk-run-cost.md`
+· **Ledger:** `docs/work/2026-08-28-OME-1029-sdk-run-cost.md`
+· **Branch:** `OME-1029-sdk-run-cost` · **Stack:** screamingface
+
+Gates: `uv run .claude/scripts/run_gates.py screamingface --base origin/main`
+→ append-only · ruff check · ruff format · pyright · pytest --cov=screamingface **--cov-fail-under=95**
+· notebook determinism · `uv build` · distribution check
+
+Note the 95% floor and the notebook check — this stack is stricter than scoreboard's.
+
+## Ordering principle
+
+The absent/zero distinction is the whole risk. A null read as zero would place an unpriced run at
+the cheapest end of the Pareto frontier `OME-923` is about to build, so that boundary is pinned
+first and separately from the happy path.
+
+---
+
+### Step 1 — the three cost states
+
+**RED:**
+- a priced run puts a decimal **string** in the payload, not a float and not a `Decimal`;
+- an unpriced run (`cost_usd is None`) puts `None` — asserted as `is None`, never `== 0`;
+- a run costing exactly nothing puts `"0"`, and that is distinguishable from the unpriced case.
+
+**GREEN:** `_cost_text()` helper plus one key in `_submission()`.
+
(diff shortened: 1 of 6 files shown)


keelancj (approved):
Reviewed and verified against the branch

- Ran the full test suite: 535 passed, 3 skipped, 3 deselected. compute_pareto_frontier is confirmed unimported anywhere else, so safe to merge alone as claimed.
- Reproduced both mutation-testing claims exactly: reverting "null cost → 0" fails 3 tests. reverting to the ticket's literal strict-both dominance fails 3 different tests.
- The revision-partitioning fix is a real well-caught correctness bug
- Spot-checked beyond the test file the 3-way exact ties all survive correctly, and same-cost chains correctly keep only the top scorer

No blocking issues, approving

PR #778, opened 2026-08-29
feat(scoreboard): compute the Pareto frontier of score against cost (OME-923 part A)

FilipBolt:
Part **A only** of [OME-923](https://linear.app/openmined/issue/OME-923) — the pure function
deciding which submissions hold the "best score for the money" claim.

**This does not close OME-923.** Parts B (mark rows) and C (accuracy-vs-cost chart) remain, and
are gated: every `run_cost_usd` on the live board is null today, so B would render a frontier of
nothing. The ticket's own gate allows A against fixtures now. Part A is safe to merge alone
because nothing imports it yet.

## Two decisions that differ from a literal reading of the ticket

Both were escalated to the owner rather than assumed.

1. **Standard Pareto dominance**, not the ticket's *"no other has **both** a higher score **and**
   a lower cost"*. Read strictly, that keeps a row scoring the same at nine times the price,
   because nobody outscored it — indefensible on a board claiming best value. Exact ties on both
   axes still both qualify, which is what the ticket's "ties both qualify" line means.
2. **Returns `frozenset[tuple[str, str | None]]`, keyed on `(spec_id, benchmark_revision)`**,
   which survives the re-sorting the ranking route does. Rows are compared only within one
   revision — a benchmark with no registered revision filters nothing and ranks one row per
   `(spec_id, revision)`, and `OME-775` exists because those numbers are not comparable.

Irina's 2026-08-24 comment on the ticket settled the other four open questions: cost is the whole
run, baselines are excluded, the frontier mark is "Pareto-frontier SOTA" distinct from absolute
SOTA, and 3D charts stay in OME-324.

## Naming

`pareto.py` / `compute_pareto_frontier`, deliberately not `frontier` anything — OME-323's
open/closed frontier owns `frontier.py`, `compute_frontier`, `FrontierPoint` and `FrontierResult`
and keeps its name, endpoint and meaning. That file is untouched.

## Test plan

11 new tests, no DB. Empty input · lone entry · strict domination · genuine tradeoff (both
qualify) · equal score (cheaper wins) · equal cost (higher score wins) · exact tie (both) ·
unpriced excluded at top score · unpriced never dominates a priced row · all-null board yields
empty without error · zero cost is real and can win.

**Mutation-checked**, re-verified after the formatter rewrote `_dominates`:

| Reverted guard | Result |
|---|---|
| null cost read as `0` | 3 tests fail |
| ticket's literal strict-both dominance | 3 tests fail |

Gates all green: append-only, ruff, ruff format, pyright (no `type: ignore`), pytest
`--cov-fail-under=80`, Node portal test. Suite 518 → **529 passed**, 3 skipped.

Ledger: `docs/work/2026-08-29-OME-923-pareto-frontier-compute.md`

https://claude.ai/code/session_014MJnetigGbLfyz6hLiGkij

---

## Updated after review (2026-08-31)

Two P2 findings were raised and both were confirmed against `_build_leaderboard_query` before
acting, then fixed in `61dcd464`:

- **Revision keying.** The bare `spec_id` key was wrong on any benchmark with no registered
  revision: the board filters nothing there, so one spec appears once per revision and a row
  beaten on BOTH axes inherited the mark its own spec earned elsewhere. Reproduced first. Now
  keyed on `(spec_id, benchmark_revision)`, with comparisons partitioned by revision.
- **`top_n` truncation.** `leaderboard()` orders by score alone before truncating, so a row past
  the cutoff can tie the boundary score at a lower cost and dominate a visible row. This module
  carries the contract as an `AIDEV-NOTE`; part B honours it by reading the board whole.

Thanks for catching the stale wording — the task mirror and the section above now match the
implementation. Both said `frozenset[str]`, which is exactly the regression not to reintroduce.

Tests 11 → 17. Mutation-checked: collapsing every revision into one cohort fails; dropping the
revision from the key fails.

### DIFF
--- a/apps/scoreboard/src/scoreboard/scores/pareto.py
+++ b/apps/scoreboard/src/scoreboard/scores/pareto.py
@@ -1,0 +1,78 @@
+"""Pareto frontier computation (OME-923 part A).
+
+Pure function, no I/O — takes already-fetched schemas so it's directly
+unit-testable without a DB, mirroring `frontier.py`.
+
+AIDEV-NOTE: deliberately NOT named `frontier` anything. `frontier.py`,
+`compute_frontier`, `FrontierPoint` and `FrontierResult` are OME-323's open/closed
+frontier — a different measure, on a different endpoint, that keeps its name and
+meaning. Qualifying this one keeps a grep for either honest (OME-923).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from decimal import Decimal
+
+from .schemas import LeaderboardEntry
+
+
+def compute_pareto_frontier(entries: Sequence[LeaderboardEntry]) -> frozenset[str]:
+    """The `spec_id`s whose score-for-cost no other entry beats.
+
+    FEATURE: OME-923 — "best score for the money" is a set, not a single row, so
+    several participants can hold a defensible claim on the same board.
+
+    INVARIANT: one row per `spec_id`. Callers pass the RANKED board
+    (`ScoreStore.leaderboard`), which collapses to best-per-spec. Do NOT pass
+    `ScoreStore.list_owned_entries` — it deliberately does not collapse, so a
+    repeated `spec_id` would mark rows that are not themselves on the frontier.
+
+    INVARIANT: an absent cost is excluded, never read as 0. Zero would dominate
+    every priced row and hand the board to an unknown (OME-770 D8). The exclusion
+    follows the missing VALUE, not the row: populate that cost later and the row
+    joins the frontier on its own, with no backfill.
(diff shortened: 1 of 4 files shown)


HupBaHa (approved):
Approved. I found only one low-severity docs/body wording issue: the task mirror and PR body still say Part A returns `frozenset[str]` of `spec_id`, but the fixed implementation now returns `(spec_id, benchmark_revision)` keys. Not a runtime blocker, just worth cleaning up so Part B doesn’t accidentally drop the revision again.

Code and tests look good. Scoreboard gates are green locally, and GitHub checks are green.

PR #782, opened 2026-08-31
feat(aigateway): cache Tavily retrieval results in the global cache

IonesioJunior:
## Why

Tavily-backed retrieval (`web_search` / `web_fetch`) executes inside the Runner Job and never
traversed aigateway, so an identical retrieval was paid for on every run.

The larger cost was **second-order**. aigateway's global cache hashes `messages` verbatim, and a
tool result is appended to `messages` and re-sent on every later iteration. Non-deterministic
Tavily output therefore permanently poisoned the continuation chat call's key:

```
identical user request
 → turn 1 chat call ......... CACHE HIT   (already worked)
 → Tavily /search ........... fresh results, DIFFERENT bytes every time
 → continuation chat call ... CACHE MISS, permanently, by construction
```

The continuation is the expensive call — it carries up to 32 KB (≈8K tokens) of injected content,
re-sent per iteration, with `web_tool_max_iterations = 12`. Caching retrieval saves Tavily credits
**and** restores byte-identical tool results, which is what unlocks the existing chat cache for the
whole loop.

## What this PR does (the aigateway half)

aigateway receives a **description** of a retrieval request, derives the cache key itself, and
stores an opaque result string. **It never holds a Tavily credential and never calls Tavily** — the
Runner keeps both. `OME-1045` wires the Runner side and is blocked on this.

| Endpoint | Behaviour |
|---|---|
| `POST /v1/retrieval/tavily/cache/lookup` | `200` with `{"status": "hit"\|"miss"\|"bypass", ...}` |
| `POST /v1/retrieval/tavily/cache/entries` | `200` with `{"outcome": "stored"\|"race_lost"\|"not_stored"}` |

A miss is a `200` **answer**, not a `404`, following the house precedent of `POST /v1/models/admit`:
the caller's next move needs the body either way.

### Files

- **`core/request_cache/canonical.py`** (new) — the one canonical form, extracted from
  `global_keys.py` so both lanes hash identically. Every export runs the json-safety guard, and it
  is a dependency-free leaf. `global_keys.py`'s public surface is unchanged
  (`CanonicalizationError` re-exported, `canonical_key_material` byte-identical).
- **`core/request_cache/tavily_retrieval.py`** (new) — the closed key DTO, the reviewed URL/query
  normalization, and the contract revision.
- **`core/request_cache/tavily_store.py`** (new) — insert-only over the **existing**
  `request_cache_entries` (`provider='tavily'`, `model=<tool>`, `expires_at=NULL`). **No migration.**
- **`routes/tavily_retrieval_cache.py`** (new) — the two endpoints.
- `store.py` — `_INFRASTRUCTURE_ERRORS` → `INFRASTRUCTURE_ERRORS` so both stores share one
  definition. Mechanical.
- `DEPLOYMENT.md` — the operator section.

## Decisions worth reviewing

- **The key is derived server-side.** A client-supplied key would let two Runner versions with
  different normalization silently split or share rows — the defect OME-777/D2 removed from the
  chat lane.
- **`excluded_domains` is IN the key, and this is security-critical.** A cached hit returns an
  already-formatted result string that can no longer be re-filtered, so the key is the only place a
  benchmark's retrieval policy (DRACO excludes arxiv/semanticscholar) can be honoured. Tested at
  both the key and the route level.
- **No expiry** — `expires_at = NULL`, 1:1 with the chat lane. Determinism over freshness, so a
  re-run keys *and* answers identically. Accepted cost: a stale web result can be served
  indefinitely; the documented per-provider reset is the correction path.
- **No configuration.** The lane is unconditional — no env var, no chart value, no availability
  gate. A test pins that the store takes no constructor argument so a config layer cannot creep
  back in.
- **`operation = "retrieval.tavily"`** sits inside the hashed material, so retrieval keys and chat
  keys are provably disjoint despite sharing one table.
- **Not stripping `utm_*`** from fetch URLs. Over-normalizing risks serving a different page's
  content; two URLs that differ are two rows.

## Testing

`56` new tests. Highlights: an exact-bytes pin on the hashed member set (so the key cannot drift
silently), the U+2028 regression that mis-keyed the DRACO backfill, the non-string-object-key
coercion guard, first-fill-wins under a simulated race, undecodable rows refused rather than
served, and the exclusion-set leakage guard end to end.

The chat lane's own suite (`test_global_cache_key.py`, `test_global_cache_plan.py`,
`test_openrouter_global_cache_keys.py`) passes unchanged, which is what demonstrates the canonical
extraction is behaviour-preserving.

Gates: `ruff check`, `ruff format --check`, `pyright`, `check_no_enterprise.py`, and
`pytest --cov=aigateway --cov-fail-under=80` all green via
`.claude/scripts/run_gates.py aigateway`.

## Follow-ups (not in this PR)

- `OME-1045` — the Runner wiring (blocked on this).
- An in-run single-flight memo, so a concurrent fan-out issuing the same query pays one Tavily call
  rather than N.
- Retrieval cache outcomes in run accounting.

Refs: OME-1044


HupBaHa (changes requested):
Requesting changes because the extra-field handling creates a real wrong-hit path.

1. Blocking: the retrieval request models accept and silently ignore unknown fields.

`apps/aigateway/src/aigateway/routes/tavily_retrieval_cache.py:59`

`_Description` / `_FillRequest` use Pydantic v2 default behavior, so extra fields are ignored. For a cache-key endpoint, every output-affecting field must either be in the key or be rejected.

I reproduced this on the PR app:

- fill unrestricted row with `excluded_domains: []`
- lookup the same query with Tavily/native spelling `exclude_domains: ["arxiv.org"]`
- actual: `200 hit` with the unrestricted cached result
- lookup with contract spelling `excluded_domains: ["arxiv.org"]` correctly returns `miss`

That defeats the load-bearing exclusion-policy guard. Please fail closed on extra fields, e.g. `ConfigDict(extra="forbid")`, and add a regression test for `exclude_domains`.

2. Please also emit the promised `Cache-Status` header.

`apps/aigateway/src/aigateway/routes/tavily_retrieval_cache.py:112`

The PR/spec/DoD say the endpoints emit `Cache-Status` plus the `X-AIGW-Cache*` fallback headers, but `_publish()` currently emits only `X-AIGW-Cache*` and `Age`. The current Runner parser tolerates this through legacy fallback, so this is not the crash, but the PR contract is not met.

3. Please register the Tavily retrieval key revision in the cache snapshot revision set.

`apps/aigateway/src/aigateway/core/request_cache/revisions.py:38`

`TAVILY_RETRIEVAL_CONTRACT_REVISION` is inside the Tavily key hash, but `active_cache_revisions()` does not report it. Since snapshots now move the shared `request_cache_entries` table including `provider='tavily'` rows, manifests should cover this revision too. Otherwise old Tavily rows can be loaded silently and never hit.

Tortoise/ORM boundary otherwise looks clean: no new model or migration, the existing `RequestCacheEntry` model layout is consistent with the app pattern, and the transaction/`IntegrityError` handling mirrors the established request-cache store discipline.

Local checks I ran:

- focused cache/key tests: 196 passed
- `git diff --check origin/main...HEAD`
- `uv run .claude/scripts/run_gates.py aigateway`

All passed, but the P1 wrong-hit path still needs fixing before merge.

HupBaHa (approved):
Code review looks clean from my pass. The previous blockers are addressed: extra fields now fail closed, Cache-Status is emitted, and the Tavily retrieval revision is included in active_cache_revisions.

Non-blocking follow-up: quote the Cache-Status key parameter per RFC 9211, i.e. emit key="<prefix>" instead of key=<prefix>, and update the route test. The current Runner parser is tolerant and the legacy X-AIGW-Cache-Key fallback is still present, so this does not block merge.

PR #785, opened 2026-09-01
fix(scoreboard): keep partial runs out of the ranked leaderboard

FilipBolt:
Closes OME-1056. Parent: OME-867.

A run covering fewer cases than the benchmark defines ranked alongside a complete one — and beat it. Fewer cases makes a perfect score easier, so partial runs were not merely admitted, they were **advantaged**.

Reproduced before any code was written:

```
rank  spec_id              score   total_questions
1     one-case-run         1.0     1
2     honest-full-run      0.85    541
```

541 is IFEval's real `CASE_COUNT`. That is the first test in this PR, and it failed on `main`.

## Why the board could not tell

`total_questions` is validated only for `> 0`. `Benchmark` stored no expected count. The ranking filtered on `benchmark_id` and `benchmark_revision` alone.

Meanwhile the Engine has **always** published `case_count` in the catalogue the seed job fetches on every deploy, and `_CatalogEntry` set `extra="ignore"` and threw it away. So the number needed to spot a partial run was arriving all along.

## What changed

- `Benchmark.case_count`, nullable, migration `0011`, no backfill.
- The seed carries it from the Engine catalogue.
- The ranking excludes rows below it.

**A board declaring no count filters nothing**, mirroring the revision rule immediately above it, so legacy and non-Engine boards rank exactly as before.

**Deployment configuration may not declare a count**, exactly as it may not declare a revision. A hand-written `case_count: 1` would re-open this hole from the other side, so such a row is refused and named.

## The subtle part

The filter sits **inside** the window, beside the revision filter, not after it.

The ranking assigns `row_number()` per `(spec_id, revision)` ordered by score, then keeps `rn == 1`. Filter partial rows *after* that, and a spec whose partial run outscored its own complete run gives the partial row `rn = 1` — so the outer filter drops the spec entirely and the participant's full run vanishes for having smoke-tested first. SQL evaluates `WHERE` before window functions, which is what makes the inner placement correct.

Mutation-checked: moved outside, exactly one test fails, with `[] == [('same-spec', 0.6)]`. The other five still pass, so the test is specific rather than redundant.

## What deliberately did not change

Partial runs stay **accepted, stored and readable** — through history, `GET /v1/scores/{id}`, and the submission response, none of which filter on revision or rank. Running `limit=1` to prove the submission path works end to end is a supported workflow, and the client already warns the run is partial (OME-922). Rejecting at write time was considered and rejected: it discards real work, contradicts that warning, and does nothing about rows already ranked.

Dedup already keeps the two apart. `total_questions` participates in `_content_hash`, and the SDK sends `Idempotency-Key: run_id`, unique per run — so a `limit=1` run and a full run of the same recipe are two rows on both paths. Neither displaces the other.

## Existing rows fix themselves

Ranking is computed at read time and `total_questions` is already stored on every row, so partial entries already in the table drop out of the ranking on the next request. The `scores` table is untouched and nothing is deleted.

## Scope

This is the common first step of every option still open on OME-867 — the main board stops ranking partials under all of them — so it does not pre-empt that design. A separate ranking for partial runs, grouped by matched coverage, can be built on the column this adds.

Pairs with OME-909: a submitter whose score does not rank should be told why.

## One thing left alone on purpose

`test_catalog_fields_the_board_does_not_display_are_ignored` names `case_count` in its comment as an example of a field the board ignores. That parenthetical is now stale, but its assertion is unaffected, so under the append-only rule I did not touch it. The staleness is recorded in the new seed test's module docstring.

## Gates

ALL GATES GREEN. 544 passed, 3 skipped. `makemigrations` reports no drift. The append-only check passes unaided.

### DIFF
--- a/apps/scoreboard/src/scoreboard/scores/migrations/0011_benchmark_case_count.py
+++ b/apps/scoreboard/src/scoreboard/scores/migrations/0011_benchmark_case_count.py
@@ -1,0 +1,39 @@
+from tortoise import fields, migrations
+from tortoise.migrations import operations as ops
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("models", "0010_idempotency_key_scheme"),
+    ]
+
+    initial = False
+
+    # FEATURE: OME-1056 — the number of cases a benchmark defines, so a partial run stops
+    # ranking against a complete one. Fewer cases makes a perfect score easier, so before this
+    # a one-case run did not merely rank, it WON.
+    #
+    # WHY nullable with NO backfill: NULL is meaningful here and is not a gap to be filled. It
+    # means "this board declares no canonical scope", and such a board filters nothing — the
+    # same rule `revision` already follows for the retained legacy entries. Inventing a count
+    # for a board the Engine does not publish would assert a scope nobody verified.
+    #
+    # The value is filled by the seed job from the Engine catalogue, which already carries
+    # `case_count` and was discarding it. So no data migration is needed: the next seed after
+    # this deploy populates every Engine-published board.
+    #
+    # AIDEV-NOTE: no `scores` change and no backfill of existing submissions. Ranking is
+    # computed at READ time from `total_questions`, which is already stored on every row, so
+    # partial rows already in the table drop out of the ranking on the next request and stay
+    # readable through history. Nothing is deleted.
+    #
+    # AIDEV-NOTE: SAFE for a rolling multi-replica rollout — old pods ignore a column they do
+    # not know. See "Breaking migrations and multi-replica rollouts" in DEPLOYMENT.md.
+
+    operations = [
+        ops.AddField(
+            model_name="Benchmark",
+            name="case_count",
+            field=fields.IntField(null=True),
+        ),
+    ]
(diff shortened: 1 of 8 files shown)


HupBaHa (approved):
Approved. I do not see a request-changes blocker for PR #785.

The scoped OME-1056 acceptance criteria are met on the merge result: partial runs are excluded from ranked `entries`, remain accepted/stored/readable, no-`case_count` boards keep prior behavior, configured `case_count` is refused, the reproduction is covered, and gates are green.

I would track these Medium findings as follow-ups rather than block this PR:

- The public `/frontier` endpoint can still report a partial run as the current holder.
- Local-stack seeding drops `case_count`, so local `screamingface up` boards can still miss the new filter.
- The Engine→Scoreboard producer contract is not pinned for `case_count`.
- The seed-to-DB `case_count` wiring is not covered by a test.
- The `total_questions >= case_count` behavior is not covered against an accidental `==` change.
- Board completeness is weaker than the SDK partial-run definition because grading coverage is not on the wire.

None of these invalidates the scoped fix in this PR. Recommended follow-up: file/record tickets for the frontier, local-stack, and completeness-definition items, and add the small contract/seed/ranking test pins if convenient.

PR #786, opened 2026-09-01
feat(scoreboard): mark Pareto frontier rows (OME-923 part B)

FilipBolt:
## Summary

- mark public leaderboard rows that are on the score-for-cost Pareto frontier
- show run cost and explain the mark and self-reported cost provenance in the portal
- keep the visible leaderboard bounded while computing the whole-board frontier from a narrow O(n log n) projection
- correlate independent projections by opaque row identity so concurrent submissions cannot transfer marks
- preserve private-board isolation and fail closed for unpinned benchmarks

## Test plan

- uv lock --check
- uv run ruff check --output-format=github
- uv run ruff format --check
- uv run pyright
- uv run pytest --tb=short --cov=scoreboard --cov-report=term-missing --cov-fail-under=80 -q
- node --test tests/portal/leaderboard-logic.test.js

Result: 551 Python tests passed, 3 skipped, 3 deselected; 88.68% coverage; 40 portal tests passed.

## Merge gate

OME-1029 is merged but the SDK cost producer was still unreleased in the latest Linear update. Do not merge Part B until a released client has submitted real non-null run cost data.


HupBaHa (approved):

FilipBolt:
Integrated merged #820/main without rebasing the reviewed stack.

New head: `bbf5ee20`
- `ad8a66de` records the published-client round trip
- `32193f16` merges main and gives the bounded display query plus the unbounded Pareto projection the same registered revision + `case_count` snapshot
- `bbf5ee20` records the integration and verification

Added an adversarial regression: a one-case row with a higher score and lower cost neither ranks nor shapes the visible Pareto marks. Local verification: 110 focused tests; full Scoreboard gate green; 576 passed, 3 skipped, 3 deselected; 88.54% coverage; 40 portal tests passed.

GitHub still displays the approval attached to old head `772a57e9`; requesting a fresh pass on the new head. The live priced-ranked-row merge gate remains open, so this is not being merged yet.

PR #820, opened 2026-09-02
fix(scoreboard): close partial-run leaderboard follow-ups

FilipBolt:
## Summary

- apply the registered `case_count` to the public frontier as well as the ranked table
- preserve omitted store updates while clearing stale scope for authoritative Engine rows
- carry `case_count` through local `screamingface up` seeding
- pin the Engine producer contract, seed-to-database persistence, and the intentional `>=` boundary

## Review follow-up boundary

This addresses five of the six non-blocking findings from #785. The remaining typed grading-coverage and incomplete-grading policy is already tracked in OME-867 and is intentionally not folded into this implementation.

## Verification

- `run_gates.py scoreboard`: all gates green
- `run_gates.py screamingface`: all gates green
- Engine append-only, Ruff, formatting, Pyright, layering, and changed 18-test catalogue suite: green
- Engine full pytest: 2,283 passed, 5 skipped, 1 unrelated local environment failure; Helm 4 defaults to Kubernetes 1.20 while the untouched chart requires 1.25+. The chart renders successfully with `--kube-version 1.25.0`.

Refs: OME-1056

### DIFF
--- a/apps/scoreboard/src/scoreboard/routes/leaderboard.py
+++ b/apps/scoreboard/src/scoreboard/routes/leaderboard.py
@@ -350,5 +350,12 @@ async def get_frontier(benchmark_id: str, request: Request) -> FrontierResponse:
             detail=FRONTIER_NOT_AVAILABLE_DETAIL,
             headers=PRIVATE_CACHE_HEADERS,
         )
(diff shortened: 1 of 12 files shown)


HupBaHa (approved):

PR #830, opened 2026-09-03
feat(screamingface): support leaderboard co-authors

keelancj:
## Summary
- add an exact co-author email list to lazy, synchronous, and asynchronous leaderboard submissions
- expose privacy-trimmed authors on immutable LeaderboardScore and LeaderboardEntry values
- validate the approved 1–10 author write contract and document the public API

## Dependency
Draft until the Scoreboard implementation for OME-1052 / OME-1054 is merged and deployed. The currently deployed Scoreboard rejects the new authors field.

Scoreboard branch: https://github.com/ScreamingFace/screamingface/compare/OME-1051-scoreboard-authors
Linear: https://linear.app/openmined/issue/OME-1053/add-co-author-emails-to-client-submissions

## Test plan
- 16 author-focused tests pass
- full ScreamingFace SDK gate suite passes: ruff check and format, pyright, pytest with 95%+ coverage, deterministic notebooks, wheel build, and distribution validation
- public API snapshot regenerated and verified

## Process note
The append-only precheck was skipped only for the owner-approved public-surface snapshot regeneration. Test behavior was added append-only.

Refs: OME-1053

Asana: none

### DIFF
--- a/docs/plan/2026-09-03-OME-1053-client-authors.md
+++ b/docs/plan/2026-09-03-OME-1053-client-authors.md
@@ -1,0 +1,13 @@
+# OME-1053 — Client authorship implementation plan
+
+1. Add failing tests for the three public submission entry points, omission semantics, exact-list
+   preservation, local validation boundaries, response decoding, and immutable returned values.
+2. Thread the keyword-only `authors` parameter through lazy, sync, and async submission methods
+   into the one shared payload builder.
+3. Validate author arguments once beside the payload builder and conditionally add the JSON field.
+4. Add optional immutable `authors` fields to `LeaderboardScore` and `LeaderboardEntry`, then
+   decode non-null Scoreboard arrays into tuples.
+5. Refresh the public API snapshot, run focused and full tests, then run the complete
+   `screamingface` gate set.
+6. Complete the ledger, commit with `Refs: OME-1053`, push, and open a draft PR that names the
+   unmerged/deployment dependency on the Scoreboard implementation.
--- a/docs/spec/2026-09-03-OME-1053-client-authors.md
+++ b/docs/spec/2026-09-03-OME-1053-client-authors.md
@@ -1,0 +1,65 @@
+# OME-1053 — Client authorship contract
+
+Status: approved for implementation on 2026-09-03.
+Parent contract: `docs/spec/2026-09-01-OME-1051-multiple-authors.md` on the Scoreboard feature
+branch.
+
+## User contract
+
+All submission surfaces accept the same keyword-only parameter:
+
+```python
+authors: Sequence[str] | None = None
+```
+
+The supported calls are the lazy `sf.leaderboards.submit`, bound synchronous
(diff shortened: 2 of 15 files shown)


FilipBolt:
Scoreboard dependency is now merged and live on dev: #833 landed as `83b15463`. I verified the deployed OpenAPI exposes `authors` on the submission and response schemas, and a live leaderboard read succeeds against the migrated database. This draft can proceed when you are ready.

FilipBolt:
Read this against the Scoreboard half (#833, `83b15463`). The contract lines up exactly — same email pattern, same 255 / 10 / 1 bounds, and `authors` omitted rather than sent as null when unspecified, which is what the nullable column needs to tell "not specified" from "specified as just me". The 409 treatment is right too: both causes on the submit route, `BenchmarkVisibilityChanged` and `ConcurrentScoreUpdate`, are genuinely transient, so marking them retryable matches what the server means.

Three things. Only the first is a request for this PR.

**1. A test for the resubmission-correction path.** Every `submit` test here returns 201 — the 200s are all `get` / `get_score`. So the one uncovered path is the one the guide promises:

> A resubmission by the same submitter can correct its author list.

That is OME-1054, merged on the Scoreboard yesterday. A fake-transport test covers it: submit, return 200 carrying a changed author list, assert the decode. Worth having before this lands, since it is the seam between the two halves and neither side tests it from the client today.

**2. Worth its own ticket — the client discards the created-vs-deduped signal.** The Scoreboard answers 201 on create and 200 when the submission dedups onto an existing row (`routes/scores.py:249`). `_response_json` only checks `is_success`, so that distinction is lost.

It matters because `content_hash` deliberately excludes `submitted_by`. On a public board, submitting a recipe someone else already submitted at the same score returns **their** row — their submitter, their authors — under a 200, indistinguishable from a new submission. Pre-existing rather than introduced here, but it gets sharper now that credit is a first-class field. Could you file it? Happy to do it instead if easier.

**3. FYI only, nothing to change.** The guide documents that duplicate authors are preserved. [OME-1109](https://linear.app/openmined/issue/OME-1109) covers collapsing duplicates and disambiguating colliding credit names — it is Backlog and not expected soon, so preserving them is the right thing to ship today. Flagging it so that docs line is a known revisit rather than a surprise later.

keelancj:
Addressed the actionable follow-ups in `5bf3b3b8`:

- Added a fake-transport regression proving a score POST that returns `200` decodes the corrected public author list as an immutable tuple. The focused test passes (`1 passed, 79 deselected`), and the complete ScreamingFace gate runner is green.
- Filed [OME-1123](https://linear.app/openmined/issue/OME-1123/preserve-the-created-versus-deduplicated-leaderboard-submission) for preserving the Scoreboard’s `201`-created versus `200`-deduplicated signal in the public Client API. It is related to OME-970 and intentionally not implemented in this PR.
- Duplicate preservation remains unchanged; OME-1109 owns that future behavior.

The new test is characterization coverage: the client already handled all successful statuses, so no production change was necessary.

PR #832, opened 2026-09-03
feat(e2e): add a setup validator for the traceability lanes (OME-1106)

sergio-bershadsky:
## Summary

`python3 e2e/failor/check_setup.py` — one command that says whether this machine can run the traceability rungs, and exactly what to type for anything missing.

Every prerequisite it checks was discovered the hard way during `OME-1105`: the `[runtime]` extra, prepared draco assets, Docker, `SCREAMINGFACE_TEST_E2E=1`, a client carrying `OME-967`, and a kubectl context.

**The failure mode it exists to prevent:** without prepared assets, *every* e2e rung skips and `pytest` exits `0`. An all-skipped run is indistinguishable from a passing one by exit code — the most likely way to believe the chain was validated when nothing ran. The validator states that outright.

## Output

```
LOCAL LANE — the correlation ladder (decides the exit code)
  OK       docker                  daemon reachable
  OK       [runtime] extra         all local-stack modules present in packages/screamingface/.venv
  OK       draco assets            found at ~/.screamingface/benchmark-assets
  OK       SCREAMINGFACE_TEST_E2E  set to 1
  OK       client has OME-967      ScreamingFaceError accepts trace_id

K8S LANE — the live notebook (reported only, never fails this run)
  MISSING  kubectl context   NO contexts defined (current-context names 'k3s-ansible';
                             2 cluster(s) present but nothing binds them)
  UNKNOWN  firecall bastion  firecall@172.190.209.255 -> AKS aks-dev-eastus; ns: sf-aigw, ...
```

## Three design points

**Read-only and offline.** It never SSHes, authenticates, or contacts a cluster. Bastion reachability is reported `UNKNOWN`, not probed — opening a session to a jump host is a credentialed action belonging to the operator, not to a status command. The bastion details are recorded as constants so nobody has to SSH in to rediscover them.

**Only the local lane decides the exit code.** The k8s lane needs credentials nobody here controls; failing on it would make the validator permanently red and therefore ignored.

**`find_spec` via subprocess, not import.** Importing `litellm` to test presence installs logging handlers at import time (see `OME-1050`) and is slow.

## It caught a defect in itself

The first version used **in-process** `importlib.util.find_spec` — but the script is invoked as `python3 e2e/failor/check_setup.py` from the repo root, so it probed whatever python was on `PATH` rather than the client venv. It reported the `[runtime]` extra **missing while it was installed in the only environment that runs the lane**.

A validator that lies about readiness is worse than no validator. It now probes `packages/screamingface/.venv/bin/python` explicitly and names which interpreter it used, so a fallback is visible rather than silent.

## Test plan

- `ruff check` + `ruff format --check` clean — they run repo-wide via pre-commit, including this path (`OME-1074` established that the hard way).
- **Both exit paths observed, not reasoned about:** `exit=1` with two prerequisites missing, then `exit=0` with all five green after `uv sync --extra runtime`.
- No stack gate applies (top-level directory outside every path filter).

Ledger: `docs/work/2026-09-03-OME-1106-setup-validator.md`
Mirror: `docs/tasks/2026-09-03-OME-1106-setup-validator.md`

Refs: OME-1106

### DIFF
--- a/e2e/failor/check_setup.py
+++ b/e2e/failor/check_setup.py
@@ -1,0 +1,265 @@
+#!/usr/bin/env python3
+"""Report whether this machine can run the traceability e2e lanes (OME-1106).
+
+    python3 e2e/failor/check_setup.py
+
+Two lanes are checked independently:
+
+- **local** — the correlation ladder (`packages/screamingface/tests/e2e/
+  test_correlation_chain.py`). Needs only Docker and a prepared board; it is the lane that
+  can actually be made green on a laptop, so it alone decides the exit code.
+- **k8s** — the live notebook (`e2e/failor/notebooks/`). Needs cluster credentials nobody
+  here controls, so it is reported but never fails the run; a permanently red validator is
+  an ignored one.
+
+INVARIANT: read-only and offline. This script never SSHes, authenticates, or contacts a
+cluster. Bastion reachability is reported UNKNOWN rather than probed — opening a session to a
+jump host is a credentialed action that belongs to the operator, not to a status command.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+CLIENT = REPO_ROOT / "packages" / "screamingface"
+
+# From `firecall-connect` on the bastion. Recorded so nobody has to SSH in to learn them.
+BASTION = "firecall@172.190.209.255"
+AKS_CLUSTER = "aks-dev-eastus"
+AKS_GROUP = "rg-aks-platform-dev-eastus"
+NAMESPACES = ("sf-aigw", "sf-fusion", "sf-scoreboard")
+
+# The modules the `[runtime]` extra provides that the local stack actually boots.
+RUNTIME_MODULES = (
+    "uvicorn",
+    "fastapi",
+    "tortoise",
+    "litellm",
+    "aiosqlite",
+    "pydantic_settings",
+)
+
+OK, MISSING, UNKNOWN = "OK", "MISSING", "UNKNOWN"
+
+
+def _client_python() -> tuple[str, str]:
+    """The interpreter to probe, and a label saying which one it is.
+
+    WHY not `sys.executable`: this script is run as `python3 e2e/failor/check_setup.py` from
+    the repo root, so `sys.executable` is whatever python is on PATH — usually NOT the client
+    venv. Probing that would report the `[runtime]` extra missing while it is installed in the
+    only environment that runs the lane, which is worse than not checking at all.
+    """
+    venv = CLIENT / ".venv" / "bin" / "python"
+    if venv.is_file():
+        return str(venv), "packages/screamingface/.venv"
+    return sys.executable, f"fallback: {sys.executable}"
+
+
+@dataclass(frozen=True, slots=True)
+class Check:
+    name: str
+    state: str
+    detail: str
+    remedy: str = ""
+
+
+def _docker() -> Check:
+    if shutil.which("docker") is None:
+        return Check(
+            "docker", MISSING, "not on PATH", "install Docker Desktop / colima"
+        )
+    probe = subprocess.run(["docker", "info"], capture_output=True, text=True)
+    if probe.returncode != 0:
+        return Check(
+            "docker", MISSING, "daemon unreachable", "start Docker, then re-run"
+        )
+    return Check("docker", OK, "daemon reachable")
+
+
+def _runtime_extra() -> Check:
+    # WHY find_spec and not import: importing litellm installs logging handlers at import
+    # time (OME-1050) and is slow. Presence is the question; execution is not needed.
+    python, label = _client_python()
+    probe = subprocess.run(
+        [
+            python,
+            "-c",
+            "import importlib.util,sys;"
+            f"mods={list(RUNTIME_MODULES)!r};"
+            "print(','.join(m for m in mods if importlib.util.find_spec(m) is None))",
+        ],
+        capture_output=True,
+        text=True,
+    )
+    if probe.returncode != 0:
+        return Check(
+            "[runtime] extra",
+            UNKNOWN,
+            f"probe failed in {label}",
+            "cd packages/screamingface && uv sync --extra runtime",
+        )
+    absent = [m for m in probe.stdout.strip().split(",") if m]
+    if absent:
+        return Check(
+            "[runtime] extra",
+            MISSING,
+            f"absent in {label}: {', '.join(absent)}",
+            "cd packages/screamingface && uv sync --extra runtime",
+        )
+    return Check("[runtime] extra", OK, f"all local-stack modules present in {label}")
+
+
+def _assets() -> Check:
+    override = os.environ.get("SCREAMINGFACE_E2E_ASSETS")
+    roots = [Path(override)] if override else []
+    roots += [
+        Path.home() / ".screamingface" / "benchmark-assets",
+        CLIENT / "tests" / "e2e" / "fixtures" / "assets",
+    ]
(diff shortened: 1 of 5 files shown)


e2e/failor/check_setup.py
```diff
@@ -0,0 +1,265 @@
...
+    return Check("[runtime] extra", OK, f"all local-stack modules present in {label}")
+
+
+def _assets() -> Check:
+    override = os.environ.get("SCREAMINGFACE_E2E_ASSETS")
+    roots = [Path(override)] if override else []
+    roots += [
+        Path.home() / ".screamingface" / "benchmark-assets",
+        CLIENT / "tests" / "e2e" / "fixtures" / "assets",
+    ]
```
IonesioJunior: **The validator accepts an assets location the test never reads.**

`_assets()` looks in three places. `test_correlation_chain.py` looks in two, and `~/.screamingface/benchmark-assets` is not one of them:

```python
# test_correlation_chain.py:57-59
def _assets_root() -> Path:
    override = os.environ.get(_ASSETS_ENV)
    return Path(override) if override else FIXTURES_DIR / "assets"
```

So when `SCREAMINGFACE_E2E_ASSETS` is unset and the assets sit in the home directory, this reports `OK` and the test still skips.

This is not hypothetical — it reproduces on a clean checkout today. `~/.screamingface/benchmark-assets/draco/` exists, because that is where `screamingface prepare draco` writes, which is the command this function recommends on line 134. Meanwhile `tests/e2e/fixtures/assets/` does not exist in the repo at all. Running the script gives:

```
OK  draco assets  found at /home/<user>/.screamingface/benchmark-assets
...
LOCAL LANE READY.
```

All five rungs would then skip and `pytest` would exit 0 — the exact all-skipped-looks-like-passing case this script was written to catch.

**The fix may belong in the test, not here.** `test_boards.py:63-71` and `test_failures.py:369-372` both default to `default_data_dir() / "benchmark-assets"`, with an explicit comment citing OME-1001: *"it is where `screamingface prepare` writes, so the assets a dev prepares for the stack are the assets these tests find."* `test_correlation_chain.py` is the only one of the three that does not follow that convention. Aligning the test would fix the mismatch and make the ladder runnable with assets a developer already has; aligning this function to the test is the smaller change. Either is fine — they just must not stay different.

Two smaller points in the same area:

- Line 124 hardcodes `Path.home() / ".screamingface"`, but `default_data_dir()` honors `SCREAMINGFACE_DATA_DIR` first (`_runtime/config.py:12-16`). Anyone who sets that variable gets a wrong answer regardless of which fix is chosen.
- The ready command printed on lines 254-256 does not export `SCREAMINGFACE_E2E_ASSETS`, so nothing forces the discrepancy into view at the moment it would matter.

e2e/failor/check_setup.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def main() -> int:
+    local = [_docker(), _runtime_extra(), _assets(), _e2e_env(), _client_has_trace_id()]
+    k8s = [_kubectl_context(), _bastion()]
+
+    _render("LOCAL LANE — the correlation ladder (decides the exit code)", local)
+    _render("K8S LANE — the live notebook (reported only, never fails this run)", k8s)
+
+    blocked = [c for c in local if c.state == MISSING]
```
IonesioJunior: **`UNKNOWN` does not block, so the script can report readiness for a check it was unable to perform.**

This line asks "which checks failed?". With three states it needs to ask "which checks did not succeed?". `UNKNOWN` means *could not check*, and readiness cannot be concluded from that.

Two local checks return `UNKNOWN`: line 105 (the interpreter probe itself failed) and line 167 (`screamingface` is not importable). Both indicate a broken or unsynced venv, which is an ordinary condition. In that state the script prints `LOCAL LANE READY` and exits 0.

The outcome is worse than the skip case above. A skip is silent but harmless; here the run would fail at import, while the script tells the reader to expect "5 xfailed, 0 failed" — an outcome that cannot happen.

Line 229 already treats `UNKNOWN` as something worth acting on, which makes the two lines inconsistent:

```python
229:  if c.remedy and c.state != OK:                       # UNKNOWN -> print the remedy
240:  blocked = [c for c in local if c.state == MISSING]   # UNKNOWN -> not blocking
```

The output shows the contradiction directly. The remedy prints, and then the ready line prints underneath it:

```
UNKNOWN  client has OME-967  screamingface not importable in ...
                             -> cd packages/screamingface && uv sync --extra runtime
...
LOCAL LANE READY.
```

```suggestion
    blocked = [c for c in local if c.state != OK]
```

The bastion's `UNKNOWN` is unaffected by this change — it sits in the k8s lane, which per D3 never touches the exit code. That decision is documented and sound; this is only about the two local-lane paths.

IonesioJunior (commented):
Two issues in the readiness logic. Both make the script report "LOCAL LANE READY" in a state where the lane is not ready — which is the failure mode the script exists to prevent. Both reproduce on a clean checkout; details inline.

The design decisions themselves (D1 not probing the bastion, D2 `find_spec` over real imports, D3 local-lane-only exit code) are sound and well documented, and the "a skipped rung proves nothing" warning is the right thing to say. These are two narrow logic bugs, not a problem with the approach.

sergio-bershadsky:
@IonesioJunior Both bugs were real, both reproduced, both fixed. Thank you — the first one had already cost me time and I'd misdiagnosed it.

**R1 — the fix went into the test, not the validator.** You called this correctly: `test_boards.py:63-71` and `test_failures.py:365-372` both default to `default_data_dir() / "benchmark-assets"` with an explicit OME-1001 comment, and `test_correlation_chain.py` was the sole outlier defaulting to `FIXTURES_DIR / "assets"` — a path `prepare` never writes and the repo doesn't ship. Aligning the validator to the ladder would have codified the outlier and left the ladder needing a hand-set override forever.

Worth saying plainly: **this bug was already biting and I read it as an environment quirk.** While landing OME-1119 I ran the ladder as `SCREAMINGFACE_E2E_ASSETS=$PWD/tests/e2e/fixtures/assets/benchmark-assets` because `prepare --data-dir` wrote one level deeper than the test looked, shrugged, and moved on. That was this defect, worked around instead of fixed.

**R2 — `UNKNOWN` now blocks.** `blocked = [c for c in local if c.state != OK]`, and the summary distinguishes the two: `LOCAL LANE NOT READY — 2 missing, 2 unverifiable`. Your framing of *could-not-check* vs *failed* is now the comment on that line.

**R3 — `SCREAMINGFACE_DATA_DIR`** is honoured, and structurally rather than by copying the rule: the validator asks the client for its data dir through the venv interpreter it already probes, so the two can't drift again.

**R4 — the printed command echoes the override** when one is set, so what's printed is what was validated.

**R5, found while fixing** — the printed expectation was stale. It promised `5 xfailed, 0 failed`, untrue since rungs 1 and 2 went green (OME-1121, OME-1119). A validator stating a wrong expected outcome trains the reader to ignore a real regression. Now `2 passed, 3 xfailed, 0 failed`, naming both tickets.

**Verified end to end rather than reasoned about.** With the venv synced and **no override set**, the validator reports five `OK`, and running the command it prints *verbatim* gives `2 passed, 3 xfailed`. Before the fix that same command skipped all five. The unsynced-venv path was exercised too: `exit=1`, "2 missing, 2 unverifiable".

`run_gates.py screamingface --skip-append-only` — ALL GREEN. Append-only flagged the ladder for the fourth consecutive time; benign by count: test functions 5 → 5, assertions 14 → 14, xfail markers 10 → 10, change confined to the `_assets_root` helper.

**One thing I did not fix, and it's why both bugs shipped:** `check_setup.py` has no tests and no CI gate — nothing under `.github/workflows/` matches `e2e/failor/**`. Both were pure logic in an ungated file, so nothing but a human reading it could have caught them. It's a standalone tool outside any uv project, so giving it a test home is a real decision rather than a detail; I'd rather raise it than quietly bolt something on. Happy to file it if you agree it's worth doing.

PR #835, opened 2026-09-03
fix(screamingface): retry replay-safe requests on transient edge failures

IonesioJunior:
## Summary

A single Cloudflare **520** on one `POST /token` ended an `sf.evaluate` of 8 candidates — after 7 had already completed **100/100 cases at 100% cache hit**. The origin was healthy the whole time: nginx logged 14 `/token` requests, all 200, including three seconds either side of the failure. The blip lived above the origin, in a flapping tunnel connection, and was environmental.

This PR is about why one blip was fatal.

- **Nothing retried a request the SDK had itself declared replay-safe.** `/token` is sent with `extensions={_REPLAY_SAFE: True}` and was never replayed.
- **`_raise_response` pasted the raw body** into the message, parsing only `problem+json` — so an edge proxy's HTML page reached the user as ~7KB of markup with the status code buried inside.

## The design decision worth reviewing

Retry is gated on the **existing `_REPLAY_SAFE` marker, never on the HTTP method.**

That marker is already default-deny, and `_core/wire.py` says why: *"`GET /?q=` starts billable work despite being a GET and is never marked… a call site that forgets the marker pays one extra round trip — never an extra paid Run."* Replay safety is a property of the REQUEST — does re-sending duplicate a side effect — not of the response status, so one marker legitimately answers both "safe to re-send after an Access login" and "safe to re-send after a transient 5xx".

The consequence is that the dangerous case is excluded **by construction**, not by a rule the next contributor must remember:

| call | `_REPLAY_SAFE` | retried |
|---|---|---|
| `POST /token` (mint) | True | yes — the call that failed |
| `DELETE /` (stop) | True | yes, idempotent |
| **`GET /?q=` (run start)** | **absent → False** | **never** |

Implemented as an httpx transport so the safety policy lives in one place and cannot be forgotten by a future call site.

Two smaller judgement calls:
- **500 is deliberately NOT retryable** — an application error is deterministic; repeating it wastes the caller's time and hides the defect. The set is the gateway/edge family plus 408/429/503.
- **`Retry-After` beyond a 30s cap stops rather than sleeping** — obeying an hour-long value is indistinguishable from a hang, so the response is surfaced and the caller decides.

## What this deliberately does NOT fix

A refused **run start** is still not retried, and must not be — retrying it would double-fire billable Runs. So this does not address:

- the **in-flight reservation leak** (a caller stays pinned at the cap for 16.3h because release depends on `status()`, which the SDK's WebSocket path never calls, and on a reaper that is disabled by default in the local rig);
- the **all-or-nothing evaluation semantics** (`_evaluation/runner.py:382-389`), where one candidate's failure cancels the siblings and discards their completed work.

Both remain open follow-ups, and together they are why a single refusal still costs a whole evaluation.

## Test plan

29 new tests in `tests/test_transient_retry.py`. The first two pin the load-bearing invariant — that an unmarked request is sent exactly once — in both sync and async form, because violating it double-fires paid work.

- retryable statuses retried; 4xx and 500 returned immediately
- attempt budget bounded; backoff grows between attempts
- `Retry-After` honoured in delta-seconds; beyond the cap it stops rather than sleeping
- transport errors retried, and the final failure still raises
- an HTML error page reduces to its status; a short plain-text reason is kept; `problem+json` detail still wins

`run_gates.py screamingface` — ALL GATES GREEN (append-only · ruff · pyright · pytest --cov ≥95 · notebooks · uv build · distribution). The append-only check passes **without** `--skip-append-only`: this adds a new test file and touches no prior test.

Ledger: `docs/work/2026-09-03-OME-1107-sdk-transient-retry.md`

Refs: OME-1107

### DIFF
--- a/docs/tasks/2026-09-03-OME-1107-sdk-transient-retry.md
+++ b/docs/tasks/2026-09-03-OME-1107-sdk-transient-retry.md
@@ -1,0 +1,32 @@
+---
+id: OME-1107
+linear_url: https://linear.app/openmined/issue/OME-1107/retry-replay-safe-sdk-requests-on-transient-edge-failures
+status: in_progress
+type: null
+priority: 2
+labels: [py-screamingface, agentic, autonomous]
+created: 2026-09-03
+closed:
+---
+
+# Retry replay-safe SDK requests on transient edge failures
+
+A single Cloudflare **520** on one `POST /token` ended an `sf.evaluate` of 8 candidates — after
+7 had already completed **100/100 cases at 100% cache hit**. The origin was healthy throughout:
+nginx logged 14 `/token` requests, all 200, including 3 seconds either side of the failure. The
+blip lived above the origin, in a flapping tunnel connection, and is environmental.
+
+This unit is about why one blip was fatal:
+
+1. **Nothing retried a request the SDK had itself declared replay-safe.** `/token` is sent with
+   `extensions={_REPLAY_SAFE: True}` and was never replayed.
+2. **The error pasted the raw body** — only `problem+json` was parsed, so ~7KB of Cloudflare
+   markup reached the user with the status code buried inside it.
+
+Retry is gated on the EXISTING `_REPLAY_SAFE` marker, never on the HTTP method. That marker is
+already default-deny precisely because `GET /?q=` starts billable work despite being a GET — so
+the one dangerous call is excluded by construction rather than by a rule someone must remember.
+
+**Explicitly NOT fixed here:** the in-flight reservation leak and the all-or-nothing evaluation
+semantics. Run start is deliberately not replay-safe, so this retry does not — and must not —
+cover a refused run start.
--- a/docs/work/2026-09-03-OME-1107-sdk-transient-retry.md
+++ b/docs/work/2026-09-03-OME-1107-sdk-transient-retry.md
@@ -1,0 +1,117 @@
+---
+ticket: OME-1107
+stack: screamingface
+status: in_progress
+started: 2026-09-03
+finished:
+---
+
+# OME-1107 — retry replay-safe SDK requests on transient edge failures
+
+## Intent
+
+A single Cloudflare 520 on one `POST /token` destroyed an `sf.evaluate` of 8 candidates, after
+7 had already completed 100/100 cases at 100% cache hit. The origin was healthy throughout —
+nginx logged 14 `/token` requests, all 200, including 3 seconds either side of the failure —
+so the 520 came from a flapping tunnel connection above the origin. The blip was environmental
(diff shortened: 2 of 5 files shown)


keelancj (changes requested):
Looking good, just retry.py needs a bit more test coverage to pass the gate (88%, CI shows 94.94% total vs 95% required). Specifically missing: the HTTP-date form of Retry-After and the attempts < 1 ValueError guard

IonesioJunior:
@keelancj Thanks for the review — pushed `ace79e08` closing both gaps you flagged:

- HTTP-date form of `Retry-After` (aware + naive, RFC 9110 §10.2.3)
- the `attempts < 1` `ValueError` guard, for both the sync and async transports

Plus one more I found while closing these: the async twin of the `Retry-After`-beyond-cap test existed only on the sync side.

No production code changed — every gap was a missing test for behavior that was already correct. `run_gates.py screamingface` is green, including the coverage gate (package back over 95%; `_core/retry.py` alone moved 88% → 94% — the only remaining misses are two provably-unreachable `AssertionError` safety nets in the retry loop, left uncovered on purpose rather than chased with a `# pragma: no cover`).

CI is green across all 12 checks. Ready for another look whenever you have a chance.

keelancj:
Thanks @IonesioJunior those added test cover all the gaps. I found one separate issue, if a 503 response’s body raises httpx.ReadError, response.read() exits before closing the response or retrying. I reproduced this in both sync and async: only one attempt happens even when the next response would succeed

IonesioJunior:
@keelancj thanks for the sharp catch — confirmed and fixed in `bbe77a43`.

The `ReadError` escaped because the `except TransportError` guarded only the *send*: the read-and-release before re-sending sat outside it, so a connection that died mid-body propagated raw — zero retries, and a pooled connection never closed. `response.read()`/`aread()` now run inside `_release_and_wait` / `_release_and_wait_async`, where a body that fails to arrive is the same transport failure as a dropped send: close the dead response, back off (never the `Retry-After` that rode on a response whose body never arrived), re-send.

One thing worth knowing: the loop returns a retryable response **unread** on its last attempt so the caller sees the terminal response intact — which made my first `if last: raise` inside the helper provably unreachable, so it's gone (invariant documented in the docstring) rather than covered by a test that only passed through httpx's own `Client.send()` body read.

3 new tests in `tests/test_transient_retry.py` — all fail against `ace79e08` with the exact one-attempt `ReadError` you reproduced, all pass now. `run_gates.py screamingface` is green (append-only included; retry.py back to 98%).

Ready for another look whenever you have a chance.

keelancj (approved):
Looking good!!

PR #840, opened 2026-09-04
feat(scoreboard): make private rollback fail closed

FilipBolt:
## Summary

- replace the misleading package-semver rollback floor with Helm revision, rendered image, and immutable runtime image-digest evidence
- add a dry-run-by-default `purge_private_benchmark` operator CLI that requires a matching canonical export SHA-256 and deletes one exact private board atomically
- make the rollback runbook executable through export, checksum, purge, `SAFE` preflight, zero pods/endpoints, rollback, and replica restoration

## Safety properties

- unknown or public benchmarks refuse
- benchmarks with baselines refuse
- an export mismatch or any failed final delete rolls the whole transaction back
- score ingestion and purge lock the same benchmark row, preventing a new submission between digest verification and deletion on PostgreSQL
- the command is database-only, has no HTTP surface or bulk mode, and requires `--yes` after a successful dry run

## Verification

- `run_gates.py scoreboard --base origin/main --skip-append-only` — all gates green
- Python: 610 passed, 3 skipped, 3 deselected; coverage gate ≥80%
- portal: 50 passed across the three explicit Node suites
- ruff lint/format and pyright clean

The append-only check is skipped under the owner's explicit Confidence-Gate approval because the prior test asserting package semver was the rollback floor had to be corrected; both safe and unsafe images package version `0.1.1`.

## Operational closure

This PR completes the code and runbook portion of OME-1027. The Linear issue should remain In Progress until an operator records the production Helm revision/image digest and runs the preflight inside the production Scoreboard pod. This workstation has no Scoreboard cluster context, so this PR does not claim that evidence.

Refs: OME-1027

### DIFF
--- a/apps/scoreboard/DEPLOYMENT.md
+++ b/apps/scoreboard/DEPLOYMENT.md
@@ -272,10 +272,34 @@ revision's hooks (`execHook(targetRelease, release.HookPreRollback, ...)` in Hel
 stored manifest has no such hook, so nothing added to this chart can run in the dangerous
 direction.
 
-Before **any** `helm rollback`, run the preflight in a pod on the current release:
+Package version is not a release identity here: both the privacy-blind base and the
+privacy-aware release package `scoreboard 0.1.1`. For every known-safe deployment, record the Helm
+revision, rendered image reference, and runtime `imageID` digest together in the deployment record:
 
 ```bash
-kubectl -n scoreboard exec deploy/scoreboard-scoreboard -- \
+NAMESPACE=scoreboard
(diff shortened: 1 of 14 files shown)


HupBaHa (approved):
Reviewed exact head 7f8b27b058e3f6371439e09646453bad123d87f8 against OME-1027 rollback/private-board purge criteria. No blocking findings found. Local isolated verification passed: Python 3.12 focused rollback/purge tests and PostgreSQL lock guard, Pyright, Ruff check, and Ruff format check.

PR #841, opened 2026-09-04
feat(scoreboard): require run cost on direct submissions

FilipBolt:
## Summary

- require a non-null `run_cost_usd` on direct `ScoreSubmission` requests
- return a field-local `422` for both omission and explicit JSON `null`, and mark the field required in OpenAPI
- keep the database column and every read DTO nullable for imported and historical rows
- update the deployment smoke request and Client guide to match the tightened contract

## Why now

The producer gate is satisfied: OME-303, OME-304, OME-306, and OME-1029 are Done, and the SDK now publishes the run total. Zero remains a valid, distinct value for a genuinely free cached run; this change never coerces missing cost to zero.

## Deliberately unchanged

- no migration or backfill
- no change to the OME-770 normalization/limits
- no cost in recipe identity
- no mutation of legacy null-cost rows on dedup
- an upstream `None` now fails direct publish instead of inventing a cost

## Verification

- Scoreboard gates all green: ruff, format, pyright, coverage ≥80%, Python suite, and the three portal suites
- Python: 602 passed, 3 skipped, 3 deselected
- portal: 50 passed
- public docs: clean `npm ci`, oxlint, eslint, Vue typecheck, and production Vite build

The append-only check is skipped under the owner-approved Confidence-Gate exception: tightening a required request field necessarily updates existing direct-submission fixtures and replaces the prior test asserting omission becomes `None`. Legacy-null read behavior remains explicitly tested through controlled historical-row fixtures.

PR #840's newly added direct-submission fixtures have already been made cost-ready, so the two independent PRs can merge in either order without breaking `main`.

Refs: OME-822

### DIFF
--- a/apps/scoreboard/src/scoreboard/scores/schemas.py
+++ b/apps/scoreboard/src/scoreboard/scores/schemas.py
@@ -24,7 +24,7 @@ COST_CEILING = Decimal("999999.999999")
 _ZERO_COST = Decimal("0").quantize(COST_QUANTUM)
 
 
-def _validate_run_cost(value: Decimal | None) -> Decimal | None:
+def _validate_run_cost(value: Decimal) -> Decimal:
     """Reject a cost that cannot be stored; normalize every cost that can.
 
     Only two things are actually unstorable and therefore rejected: a negative or
@@ -45,8 +45,6 @@ def _validate_run_cost(value: Decimal | None) -> Decimal | None:
     Those constraints run BEFORE this validator, so they would reject the very
     values it exists to normalize.
     """
-    if value is None:
-        return None
     # ge=0 on the field already rejects negatives, and NaN fails that comparison,
     # but +Infinity passes it — and quantize() raises InvalidOperation rather than
     # returning a value, so non-finites have to go before any arithmetic.
@@ -324,17 +322,11 @@ class ScoreSubmission(BaseModel):
     # (D-SCORE-006). Persisted onto the flat client_* columns by the store.
     client: ClientInfo | None = None
     metadata: dict[str, Any] | None = None
-    # INVARIANT: absent (None) means "no cost was reported" and is NOT the same as
-    # 0. A fully cache-served run genuinely costing nothing is a legitimate 0, so
-    # OME-770's Pareto frontier must exclude None rather than rank it as the
-    # cheapest entry. Decimal, not float — this is money.
-    #
-    # AIDEV-NOTE: optional only because nothing emits a run cost yet (OME-303 is
-    # unmerged, the Engine does not roll per-call cost into a run total, and the
-    # Client has no field for it), and because the column lands on an already
-    # populated table. Once a client can send it, a direct submission arriving
-    # without one is a client bug and should be REJECTED — null then means
-    # "imported or legacy" only. Tracked on OME-770.
+    # INVARIANT (OME-822): every direct submission reports a cost. A fully
+    # cache-served run genuinely costing nothing is represented by 0; omission or
+    # null is a client bug and is rejected by this non-nullable required field.
+    # Database and read DTOs deliberately remain nullable because imported and
+    # legacy rows can still have no known cost. Decimal, not float — this is money.
     # INVARIANT: the request contract mirrors the column exactly — DECIMAL(12, 6).
     # `ge=0` alone let three failures through, each reproduced live:
     #   0.0000009 -> accepted (201) and silently stored as 0.000001, publishing a
@@ -350,11 +342,11 @@ class ScoreSubmission(BaseModel):
     # requires us to quantize and accept. `ge=0` stays here (it also rejects NaN,
     # which fails the comparison); allow_inf_nan=False stops +Infinity, which
     # would pass ge=0 and then raise inside quantize().
-    run_cost_usd: Decimal | None = Field(default=None, ge=0, allow_inf_nan=False)
+    run_cost_usd: Decimal = Field(ge=0, allow_inf_nan=False)
 
     @field_validator("run_cost_usd")
     @classmethod
-    def validate_run_cost(cls, value: Decimal | None) -> Decimal | None:
+    def validate_run_cost(cls, value: Decimal) -> Decimal:
         return _validate_run_cost(value)
 
     @field_validator("url4_expression")
--- a/apps/scoreboard/DEPLOYMENT.md
+++ b/apps/scoreboard/DEPLOYMENT.md
@@ -170,7 +170,7 @@ Submit a smoke score with an idempotency key. **`benchmark_id` must name a bench
 curl -fsS -X POST http://scoreboard.40.76.107.241.nip.io/v1/scores \
   -H "Content-Type: application/json" \
   -H "Idempotency-Key: score-007-smoke-1" \
-  -d '{"version":1,"benchmark_id":"smoke","spec_id":"score-007-smoke","url4_expression":"url4://smoke","submitted_by":"score-007","score":0.5,"total_questions":2,"ran_with_providers":["smoke"],"client":{"name":"curl","version":"0.1.0","platform":"k3s"}}'
+  -d '{"version":1,"benchmark_id":"smoke","spec_id":"score-007-smoke","url4_expression":"url4://smoke","submitted_by":"score-007","score":0.5,"total_questions":2,"ran_with_providers":["smoke"],"run_cost_usd":"0.000000","client":{"name":"curl","version":"0.1.0","platform":"k3s"}}'
 
 curl -fsS http://scoreboard.40.76.107.241.nip.io/v1/leaderboard/smoke
 ```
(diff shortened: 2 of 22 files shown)


HupBaHa (approved):
Reviewed against OME-822. The direct submission contract now requires a non-null run cost while preserving nullable historical and imported reads, genuine zero-cost semantics, existing normalization, and dedup behavior. No blocking findings.

keelancj (approved):
Good to go

apps/scoreboard/src/scoreboard/scores/schemas.py
```diff
@@ -414,7 +427,38 @@ class ScoreSubmission(BaseModel):
...
+    # the model validator below enforces that pairing. Omitting both is still rejected.
     run_cost_usd: Decimal | None = Field(default=None, ge=0, allow_inf_nan=False)
+    # INVARIANT (OME-1251 D4): a RUN-level vocabulary, deliberately not the gateway's per-call
+    # `DirectCostStatus`. A run has many calls, and no member of that vocabulary can express
+    # "forty priced, three not" — the common case and the one that matters.
+    #
+    # WHY a required field rather than a nullable one: this is the whole point of OME-822. A
+    # client that can determine its cost and says nothing is a client bug; making the status
+    # required is what turns that silence into a 422 instead of a null the board must interpret.
+    run_cost_status: RunCostStatus
```
HupBaHa: [P1] Preserve the deployed-client path before making this field required. The current `origin/main` producer at `packages/screamingface/src/screamingface/_scoreboard/leaderboards.py:428-461` sends `run_cost_usd` but no `run_cost_status`, and this PR explicitly has to deploy before the OME-1252 client because older boards forbid unknown fields. With this required field, that documented order cannot work: immediately after the Scoreboard deploy, every existing SDK submission gets a 422, including payloads carrying a valid cost (reproduced against this merge result on Pydantic 2.13.5). Please introduce a rollout-compatible phase/default/inference for old payloads, or otherwise provide a deploy sequence that does not interrupt submissions; the client update cannot safely go first either.

apps/scoreboard/src/scoreboard/scores/store.py
```diff
@@ -204,6 +215,19 @@ def _replay_updates(submission: ScoreSubmission, existing: Score) -> dict[str, o
...
+    # leave a row `complete` with a null amount, or an amount whose status says it is
+    # unknowable — the exact incoherence the request validator refuses, reached through the
+    # back door. `_content_hash` excludes both, so a replay carrying a cost dedups to the
+    # stored row and would otherwise discard it silently (OME-770 D8).
+    #
+    # INVARIANT: FILL ONLY, never replace, for the reason `models` records above. A published
+    # cost is a frontier position; a replay must not be able to move one.
+    if existing.run_cost_status is None:
+        updates["run_cost_status"] = submission.run_cost_status
+        updates["run_cost_usd"] = submission.run_cost_usd
```
HupBaHa: [P1] Do not treat a null status as proof that the amount is unfilled. Migration 0014 deliberately leaves `run_cost_status` null on every existing row, including rows with a non-null published `run_cost_usd`. This branch then replaces both fields on the first same-owner replay: an incoming `unavailable` replay erases an existing amount, while a `complete` replay with a different execution cost moves the published frontier position. A direct probe with an existing `9.000000` amount/null status and an incoming `unavailable` submission returns `{run_cost_status: unavailable, run_cost_usd: None}`. The new cannot-move-cost test only creates a post-migration row whose status is already `complete`, so it misses this production state. Please preserve/backfill existing priced amounts and add the migrated priced-row replay case.

HupBaHa (changes requested):
Requesting changes for two cross-version/data-integrity blockers introduced after the previously approved head. The full Scoreboard gates and migration smoke are green on the synthetic merge with current main, but those tests only construct new-contract payloads and new rows; they do not cover the deployed SDK or migrated priced rows.

FilipBolt:
Both fixed in `c4346e02`. Both findings were correct, and both were mine.

## P1-1 — the deploy order

You're right that it could not work, and the sharper problem is that **this PR's own body prescribed it**. I designed the requirement around "silence is rejected" and never checked what the live client sends. `leaderboards.py:445` on `main` is the only cost line there — an amount, no status.

Fixed as **expand/contract**, split across two tickets:

| phase | contract |
|---|---|
| **this PR** (OME-822) | `run_cost_status` **optional**. Absent beside an amount resolves to `complete`. Absent with no amount stays absent |
| OME-1252 | the SDK sends the status explicitly |
| **OME-1258** (filed) | flip to required; silence becomes a 422 |

The resolution is deliberately not a guess — **an amount IS the claim `complete` makes.** It happens in the model validator rather than the store, so the submission object, the stored row and the response all carry the same fact, and pre-OME-1252 clients produce correctly labelled rows instead of a population OME-1258 would have to clean up afterwards.

Absent-with-no-amount stays `null` rather than becoming `unavailable`: the board doesn't know whether the client looked, and claiming it did would be inventing evidence. That's the row OME-1258 starts refusing.

## P1-2 — the wrong sentinel

Confirmed exactly as you describe. Gating on `existing.run_cost_status is None` treats every migrated row as empty, because `0014` leaves the status null on all of them — including rows with real published money.

**The amount is the sentinel; the status is a label on it.** Now:

- both null → fill both from the submission
- amount present, status null → **heal the label to `complete`, never touch the money**

The heal is sound for the same reason the inference above is: a stored amount is itself the claim. It also means the population OME-1258 inherits is already correct rather than needing a backfill later.

Your point about my test missing this is the part worth recording — it built a post-migration row whose status was already `complete`, which is not the production state. Four tests added, including the one that reproduces yours: a `9.000000` amount with a null status, an incoming `unavailable` replay, asserting the amount survives.

## Three of my round-1 tests were rewritten

They asserted the required contract this round proves undeployable. None is a prior test on `main`:

- `test_a_submission_without_a_status_is_rejected` → `…_resolves_it_from_the_amount`, plus a new neither-field case
- `test_post_score_requires_a_cost_or_a_reason_it_is_absent` → `…_still_accepts_a_deployed_client_payload`
- the OpenAPI assertion now pins that both cost fields are **published but not required**

## Deviation, stated plainly

**Fix preceded test on both findings.** You supplied the reproductions, so a failing case existed before the change — but by you, not as a committed RED from me. The migrated-row test does fail against the previous code; I confirmed that by reasoning through the old branch rather than by reverting and running it. Recorded in the ledger rather than presented as TDD.

## Verification

`run_gates.py scoreboard --base origin/main --skip-append-only` — ALL GREEN. Full suite **722 passed / 3 skipped / 3 deselected**. CI on this head: 20 checks, 0 failures.

The deploy note in the PR body is now accurate: this is safe to deploy against the live SDK, and OME-1252 is what makes the status mean anything.

docs/spec/2026-09-21-OME-822-cost-status.md
```diff
@@ -0,0 +1,71 @@
...
+| known | every component priced | amount + `complete` | the amount | `complete` |
+| unknowable | run happened, cost not derivable | `partial` or `unavailable` | `null` | that member |
+| no cost concept | imported baseline, pre-`OME-770` row | n/a — not a submission | `null` | `null` |
+
+Today the last two collapse into a bare `null`, which is why the frontier cannot tell a row it
+should exclude from a row that never had a cost. The stored status is what separates them.
+
+## §2 The wire contract
+
+`run_cost_usd` is **required on the request**. `run_cost_status` is **required** alongside it.
```
HupBaHa: [P2] Update the governing contract to describe the expand phase that this fix now implements. `ScoreSubmission` deliberately makes both fields optional: amount-without-status resolves to `complete`, and neither field is accepted until OME-1258. This section still says both are required, while the task mirror (`What ships`), plan (RED/GREEN steps), Linear OME-822 scope/checked acceptance, PR summary, and the stale `schemas.py:410-414` comment still say silence is rejected by this PR. Those are the authoritative rollout contract, not historical notes, so merging them as written would document a different API from the one deployed and can cause the next phase to assume the 422 gate is already active. Please rewrite those surfaces around the explicit expand/client/contract sequence and move the required/422 acceptance entirely to OME-1258.

HupBaHa (changes requested):
The two previous P1 findings are fixed correctly: the deployed SDK path now resolves an omitted status beside an amount to complete, and migrated priced rows preserve their amount while healing the missing label. Focused tests and the full Scoreboard gate suite pass on a clean synthetic merge with current main. One P2 remains: the governing spec, task/plan, Linear acceptance, PR summary, and a source comment still describe status as required and silence as rejected, which contradicts the intentional expand-phase API on this head. Please align those contract surfaces with OME-1258 before merge; the inline comment identifies the exact locations and required sequence.

HupBaHa (approved):
Reconsidered the remaining documentation mismatch: it is valid follow-up work, but not a runtime, rollout, or data-integrity blocker. Both prior P1 findings are fixed correctly, the exact deployed-client and migrated-row regressions are covered, the synthetic merge is clean, and focused plus full Scoreboard gates pass. Approved at c4346e02; the existing inline note is now explicitly non-blocking.

PR #842, opened 2026-09-04
feat(screamingface): carry the run's trace id onto the candidate result (OME-1121)

sergio-bershadsky:
## Summary

First child of the Phase 1 epic (`OME-1118`). **Flips rung 1 of the correlation ladder green — `1 passed, 4 xfailed`.**

`OME-967` put the client-minted trace id on the **error hierarchy only**. But a board run *does not raise*: DRACO fans out with `on_error="collect"`, so case failures become rows and the run returns a `Report` normally. The result was backwards from what the roadmap wants — the user whose run produced **bad results rather than an exception**, the person most likely to file a report, was precisely the one who could not obtain an id to quote.

## The field goes on `CandidateResult`, not `Report`

The ticket title says "on Report". That's the wrong shape, and the PR deviates deliberately.

A `Report` holds several **independently executed runs**, each with its own client-minted trace — which is exactly why `CandidateResult` already carries a per-run `run_id`. A single `Report.trace_id` would have to pick one, and there's no defensible choice. One run, one trace, one field, beside the field that already identifies that run.

`test_two_candidates_in_one_report_keep_their_own_trace_ids` exists specifically to pin this.

The common single-candidate case reads `report.candidates.only.trace_id`.

## The fixture bug this exposed

Rung 1 kept failing *after* the fix landed. The cause was in the ladder's own fixture from `OME-1105`:

```
RAISED PlanningError | trace_id = None
msg: Model 'openrouter/anthropic/claude-haiku-4.5' is not available on this Engine
```

That model isn't in the synthetic tape's catalog projection, so `evaluate` raised in the availability probe (`runner.py::_missing_required_models`) **before the transport ever ran**. No transport means no trace context — so every rung read an empty id set, and rung 1 could never have passed whatever this change did. Repointed at `openrouter/openai/gpt-5.5`, which the tape carries.

**The ladder was reporting a real absence for the wrong reason.** Worth knowing before the next rung is debugged.

## Deliberately not done

**`to_dict()` does not gain the field.** It carries `run_id`, and the exported artifact arguably should carry the trace id too — it's the "attach this to a report" surface. But `to_dict` output is compared against e2e **golden fixtures**, which are prior tests, so changing it risks a regression outside this unit's scope for a benefit the ticket doesn't require. Follow-up.

**No `Report.trace_ids` convenience.** Considered for the multi-candidate SigNoz paste, but it's a second public surface for a case nobody has hit, and `tuple(c.trace_id for c in report.candidates)` is one line. YAGNI.

## Rule 5 — owner-approved test edits

Three files flagged. **Approved on the numbers**, which are the point:

- `test_correlation_chain.py`: 52+/46− churn, but assertions **12 → 12** and test functions **5 → 5**. The delta is one xfail marker, a docstring rewrite, a fixture refactor (extracting `_one_run` to satisfy ruff's `PLR0915` — restructured, never suppressed), and the model constant.
- `test_report.py`: purely additive — +47, 3 new tests, 0 removed.
- `public_surface_snapshot.json`: regenerated via the sanctioned `UPDATE_SURFACE_SNAPSHOT=1` path; additive only (keyword with a default, so no caller breaks).

**Worth deciding separately:** deleting an xfail marker *is* a test edit, so all four remaining rung PRs will trip this same gate for the same benign reason. Exempting the ladder file beats approving the identical shape four more times.

## Test plan

- RED confirmed first: `TypeError: CandidateResult.__init__() got an unexpected keyword argument 'trace_id'`.
- `run_gates.py screamingface --skip-append-only` — **ALL GATES GREEN**.
- Ladder: **1 passed, 4 xfailed**.

Ledger: `docs/work/2026-09-04-OME-1121-report-trace-id.md`
Mirror: `docs/tasks/2026-09-04-OME-1121-report-trace-id.md`

Refs: OME-1121

### DIFF
--- a/docs/tasks/2026-09-04-OME-1121-report-trace-id.md
+++ b/docs/tasks/2026-09-04-OME-1121-report-trace-id.md
@@ -1,0 +1,29 @@
+---
+id: OME-1121
+linear_url: https://linear.app/openmined/issue/OME-1121/expose-trace-id-on-report-so-a-completed-run-can-be-quoted
+status: in_progress
+type: null
+priority: 2
+labels: [py-screamingface, agentic, autonomous]
+created: 2026-09-04
+closed:
+---
+
+# Expose trace_id on the public result so a completed run can be quoted
+
+First child of the Phase 1 epic (`OME-1118`). `OME-967` put the client-minted trace id on the
+error hierarchy only — but a board run does not raise (DRACO collects case errors into rows),
+so the user with bad results returned a clean `Report` carrying no id anywhere.
+
+`CandidateResult.trace_id` closes it. **The field is per-candidate, not per-Report**: a
+Report holds several independently executed runs, each with its own trace, exactly as it
+already holds a per-run `run_id`. A single `Report.trace_id` would have to pick one.
+
+Rung 1 of the correlation ladder is now green (**1 passed, 4 xfailed**).
+
+Found while closing it: the ladder's own fixture named a model the synthetic tape does not
+carry, so `evaluate` raised `PlanningError` in the availability probe *before the transport
+ran* — no transport, no trace, and rung 1 could never have passed. A defect in `OME-1105`,
+surfaced only by trying to make its rung pass.
+
+Ledger: `docs/work/2026-09-04-OME-1121-report-trace-id.md`.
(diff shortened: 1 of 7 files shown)


keelancj (approved):
clean change + good catch on that model name in the test

PR #843, opened 2026-09-04
feat(scoreboard): explain revision-mismatch submissions

FilipBolt:
## Summary
- return a machine-readable benchmark revision mismatch notice while still storing and accepting the score
- expose the notice as an immutable Client value and render an accessible not-ranked warning in the published-score card
- preserve matching submission output, GET/private-export output, and the existing leaderboard revision filter
- compose with the author fields merged in #830

## Verification
- Scoreboard official gates: all green, including append-only, coverage, and portal suites
- ScreamingFace official gates: all green, including 95% coverage, notebooks, build, and distribution verification
- owner-approved skip-append-only used only for the deliberately regenerated public API surface snapshot; existing behavioral tests were not changed

Refs: OME-909

### DIFF
--- a/apps/scoreboard/src/scoreboard/routes/scores.py
+++ b/apps/scoreboard/src/scoreboard/routes/scores.py
@@ -37,6 +37,7 @@ from scoreboard.scores.schemas import (
     FieldErrorDetail,
     FieldErrorResponse,
     MessageErrorResponse,
+    ScoreRankingNotice,
     ScoreSchema,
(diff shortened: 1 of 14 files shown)


HupBaHa (approved):
Looks good. Revision-mismatched submissions remain successful and persisted while returning a typed notice with both revisions. The Client surfaces the not-ranked state, matching and read responses remain unchanged, and the existing ranking filter is preserved.

PR #848, opened 2026-09-07
docs(tasks): close the five stale observability mirrors and ledgers (OME-1133)

sergio-bershadsky:
Closes the repo-side status for five observability items that shipped but whose mirrors and ledgers were never closed.

## Why

Linear is the status authority and `docs/tasks/` is its repo-side reflection. On `main` today, `OME-967`, `OME-990`, `OME-1074`, `OME-1105` and `OME-1121` are all **Done** in Linear but all ten of their mirror and ledger files still read `status: in_progress` with an empty `finished:`. Anyone reading the checkout without Linear access — the normal case — sees five observability items apparently still in flight, while Phase 1 (`OME-1118`) is actively being built on top of them. That is how a duplicate ticket gets filed for merged work.

The root cause is worth naming rather than quietly fixing: the mirror close belongs in each shipping PR and was missed five consecutive times. That makes it a process defect, not five slips.

## What changed

Frontmatter only — `status: in_progress` → `done`, and `finished:`/`closed:` → the PR's merge date.

| Item | Merged | PR |
|---|---|---|
| `OME-967` | 2026-09-02 | #789 |
| `OME-990` | 2026-09-02 | #780 |
| `OME-1074` | 2026-09-02 | #812 |
| `OME-1105` | 2026-09-03 | #831 |
| `OME-1121` | 2026-09-07 | #842 |

Plus this unit's own ledger and mirror, closed in the same commit — deferring them would reproduce the exact defect being fixed.

## Test plan

No code changes, so no stack gate applies and CI's path filters run nothing. The claim is a state one and was checked mechanically rather than by eye:

- A checker listing every `docs/tasks` + `docs/work` file for a Linear-Done item that still reads `in_progress`: **10 stale before, 0 after.**
- `git diff -U0` confirms **20 changed lines, every one a `status:`/`finished:`/`closed:` field** — no prose or ledger Outcome content altered.

One thing worth flagging: the checker's first version used `finished:\s*(\S*)`, where `\s*` crosses the newline on an empty field and captures the `---` fence below it — so a file marked `done` with **no date** would have reported clean. Caught on the baseline run, fixed to `[ \t]*`, baseline re-run before any edit.

## Not in scope

Adding a gate that enforces mirror closure at merge. It is worth doing, but `run_gates.py` cannot see Linear state, so where the check lives (PR template · CI job against the Linear API · a close-time step in the SDLC skill) is a decision rather than a detail. Left to its own item.

Refs: OME-1133

### DIFF
--- a/docs/tasks/2026-08-24-OME-967-client-originates-traceparent.md
+++ b/docs/tasks/2026-08-24-OME-967-client-originates-traceparent.md
@@ -1,12 +1,12 @@
 ---
 id: OME-967
 linear_url: https://linear.app/openmined/issue/OME-967/originate-the-traceparent-in-the-client-and-surface-trace-id-to-the
-status: in_progress
+status: done
 type: null
 priority: 2
 labels: [py-screamingface, agentic, autonomous]
 created: 2026-08-24
-closed:
+closed: 2026-09-02
 ---
 
 # Originate the traceparent in the client and surface `trace_id` to the user
--- a/docs/tasks/2026-08-25-OME-990-runtime-log-prompts.md
+++ b/docs/tasks/2026-08-25-OME-990-runtime-log-prompts.md
@@ -1,12 +1,12 @@
 ---
 id: OME-990
 linear_url: https://linear.app/openmined/issue/OME-990/runtimelog-records-user-prompts-in-cleartext
-status: in_progress
+status: done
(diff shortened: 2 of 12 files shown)


IonesioJunior (approved):
LGTM!

PR #849, opened 2026-09-07
feat(screamingface-engine): propagate the run's traceparent to aigateway (OME-1119)

sergio-bershadsky:
Rung 2 of the correlation ladder. Second child of the Phase 1 epic (`OME-1118`), after `OME-1121`.

## The problem

The engine **adopted** a caller's traceparent — the control plane forwards it into the runner as `job_env.TRACEPARENT` and the run's own frames carry it — and then sent nothing onward. The full header set aigateway received during the audit:

`Host, Accept, Accept-Encoding, Connection, User-Agent, X-User-Email, X-Profile, Content-Length, Content-Type`

The trace stopped at the engine, so a user quoting the `trace_id` from their `Report` (`OME-1121`) could see the engine's half of a run and nothing about the model calls — which is where the interesting failures are.

## Two sources, because there are two kinds of caller

**Run mode** — a ContextVar bound inside the executor's driving task, from the `TraceContext` that `url4.streaming.lifecycle.run` hands to `Executor.execute`.

Deliberately **not** `logs.RunContext.trace_id`, which is the tempting one-line answer and is wrong:

```python
# runner/main.py:477
trace_id = parse_traceparent(traceparent)   # None when the caller sent none
with run_scope(params.topic, trace_id):
```

When the caller sends no traceparent, url4 **mints** one and that id never reaches `RunContext`. Reading it would propagate for client-originated runs — passing rung 2 — and silently propagate nothing for every other run. A hole shaped exactly like a passing test. There is a test for that case specifically.

**Connections** — carried on `Caller`, which already *is* the per-request object, so no middleware and no new global. One `_headers(caller)` at `_request` covers **every** connections call, and `/v1/providers` gains the `X-Profile` it had also been missing (the same header-assembly gap the issue bundles).

## Why the scope binds inside `_drive`

`Url4Executor.execute` is an **async generator**, and consecutive steps of one can be driven from different contexts — `asyncio.ensure_future(gen.__anext__())` runs that step in a new Task with a **copied** context. A ContextVar token set in one and reset in another raises `ValueError: Token was created in a different Context`.

The first version wrapped `execute`'s `yield` and did exactly that. It was caught by an **existing** test (`test_runner_summary.py::test_a_cancelled_run_records_stopped`), not by a new one — the append-only rule earning its keep. The driving task is one context for its whole life and is where every model call actually happens. Restated as a named regression test so the constraint is findable next time.

## The catalog path deliberately sends nothing

This is the one place the issue's "three client paths" premise did not survive contact with the code, so it is flagged rather than quietly satisfied.

`CachedCatalog` coalesces concurrent misses for one credential onto a single upstream fetch (`cache.py`'s `_inflight`) and serves the result to every waiter. A catalog call is therefore *caused by* whichever caller led the refresh and *consumed by* N others. Stamping the leader's trace would put a call in their trace that also served people they never heard of, and leave the other N-1 with a gap where a catalog read should be — both readings wrong, and the wrong one looks right.

Sending nothing is the honest answer for Phase 1, whose payoff is grepping one id across services for one run. Representing a coalesced fetch properly needs an OTel span Link, which is `OME-1130`.

## Test plan

**Engine** — `run_gates.py screamingface-engine`: **ALL GREEN** (append-only, ruff, ruff format, pyright, check_layering, `pytest --cov --cov-fail-under=80`) — **2366 passed, 5 skipped**.

17 new tests in `tests/unit/test_traceparent_propagation.py`, including the boundaries an e2e run cannot reach:

- a run whose caller sent **no** inbound traceparent still propagates the id url4 minted;
- **two runs on one cached world do not share a trace id** — the regression a constructor argument would cause, since `_resolve_world` caches `self._io`;
- absent is **absent** — no header, never a zero or invented one that would parse everywhere and join nothing;
- the gateway-owned headers still win over an inbound mapping that carries its own `X-Profile`/`traceparent`;
- a malformed inbound traceparent degrades to absent, never forwarded and never an error;
- a cancelled run does not raise out of the trace scope.

**Client** — `run_gates.py screamingface --skip-append-only`: **ALL GREEN**.

**The ladder**, against the real engine and a real gateway subprocess: **2 passed, 3 xfailed** — rung 2 flipped green, rungs 3/4a/4b still fail as designed.

## Append-only gate — third consecutive rung PR

Deleting rung 2's strict-xfail marker modifies a prior test, so the gate flags it. Benign and verified by count: test functions **5 → 5**, assertions **14 → 14**, one marker removed, no assertion weakened.

This is the third PR to trip it for the identical reason and there are two more rungs to go. Standing recommendation: exempt `test_correlation_chain.py` from the check rather than approve the same shape twice more.

## Not visible in SigNoz yet

This puts the id on the wire. Nothing logs it until `OME-940` (engine side) and `OME-1120` (gateway side) land — see the parent epic for what becomes validatable after deploy.

Refs: OME-1119

### DIFF
--- a/apps/screamingface-engine/src/screamingface_engine/catalog/aigateway.py
+++ b/apps/screamingface-engine/src/screamingface_engine/catalog/aigateway.py
@@ -199,17 +199,25 @@ def _headers(credential: Credential) -> dict[str, str]:
     No ``Authorization``: a deployed aigateway (``cloudflare_headers``) reads only the identity
     header, and a local one (``disabled``) reads nothing at all.
 
-    WHY no ``traceparent`` here, unlike the other two aigateway clients (OME-1119): this call is
-    NOT one-to-one with a caller. ``CachedCatalog`` coalesces concurrent misses for one
-    credential onto a single upstream fetch (``cache.py``'s ``_inflight``), and serves the result
-    to every waiter — so a catalog fetch is caused by whichever caller happened to lead the
-    refresh and is consumed by N others. Stamping the leader's trace would put a call in their
-    trace that also served people they never heard of, and leave the other N-1 with a gap where
-    a catalog read should be. Both readings are wrong, and the wrong one looks right.
+    AIDEV-NOTE (OME-1119): this function carries NO ``traceparent``, and the three callers do not
+    all want that for the same reason — so do not "fix" it in one place.
 
-    Sending nothing is the honest answer for Phase 1, whose payoff is grepping ONE id across
-    services for ONE run. Representing a coalesced fetch properly needs an OTel span Link (one
-    span, many causes), which is Phase 2 — see ``OME-1130``.
+    - ``fetch`` genuinely must not carry one. ``CachedCatalog`` coalesces concurrent misses for a
+      credential onto a single upstream fetch (``cache.py``'s ``_inflight``, typed
+      ``Future[ModelCatalog]``) and serves the result to every waiter, so that call is *caused by*
+      whichever caller led the refresh and *consumed by* N others. Stamping the leader's trace
+      would put a call in their trace that also served people they never heard of, and leave the
+      other N-1 with a gap where a catalog read should be — both readings wrong, and the wrong one
+      looks right. Representing it honestly needs an OTel span Link (one span, many causes), which
+      is Phase 2, ``OME-1130``.
+    - ``fetch_model_parameters`` and ``admit_model`` are NOT coalesced — ``_inflight`` covers only
+      ``fetch``, and ``CachedCatalog.model_parameter_source`` is documented as "the uncached detail
+      source". ``rest/catalog.py`` calls the first straight from a route, one upstream call per
+      inbound request. Those two SHOULD carry the caller's traceparent; they do not yet, because
+      the value has to be threaded from the REST edge through ``catalog/executable.py`` rather than
+      taken from ``Credential`` — ``Credential`` holds a derived cache ``key``, so a per-request
+      field on it would give every request its own cache entry and destroy the catalog cache.
+      Tracked as ``OME-1134``.
     """
     headers = dict(credential.identity)
     if credential.profile is not None:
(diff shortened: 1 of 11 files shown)


keelancj (approved):
@sergio-bershadsky Looks good, approving. One non-blocking question, I see why shared cached requests are excluded but are uncached model-parameter requests intentionally deferred too?

IonesioJunior (approved):

sergio-bershadsky:
@keelancj Good catch — not intentionally deferred, and my exclusion argument was wrong for exactly the calls you're asking about.

I justified skipping the catalog path on `CachedCatalog` coalescing, then wrote that as a blanket claim about the whole file. It only holds for `fetch`:

- `_inflight` is typed `dict[str, asyncio.Future[ModelCatalog]]` — it covers `fetch` and nothing else.
- `CachedCatalog.model_parameter_source` is documented in its own docstring as "the uncached detail source".
- `rest/catalog.py:190` calls `fetch_model_parameters` straight from a route — one upstream call per inbound request.
- `admit_model` isn't in `cache.py` at all; it's reached from `catalog/executable.py:222`.

So those two are one-to-one with a caller and should carry the traceparent. They don't in this PR.

I've split the reasoning at the call site (0aaa875, comment only, no behaviour change) rather than leave a note that reads as "the catalog path is settled", and filed the actual propagation as **OME-1134**, blocked by this PR.

Worth flagging why it isn't a quick follow-on: the tempting fix is to add `traceparent` to `Credential` the way this PR added it to `Caller`, and that's a trap — `Credential` carries a derived cache `key`, so a per-request field on it gives every request its own cache entry and silently destroys the catalog cache. The value has to be threaded from the REST edge through `catalog/executable.py` instead, with `_headers` taking an explicit optional parameter so `fetch`'s abstention stays visible at its call site rather than depending on a scope not being bound. OME-1134 carries that, plus a test asserting the cache key is unchanged when only the traceparent differs.

keelancj:
@sergio-bershadsky  Cool thanks for checking and adding it to OME-1134. Looks like we're good to merge this one in that case

PR #853, opened 2026-09-08
fix(scoreboard): publish unambiguous author credits

FilipBolt:
## Summary

- collapse repeated author identities case-insensitively when publishing while preserving the exact submitted list in storage and staff exports
- show domains only when distinct authors on the same row share a local part
- enforce the ten-distinct-person cap plus a 4 KiB canonical serialized-field bound
- record the Python client contract alignment separately in OME-1139

**Asana:** N/A — tracked in Linear as OME-1109

## Components touched

- `apps/scoreboard`
- `docs/spec`, `docs/plan`, `docs/tasks`, and `docs/work`

## Test plan

- [x] `uv run .claude/scripts/run_gates.py scoreboard --base origin/main`
- [x] append-only guard, Ruff, formatting, Pyright, coverage, and all three portal suites green
- [x] full Scoreboard suite: 621 passed, 3 skipped, 3 deselected
- [ ] screenshots / recording attached — not applicable; serialization and validation only

## Cross-service notes

The Python client still rejects raw author lists longer than ten before HTTP. OME-1139 tracks aligning it with this distinct-person and 4 KiB contract; no client code is included here.

This is independent of the still-unmerged #841. Both touch the Scoreboard submission schema and tests, so whichever lands second may need a small mechanical rebase, but their contracts do not conflict.

Fixes OME-1109

### DIFF
--- a/apps/scoreboard/src/scoreboard/scores/schemas.py
+++ b/apps/scoreboard/src/scoreboard/scores/schemas.py
@@ -223,22 +223,50 @@ SubmittedBy = Annotated[
     PlainSerializer(_publish_submitter, return_type=str | None, when_used="json"),
 ]
 
+# INVARIANT: counting distinct people must not turn repeated author entries into
+# an unbounded public write. This matches the established metadata envelope.
+_AUTHORS_MAX_BYTES = 4096
+_AUTHORS_MAX_DISTINCT = 10
+
+
+def _author_identity(author: str) -> str:
+    """The one comparison identity for author validation and publication."""
+    return author.casefold()
+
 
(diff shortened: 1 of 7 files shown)


keelancj (approved):
Approved

- Traced the dedup/collision logic against the new tests including a 3-way local-part collision 
- Confirmed AuthorEmail's validation means every author string reaching _publish_submitter is already well-formed
- 10 max-length authors serialize to ~2.6 KB, well under the new 4 KiB limit so no legitimate submission can be rejected by it

Looks like a merge conflict appeared which needs resolved but good for merge other than that

PR #859, opened 2026-09-08
feat(scoreboard): drop the Submitter column from the leaderboard table

FilipBolt:
## Summary

- remove the `Submitter` column from the benchmark leaderboard table (`portal/benchmark.js`) — it duplicated `Authors` for every submission without an explicit author list, because `_resolved_authors` falls back to `[submitted_by]`
- portal display only: `submitted_by` is unchanged on the model, the API, the private JSONL export, and the spec detail page
- add a guard that pins the removal without pinning the two wrong ways to satisfy it

**Asana:** N/A — tracked in Linear as OME-1144

## Components touched

- `apps/scoreboard` (portal + one test file)
- `docs/spec`, `docs/plan`, `docs/tasks`, `docs/work`

## Two deviations from the ticket text

1. The ticket says `formatSubmitter` "becomes unused; drop it if no other caller". **There is another caller** — `portal/spec.js:17` renders the submitter on each spec's history table — so `portal/main.js` is untouched.
2. The ticket says to update "any portal tests asserting the Submitter header/column". Nothing under `tests/portal/` asserts it; the assertions live in `tests/unit/test_multiple_authors.py`, which is where the edit landed.

## Accepted trade-off, recorded in the spec

`submitted_by` is resolved from verified Cloudflare Access identity; `authors` is a client-supplied list that nothing verifies. Removing Submitter leaves the public table showing only the unverified claim. The owner reviewed this and chose the removal as filed (2026-09-08) — spec D5 records it so the next reader does not rediscover it as a defect. The verified value stays on the spec page and the API.

## Test plan

- [x] `uv run .claude/scripts/run_gates.py scoreboard --base origin/main --skip-append-only` → ALL GATES GREEN (ruff check, ruff format, pyright, pytest `--cov-fail-under=80`, all three portal suites)
- [x] new `test_portal_leaderboard_table_has_no_submitter_column` written first and confirmed RED for the right reason
- [x] mutation-checked, each mutation reverted and the file confirmed byte-identical afterwards:

| Mutation of `benchmark.js` | Result |
| --- | --- |
| restore the header without its cell | CAUGHT |
| restore the cell without its header | CAUGHT |
| delete the Authors cell too — both identity columns gone | CAUGHT |

The third is why the guard carries presence assertions as well as absence ones: `COLUMNS` and `renderRows` are positional, and a bare "Submitter is absent" check would pass on a board that lost every identity column.

- [ ] screenshots — not attached; the change is the removal of one column from an existing table

## Confidence-Gate exception (sdlc rule 5)

`test_portal_renders_author_lists_in_leaderboard_and_history` asserted the deleted column config and cell render as source text (old lines 197–198). Those two assertions describe the behaviour this PR deliberately removes, so they cannot survive it; the owner's decision to remove the column as filed entails the edit. Its other five assertions are unchanged, and the new guard covers the area, so coverage does not shrink.

`run_gates.py` without `--skip-append-only` flags exactly those two lines and nothing else.

## Cross-service notes

None. No API, schema, or client change.

### DIFF
--- a/apps/scoreboard/portal/benchmark.js
+++ b/apps/scoreboard/portal/benchmark.js
@@ -31,7 +31,6 @@
     // identities, and providers.length > 1 is not a valid fusion/solo test.
     // Keeping the honest label until a backend field exists.
     { key: "ran_with_providers", label: "Backends", sort: null },
-    { key: "submitted_by", label: "Submitter", sort: "string", dir: "asc" },
     { key: "authors", label: "Authors", sort: "string", dir: "asc" },
     { key: "score", label: "Score", sort: "number", dir: "desc", cls: "num" },
     // WHY Questions is gone: OME-769's column list is #, Name, Models, Author,
@@ -212,7 +211,6 @@
       tr.appendChild(specTd);
 
       tr.appendChild(P.el("td", null, P.formatProviders(entry.ran_with_providers)));
-      tr.appendChild(P.el("td", null, P.formatSubmitter(entry.submitted_by)));
       tr.appendChild(P.el("td", null, P.formatAuthors(entry.authors)));
       tr.appendChild(renderScoreCell(entry.score, barMin, barMax));
       // WHY the title: the cell rounds to cents, but the frontier compares the full stored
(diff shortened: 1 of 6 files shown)


keelancj (approved):
clean removal looks good to merge @FilipBolt

PR #865, opened 2026-09-08
Ome 1148 contracteval

siddhant230:
## Summary

Onboards [ContractEval](https://arxiv.org/abs/2508.03080) (NLLP 2025) as an Engine-owned board: 4,182 clause-extraction questions over 102 real commercial contracts (CUAD test split). The model gets a full contract plus one of 41 clause categories and must quote the answering sentences verbatim or reply `"No related clause."` — graded by deterministic string containment, so **no judge and zero grading tokens**. Reading the paper's MIT-licensed harness corrected our July screening note: ContractEval is **not** span-overlap with partial credit, it is all-or-nothing verbatim containment, and its F1/F2 come from a **dataset-level confusion matrix** rather than a mean of case scores.

**Asana:** n/a — technical work never goes to Asana (CLAUDE.md rule 7). Work item: `OME-1148`.

## Components touched

- `apps/screamingface-engine` — new `benchmarks/contracteval/` board (9 modules), registered in `builtins.py`; one additive row in `test_benchmark_declaration.py`
- `packages/screamingface` — `_BENCHMARKS` + asset manifest in `_runtime/cli.py`, `build_notebooks.py`, generated `examples/12_contracteval.ipynb`
- `docs/` — spec, plan, work ledger, task mirror

## Test plan

- [x] Ran the touched app's local gates — `run_gates.py screamingface-engine` **ALL GREEN** (2,667 tests, 92% coverage) and `run_gates.py screamingface` **ALL GREEN** (≥95% coverage, notebook + `uv build` + distribution checks)
- [ ] Gateway request/refresh change → n/a, no gateway change
- [ ] Screenshots / recording attached → n/a, no UI change
- [x] 68 new ContractEval unit tests (grading · prepare · case-evaluation · aggregate)
- [x] **Differential test against the reference harness** — 2,970 (output, gold) pairs over 400 real CUAD rows: **zero** mismatches in verdict, abstention and Jaccard
- [x] Assets baked end to end — 4,182 cases, 1,244 positive, 2,938 negative (the 1,244 is the exact constant the paper's `Evaluation.py` hardcodes, confirming we bake the exam it scored)
- [x] Live runs on the local stack — solo `limit=8` F1 0.80 and fusion `limit=5` F1 0.75, coverage 1.0, no failures. Every published metric reproduces by hand from the per-case list.

## Cross-service notes

Spans `apps/screamingface-engine` and `packages/screamingface`. Filed as **one ticket** rather than a D9 epic + sub-issues, following the GDPval (OME-971) and MedXpertQA (OME-1126) precedent — the SDK side is a three-line registration plus a generated notebook. Deliberate, flag if you'd rather split it.

**Reviewer notes:**

1. **F1, not accuracy, is the headline score.** 70.3% of rows have no clause, so a model that always abstains scores ~70% accuracy while answering nothing. F1 scores it 0. Accuracy is still reported as a metric.
2. **Five deviations from the reference, all deliberate and tested:** a zero guard on `2PR/(P+R)` (the reference raises `ZeroDivisionError` for an always-abstaining model); laziness divided by *selected* positives rather than the hardcoded `1244`, identical at full split; `temperature` omitted from the notebook because `openai/gpt-5.5` answers **404** and `claude-opus-4.8` **400** for any value; the full 4,182 rows vs the paper's 4,128; and the system-prompt item below.
3. **⚠️ System prompt is sent as user text.** The reference uses `role: "system"`; `candidate()` accepts a single `input` string and gives boards no system-role channel. **MedXpertQA has the same constraint and handles it worse** — it defines `ANSWER_SYSTEM` and never uses it, silently dropping the official system prompt on a merged board. Suggest a shared ticket rather than two per-board workarounds.
4. **One prior test modified** — a single additive row in the `expected` table of `test_every_builtin_board_declares_its_actual_policy`, which the test's own note requires each new board to add. Committed with `--skip-append-only` after owner approval.
5. **A gap worth generalising:** `prepare.py` initially shipped without a `main()`, and the bake failed with `prepared output is missing [...]` — symptom named, cause hidden. `test_the_family_guard_covers_every_family_preparer_package` checks the preparer *command string* against a regex, not that the module answers it. Pinned per-board here; a shared test would be better.
6. **Naming:** "span-extraction" in the Linear title is stale — the code says `method: "containment"`. Worth renaming the issue and the doc files together.


keelancj (changes requested):
Nice work @siddhant230 just 2 things came up in review:

runtime.py:99 returns candidate inputs without calling preflight(). Reproduced with a missing answer file: the contract is served, inference runs, then grading discovers the missing key

aggregate.py:406 excludes failed questions from the laziness denominator but the spec includes them. With two questions that both have a clause (one incorrect abstention and one failed request) it reports 100% instead of 50%. Either calculation can be intentional the code and spec just need to agree

khoaguin:
let's onboard this according to the new abstraction layer according to [this docs here](https://github.com/ScreamingFace/screamingface/blob/main/apps/screamingface-engine/docs/adding-a-benchmark.md)

PR #870, opened 2026-09-09
feat(py-screamingface): guard ifeval and gdpval-text with e2e replay goldens

khoaguin:
Puts the last two unguarded boards — ifeval (the only deterministic-grading board, pinned as a **CorrectiveLoop**, a V1 recipe) and gdpval-text (fusion) — under the e2e replay guard: one paid recording each, replayed for $0 on every PR.

## Problem / Solution

**Problem:** two of our leaderboard numbers had no regression guard — a code change that silently shifted an ifeval or gdpval-text score would pass CI green. Worse, no existing recording route could even capture an ifeval CorrectiveLoop run: multi-round conversations were unreplayable.

**Solution:** teach the bless harness a third route — replay a *fresh* recording exactly as recorded, byte for byte — and use it to freeze one real run of each board as its permanent answer key. CI now re-runs both boards from those frozen answers, free, on every PR.

### Before
- Trigger: the spine folds ifeval (deterministic grading) and gdpval-text onto shared code (OME-1101 / OME-1097 chain).
- Today: only draco-3pass and healthbench-worst30 have goldens. The folds would ride on unit tests alone; a grading change could shift a published number with CI green. No bless route accepts a CorrectiveLoop recording at all.
- Cost: a silently wrong leaderboard number — the worst failure this product can have.

### After
- Same trigger.
- Now: **both boards have blessed goldens in this PR.** ifeval pins the `CorrectiveLoop([ministral, phi], judge=deepseek-v4-flash)` run — score 0.9184, coverage 0.98, one real `model_token_cap` judge failure pinned by its code; gdpval-text pins the `open_panel` fusion — score 0.8044, coverage 0.8, five provider-error cases pinned as recorded. `test_boards[ifeval]` replays keylessly in ~11s, `[gdpval-text]` in ~15s.
- Win: all four grading paths (rubric solo, rubric fusion, deterministic loop, gdpval rubric) are guarded before their folds.

### Don't regress
- Existing goldens (draco-3pass, healthbench-worst30) untouched — the golden schema growth is default-valid for old files.
- Spend stays impossible in CI: no provider keys; a cache miss is a loud failure.
- The dump/answers and report bless routes behave exactly as before.

## What the harness gained

1. **Fresh-dump bless mode** (`--dump-fresh` + `--candidate`, `just e2e-bless-fresh`): a recording made through the *current* gateway already carries correct cache keys, so this mode skips re-keying and tape synthesis entirely — load dump → keyless verified replay → slice → golden. It never inspects request bodies, which is what makes any candidate shape (a multi-round loop included) replayable. It refuses unless the replay reproduces the report's score/coverage/statuses **and** rendered expression.
2. **A third golden kind, `corrective_loop`**: unlike fusion goldens (routes only — report mode re-keys), a fresh-dump replay must render byte-identical requests, so the golden stores each member and the judge as a full spec (route + prompt + params) and `build_candidate` rebuilds the exact `sf.CorrectiveLoop`.
3. **gdpval-text board registration** + the two fixture triples + per-board manifest whitelist lines in `.gitignore`.

Two engine-side defects this work surfaced were fixed on main mid-flight and this branch is rebased on them: OME-1167 (model parameter metadata needed a connected provider — the keyless replay's preflight now just works; an interim SDK escape hatch was added and then fully deleted) and OME-1168 (loop judge prompts embedded token accounting, making recordings unreplayable — the ifeval recording was redone after the fix).

## Review order

1. **`packages/screamingface/tests/e2e/harness/goldens.py`** — the contract's home: the frozen answer key learns the `corrective_loop` kind. Verify the new spec/validator defaults leave every pre-existing golden valid, and that `build_candidate`'s loop branch carries prompt and params through — the byte-identity requirement the PR pivots on.
2. **`packages/screamingface/tests/e2e/fixtures/slice_snapshot.py`** — the bless tool: the fresh-dump flow (parse seams → boot → seed → verified replay → slice) and the `author_golden` loop shape. Verify every mismatch refuses rather than writing a golden, and the three modes stay mutually exclusive.
3. **`packages/screamingface/tests/e2e/test_fresh_dump_contracts.py`** — the spec of 1–2, default lane. Check the refusal tests name real failure modes, not just "it raised".
4. **The two fixture triples** (`fixtures/goldens/*.golden.json`, `fixtures/snapshots/*.snapshot.gz` + `.manifest.json`) — spot-check the goldens' pinned numbers against the descriptions above; the snapshot headers cite the owner-held recordings by content SHA, never by path.
5. **`packages/screamingface/tests/e2e/test_boards.py`** — gdpval-text joins the board list; the golden-driven candidate build is unchanged otherwise.
6. **`packages/screamingface/justfile` + `tests/e2e/README.md` + `.gitignore`** — owner-facing commands and the manifest whitelist; mechanical skim.
7. **`docs/work/`, `docs/tasks/`** — ledger (the full two-day story, deviations disclosed) + closed mirror.

**Genuinely new runtime behavior:** only the bless tool's new owner-run mode. What CI runs gained is two more replayed boards.

Refs: OME-1098


khoaguin:
Status: the fixtures are not in yet. First bless attempt surfaced two findings, both now tracked:

- The SDK's free model-parameters preflight cannot pass on the sealed keyless stack, so params-carrying candidates (which a corrective_loop golden must be) could never replay. This PR now carries an env-gated skip (`SCREAMINGFACE_SKIP_PARAMETER_PREFLIGHT=1`, set only by the replay harness; two tests pin that only the literal `1` disarms). Principled replacement: OME-1167.
- The bless then refused correctly: all 39 single-round cases replay byte-identically, all 9 multi-round cases miss, because the engine embeds runtime token accounting inside coach/tie prompts (live counts vs cache-hit zeros → different cache keys). Engine fix: OME-1168, which blocks the ifeval bless; after it lands, one owner re-record of the 50-case loop run and `just e2e-bless-fresh` produce the fixtures.

gdpval-text is unaffected (report-mode bless, recording still pending).

khoaguin:
**ifeval golden landed** (commit `5d470bdf`): the re-recorded 50-case CorrectiveLoop run (post-OME-1168 minimal coach prompts) blessed cleanly on the first try — the fresh-dump replay reproduced score 0.9184 / coverage 0.98 / all 50 statuses / the recorded expression, and `test_boards[ifeval]` now replays keylessly in ~11s. The golden pins one real `model_token_cap` judge failure (present in both recordings — a genuine property of this candidate).

Remaining before ready-for-review: the gdpval-text fusion recording + `just e2e-bless-report` (owner-gated, independent).

keelancj (commented):
@khoaguin two things came up in review, one blocking:

- Blocking: --expect-score can bypass the saved report check. Fror example blessing a replay scoring 0.8 against a report scoring 0.9 by passing --expect-score 0.8

- Non-blocking: Refreshing the ifeval golden fails after replay because _refresh_golden doesn’t pass the member specs, judge spec, and round limit to author_golden

khoaguin:
@keelancj both addressed in `374dfc8f`:

- **Blocking:** the report-backed modes (`--report` and `--dump-fresh`) now **refuse** `--expect-score`/`--expect-coverage` outright — the saved report is the only outcome authority there (`pin_report_expectations`, one shared function for both modes). The flags remain valid only on the `--dump/--answers` path, where no report exists and they are the sole cross-check.
- **Non-blocking:** `_refresh_golden` now builds its `author_golden` arguments through `replay_input_fields` — a single kind-complete mapping — so a corrective_loop refresh carries its member/judge specs and `max_rounds` through instead of refusing validation after the replay.

Three contract tests pin both (flag refusal, report pinning, loop-spec round-trip through a refresh); gates green.

PR #877, opened 2026-09-09
feat(url4): add safe node-scoped structured Log emission

keelancj:
Long-running ensemble operations need to explain what is happening while the caller waits. This PR adds the URL4 reporting hook needed for structured activity such as completion timing or cache status to reach a future activity UI. Producers and client rendering remain separate work; this change does not generate or display that activity by itself.

## Example: reporting a retry inside a node

Illustrative instrumentation only; this change does not implement retry detection, a production activity schema or UI rendering.

While an observed URL4 node is resolving, adapter code that has detected a retry could call:

```python
from url4.observe import current_log_sink

sink = current_log_sink()
if sink is not None:
    sink("Retrying request", {"attempt": 2}, severity="WARN")
```

The executor binds the sink in the node's async execution context before calling resolve. The accessor retrieves that binding, so the adapter does not pass a node ID or register this message against a URL4 function. URL4 emits a Log containing the currently resolving node's span ID, the message and the attributes. The sink expires when resolution ends. Attribution is to that DAG node; distinguishing multiple model calls within one node requires separate application-defined fields.

The complete path is: adapter explicitly reports the retry → URL4 observer receives the node-attached Log → Engine forwarding from `OME-934` carries it to the existing Log stream → later client rendering can display it. URL4 does not discover retries automatically, and the attribute name `attempt` has no built-in meaning to URL4.

The earlier Engine-only proposal could carry the same information through a separate run-scoped input. This approach reuses URL4 observation and provides automatic node attachment; it does not make previously impossible information available. Evaluation-wide progress and scoring state remain Engine-owned under `OME-932`.

## Why this belongs in URL4

URL4 already owns node observation, Logs and task-local usage/response sinks. An adapter can now report a message with explicit fields, automatically attached to the currently resolving node, without receiving the whole ExecutionContext through its call chain. This keeps concurrent node activity distinguishable and lets consumers read fields instead of parsing prose. A node span identifies the observation context, not exact model-call or benchmark Case ownership.

OME-934 builds on this hook to forward attributes through the existing Engine Log stream and safely buffer optional activity around essential execution events. Reusing the existing observation path avoids a second Engine-specific logging input, factory or queue. Only three production modules change; no URL grammar, graph identity, scoring, external dependency or transport change is introduced.

## Why this is a generic URL4 capability

Adapters are an existing URL4 extension mechanism: the `IOLayer` protocol defines `fetch(...)`, with HTTP and in-memory implementations supplied in separate modules. Applications can supply their own implementations. Code implementing that interface can emit structured observations while it runs within an observed node's resolution on the event-loop thread. For example, an HTTP adapter could report bytes fetched or request duration; these illustrate possible instrumentation, not producers added by this PR. URL4 attaches the observation to the node without interpreting those fields or knowing the adapter's implementation.

This is architecturally consistent with URL4's existing separation of execution, I/O implementations and observers, and follows the existing task-local usage/response sink pattern. Validation, expiry and best-effort failure handling are new observation policies specified by the approved design in [PR #876](https://github.com/ScreamingFace/screamingface/pull/876); they are not requirements already imposed by URL4's grammar.

URL4 supplies the part it already owns: the active node's observation context and lifetime. It attaches a Log to that node, validates the generic record shape, snapshots attributes and expires the safe emitter when resolution ends. Centralizing this avoids each integration implementing its own task-local attribution and cleanup. Structured fields preserve machine-readable facts without requiring consumers to parse message prose.

The dependency direction stays outward: producer adapters use the URL4 interface, and a caller supplies an Observer implementing URL4's protocol. URL4 does not import the Engine, a producer or an exporter, and does not select a destination. An embedder can consume observations in-process without the ScreamingFace Engine, network transport or UI. The added observation code uses only the Python standard library.

Attribute names and meanings belong to producers and consumers. URL4 does not require or interpret model names, cache fields, benchmark Case IDs, scores or any other application vocabulary, and does not branch execution based on those attributes. Domain schemas, privacy filtering, payload/rate limits, delivery and rendering remain outside this interface.

“Agnostic” therefore means application- and exporter-agnostic, not policy-free or dependency-free for consumers. Producers choosing the accessor depend on URL4's observation API. URL4 deliberately specifies node scope, synchronous same-thread emission, flat scalar attributes, accepted severities and best-effort failure behavior. This is an additive library observation contract; it is not a change to URL grammar or a universal logging system outside URL4 execution.

## Implementation

Adapters inside a resolving URL4 node can now obtain `current_log_sink()` and emit structured Logs attached to that node's span. Attributes are immutable snapshots, and the sink drops invalid, expired, off-thread or ordinarily failing submissions without changing the node result. Existing direct logging and observer failure behavior remain intact.

Implements the package portion of the approved design in #876. Child-task bindings are revoked on scope exit; observed nesting binds/restores its own sink, while unobserved nesting inherits an active outer. Failure and cancellation expire the sink before terminal observation. Accepted severity names and flat scalar validation are documented in the package README.

## Scope and limitations

Logs are optional, best-effort activity. Invalid, expired or off-thread calls and ordinary observer failures may silently lose a message. A blocking observer can still slow execution. Reliable completion, accounting and saved failure diagnostics use separate authoritative paths.

URL4 observation and executor wiring only. OME-934 separately owns Engine forwarding and safe bridge buffering. Concrete activity schemas, Client rendering and benchmark progress remain their own issues. Keep this PR draft.

## Validation

- 45 new tests covering real DAG lifecycle, concurrent/nested scopes, child expiry, off-thread calls, immutable snapshots, validation, observer errors and process-control propagation.
- Full `uv run .claude/scripts/run_gates.py url4` passed: append-only, Ruff lint/format, Pyright and full pytest coverage gate. 1,215 tests collected; 98% overall coverage and 99% in observe.py.
- No existing tests modified; no live provider calls.

Asana: not applicable.

Refs: OME-1165

### DIFF
--- a/packages/url4/src/url4/observe.py
+++ b/packages/url4/src/url4/observe.py
@@ -11,18 +11,21 @@ downstream tracing product is deliberately kept out of this module.
 ``Observer.on_event`` is synchronous and non-blocking by contract — the
 executor calls it inline from its own coroutines, never behind a task or a
 queue, so a slow or blocking observer would slow the run itself. An observer
-that raises is not caught anywhere in the engine (an embedder bug should be
-loud, not swallowed): the exception propagates out of the run exactly like
-any other node failure.
+that raises propagates out of direct observation like any other node failure.
+The opt-in ``current_log_sink`` convenience alone contains ordinary submission
+failures; it never changes lifecycle or direct logging failure semantics.
 """
 
 from __future__ import annotations
 
 import contextlib
 import contextvars
-from collections.abc import Callable, Iterator
-from dataclasses import dataclass
+import math
+import threading
+from collections.abc import Callable, Iterator, Mapping
+from dataclasses import dataclass, field
 from decimal import Decimal
+from types import MappingProxyType
 from typing import Literal, Protocol, runtime_checkable
 
 
@@ -56,11 +59,19 @@ class NodeFinished:
     permanent: bool | None = None
 
 
+type LogScalar = str | int | float | bool | None
+
+
 @dataclass(frozen=True, slots=True)
 class Log:
     span_id: str | None
     severity: str
     body: str
+    attributes: Mapping[str, LogScalar] = field(default_factory=dict, hash=False)
+
+    def __post_init__(self) -> None:
+        # INVARIANT: observers and caller mutation cannot rewrite queued evidence.
+        object.__setattr__(self, "attributes", MappingProxyType(dict(self.attributes)))
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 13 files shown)


packages/url4/src/url4/observe.py
```diff
@@ -56,11 +59,19 @@ class NodeFinished:
...
 @dataclass(frozen=True, slots=True)
 class Log:
     span_id: str | None
     severity: str
     body: str
+    attributes: Mapping[str, LogScalar] = field(default_factory=dict, hash=False)
+
+    def __post_init__(self) -> None:
+        # INVARIANT: observers and caller mutation cannot rewrite queued evidence.
+        object.__setattr__(self, "attributes", MappingProxyType(dict(self.attributes)))
```
IonesioJunior: **`Log` can no longer be pickled, deep-copied, or passed to `asdict`**

`MappingProxyType` has no reducer, so this line breaks three things for *every* `Log` — including plain 3-argument construction with no attributes.

```python
Log("s", "INFO", "b")
# pickle.dumps        → TypeError: cannot pickle 'mappingproxy' object
# copy.deepcopy       → TypeError
# dataclasses.asdict  → TypeError
```

I verified against the base commit: all three worked at `1a53690c`. `Log` is now the only member of the public `ObservationEvent` union that fails — `NodeStarted`, `Usage`, `ModelResponse` and `RunFinished` all still pickle.

Nothing in this repo pickles observation events today, so it isn't breaking us yet. But `Log` is exported, and an embedder who fans events to a worker process or calls `asdict(event)` to build JSON gets an error naming a type they never used.

This keeps immutability and repairs all three at once:

```python
def __reduce__(self):
    return (self.__class__, (self.span_id, self.severity, self.body, dict(self.attributes)))
```

Worth a round-trip test with and without attributes — nothing covers this right now.

packages/url4/src/url4/observe.py
```diff
@@ -222,35 +233,131 @@ def current_response_sink() -> ResponseSink | None:
...
+                return
+            normalized = severity.strip().upper()
+            if normalized not in ("DEBUG", "INFO", "WARN", "ERROR"):
+                return
+            self._emit(normalized, body, attributes=_log_attributes(attributes))
+        except Exception:
+            # WHY: only this opt-in emission path is fail-open. No diagnostic
+            # logging here: it could recurse, leak payload or fail through a handler.
+            # BaseException (cancellation/process control) deliberately propagates.
+            pass
```
IonesioJunior: **A permanently broken producer is completely invisible**

Every drop is silent — bad severity, bad scalar, off-thread, expired, observer raising. No counter, no signal.

```python
sink("api call failed", severity="WARNING")   # not in the 4-value whitelist
# dropped, forever, silently — indistinguishable from "the adapter never called the sink"
```

`WARNING` is a very plausible first-integration mistake for OME-1161, and whoever debugs it has nothing to look at.

The spec clause is conditional ("*Any* internal warning is bounded and contains only phase/exception type"), so silence satisfies it vacuously. But required-verification item 3 asks for tests of "payload-free bounded diagnostics", and the tests here assert the opposite (`caplog.text == ""`).

A monotonic drop counter keyed by phase (`body` / `severity` / `attributes` / `emit`) is bounded, payload-free, can't recurse and can't raise. Fine by me as a follow-up ticket rather than this PR — I'd just rather it wasn't an undocumented absence.

keelancj:
Both concerns are addressed: Log serialization works while preserving immutable attributes, and bounded, payload-free counters make dropped submissions diagnosable

IonesioJunior (approved):
LGTM!

PR #884, opened 2026-09-09
fix(engine): forward structured Logs with safe buffer admission

keelancj:
URL4's structured Logs from #877 currently lose their attributes when the Engine converts them for the wire. Optional Logs can also trigger overflow when the bridge's hard capacity is below its soft capacity. This change forwards the attributes and applies the existing drop/eviction policy at whichever limit is reached first.

For example, an instrumented endpoint can emit `"Retrying request"` with `{"attempt": 2}` during node resolution. The fields and node trace attachment now survive Engine publication and existing Client decoding. This PR supplies forwarding only: OME-1161 owns actual model-call activity producers, and OME-1135 owns the Logs tab.

## Implementation

- Forward a detached copy of Log attributes into existing LogData, preserving severity fallback, sequencing and span attachment.
- Drop incoming optional Logs or evict the oldest buffered Log at either limit; count each loss. Authoritative-only hard-cap exhaustion still fails.
- Keep the production change in the existing executor module, with no new queue, event type, domain schema or SDK change.

## Validation

- 27 new cases cover scalar fields, concurrent attribution, live emission before completion, real wire/Client decoding, exact loss counts and paired lifecycle outcomes with/without Log pressure.
- Full `uv run .claude/scripts/run_gates.py screamingface-engine` passed: append-only, Ruff lint/format, Pyright, layering and full pytest coverage gate; 2,682 tests collected and 93% aggregate coverage.
- No existing tests changed; no live provider calls.
- Standards and spec reviews found no actionable issues.

Implements the Engine portion of the approved design in #876, after #877 merged. Keep draft pending owner review.

Ledger: `docs/work/2026-09-09-OME-934-structured-log-forwarding.md`

Asana: not applicable.

Refs: OME-934

### DIFF
--- a/apps/screamingface-engine/src/screamingface_engine/runner/executor.py
+++ b/apps/screamingface-engine/src/screamingface_engine/runner/executor.py
@@ -89,7 +89,7 @@ class _Bridge:
     """A bounded async event queue between the engine's synchronous `url4.observe` callback
     and the async streaming loop draining it.
 
-    Buffers up to `maxsize` events; beyond that, an incoming `Log` is dropped outright, while
+    At either capacity limit, an incoming `Log` is dropped outright, while
     an incoming non-Log event instead evicts the oldest buffered `Log` to make room (or, if no
     `Log` is buffered, evicts nothing and the buffer grows toward the hard cap) — since a `Log`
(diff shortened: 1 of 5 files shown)


IonesioJunior (approved):
LGTM!

PR #885, opened 2026-09-09
docs: propose major-stage evaluation activity design

keelancj:
Researchers need to see whether evaluation work is progressing and where a delay or failure occurred. This proposes structured observations for case loading, answering, model calls, grading checkpoints and aggregation, presented as a compact timeline with expandable safe details.

Reuse the merged URL4 Log sink and Engine forwarding. One small Engine activity module owns timing, validation, cancellation and bounded capture; producers supply facts where the work happens. Generic Runner/URL4 code remains unaware of benchmark stages. The Client interprets the versioned attributes through a bounded projection. Existing results, scores and accounting remain authoritative.

The spec defines a proposed schema, rolling capture/display limits with no lifetime emission cutoff, privacy allowlists, honest handling of lost records and unknown support, and coverage requirements for all seven registered boards. Owner refinements keep activity ephemeral with no new server-side archive, use fixed 60-second heartbeats, prioritize outcomes within bounded admission, retain bounded active/failure summaries, and disclose reconnect gaps. Optional logging limits must not block or fail evaluation; this requires implementation tests. Discarded history is not promised recoverable. The plan separates OME-1161 model producers, a proposed Engine benchmark-stage child, and OME-1135 presentation. Exact semantic attribution, criterion progress and provisional scoring remain separate work.

This is a docs-only proposal. Schema, limits and delivery boundaries need owner review before implementation; merging it does not complete OME-887.

Validation: checked source placements against main at b47853ea; independent Standards and Spec reviews, with findings addressed; relative document links and diff hygiene checked; commit hooks passed. Follow-up review covers rolling limits, fixed heartbeats and reconnect freshness. No product code changed.

Refs: OME-887

Deployment policy: full/off enforcement belongs to OME-1161; aggregate mode requires a follow-up contract, and limited is unsupported. Off disables new optional activity without claiming existing telemetry is private. Review findings add bounded pre-projection decoder ID tracking (with explicit uniqueness-validation tradeoff), safe structured bridge-loss totals on the existing Engine closing diagnostic, and stale rows that become outcome unknown. The owner approved 180 seconds since the last fresh operation update, not since call start. Long calls with fresh 60-second heartbeats stay active. A stale call displays “No recent update; call may still be running”; it is never cancelled or failed by this display threshold. Concrete proposed mechanics now use producer observation timestamps with a documented 30-second relative-clock assumption and conservative age deadlines, a single 100/s burst-200 bucket reserving 40 tokens from routine records, and a 4,096-ID / 1 MiB collision-check window. Under clock skew the freshness warning can appear up to 60 seconds early; unavailable clock assumptions leave freshness unknown. Timestamp, priority and bounded uniqueness semantics are specified with regression cases and remain subject to design approval. No server-side archive is added.


khoaguin:
@keelancj nice doc man.  I think we can have a deployment-owned activity level for different privacy-tier deployments:
- full (default for public runs): everything in the v1 contract — case IDs, per-case timing, live stream.
- aggregate: run-level only — kind + state, counts, quantized durations; no case_id, no per-case records.
- enclave/private runs (off / limited): no activity / very limited records; lifecycle events only.

PR #892, opened 2026-09-10
feat(scoreboard): name the Pareto chart after what it plots

FilipBolt:
## Summary

- rename the benchmark page's chart heading from **"Score for cost"** to **"Pareto Frontier (cost/score)"**, and the scroll region's `aria-label` with it
- add a guard pinning both names, the absence of the old one, and the scope boundary of this change

**Asana:** N/A — tracked in Linear as OME-1146

## This is half of OME-1146, and the ticket stays open

OME-1146 asks for two things. This PR does the rename. It deliberately does **not** delete the self-reported-costs disclaimer, because that half reverses `OME-770` D11 and contradicts two comments already in the code:

`scores/models/score.py`, directly above `verified_by_screamingface = True`:

> INVARIANT: the public portal must not claim more than this… **Change the default and that copy together, or the board lies.**

That default is a placeholder from `OME-820`; `OME-821`, which would give the flag a real signal, is still Backlog.

`scores/schemas.py`, on `LeaderboardEntry.run_cost_usd`:

> Self-reported and unverifiable: re-running a submission tells us what *we* paid, not what the submitter paid… never as a verified figure.

Cost verification is not score verification, so the disclaimer stays true even after `OME-821`. And `OME-1143` — filed the same afternoon, also High — reports that the costs on this exact chart are currently wrong for cache-served runs.

The question is posted on the ticket and awaits the owner. Spec D2 records the reasoning so it does not have to be rediscovered.

## Components touched

- `apps/scoreboard/portal/benchmark.html` — two strings
- `apps/scoreboard/tests/unit/test_portal_static.py`
- `docs/spec`, `docs/plan`, `docs/tasks`, `docs/work`

No JS. The ticket's stray-check step was run first: `grep -rn "Score for cost" apps/scoreboard/` returns exactly three hits, none in `pareto-chart.js` or under `tests/portal/`.

## Test plan

- [x] `uv run .claude/scripts/run_gates.py scoreboard --base origin/main --skip-append-only` → ALL GATES GREEN (ruff check, ruff format, pyright, pytest `--cov-fail-under=80`, all three portal suites). `test_portal_static.py` alone: 16 passed.
- [x] new test written first and confirmed RED on the heading assertion
- [x] mutation-checked, each mutation reverted and the file confirmed byte-identical afterwards:

| Mutation of `benchmark.html` | Result |
| --- | --- |
| heading renamed, `aria-label` left behind | CAUGHT |
| `aria-label` renamed, heading left behind | CAUGHT |
| disclaimer deleted as well — the out-of-scope half | CAUGHT |

The third matters most. It is the only assertion anywhere that pins what this unit actually ships: renamed, and still disclaimed. Without it a future change could strip the provenance line and no test would object.

- [ ] screenshots — not attached; the change is two strings of copy

## Confidence-Gate exception (sdlc rule 5)

`test_pareto_chart_shell_is_bounded_provenanced_and_loaded_before_its_caller` asserted the old `aria-label` at line 189. That assertion names the exact text being renamed, so it cannot survive. Its seven other assertions are unchanged — including the disclaimer at line 192, which this PR preserves on purpose.

`run_gates.py` without `--skip-append-only` flags that one line and nothing else.

## Cross-service notes

None. Copy only; no API, schema, or client change.

### DIFF
--- a/apps/scoreboard/portal/benchmark.html
+++ b/apps/scoreboard/portal/benchmark.html
@@ -64,7 +64,7 @@
     </div>
 
     <section id="pareto-chart-section" class="pareto-chart-section" hidden>
-      <h2>Score for cost</h2>
+      <h2>Pareto Frontier (cost/score)</h2>
       <figure class="pareto-chart-figure">
         <!-- The table below is the accessible data source. This SVG is a visual summary with
              native hover tooltips, so hiding it prevents dozens of duplicate data announcements. -->
@@ -72,7 +72,7 @@
           class="pareto-chart-scroll"
           tabindex="0"
           role="region"
-          aria-label="Score for cost chart, horizontally scrollable"
+          aria-label="Pareto Frontier (cost/score) chart, horizontally scrollable"
         >
           <div id="pareto-chart" aria-hidden="true"></div>
         </div>
--- a/apps/scoreboard/tests/unit/test_portal_static.py
+++ b/apps/scoreboard/tests/unit/test_portal_static.py
@@ -186,7 +186,7 @@ def test_pareto_chart_shell_is_bounded_provenanced_and_loaded_before_its_caller(
 
     assert re.search(r'<section[^>]*id="pareto-chart-section"[^>]*hidden', html)
     assert re.search(r'<div[^>]*class="pareto-chart-scroll"[^>]*tabindex="0"', html)
-    assert 'aria-label="Score for cost chart, horizontally scrollable"' in html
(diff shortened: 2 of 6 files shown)


HupBaHa (approved):
Reviewed at cebdc7f7 against current main b052aca6. No blocking findings. The visible heading and accessible region label stay aligned, the old title is removed, and the self-reported-cost disclaimer remains intact as the explicitly deferred half of OME-1146. Verified the clean merge tree, green PR checks, and apps/scoreboard/tests/unit/test_portal_static.py locally (16 passed).

PR #893, opened 2026-09-10
feat(scoreboard): list only established boards on the portal index

FilipBolt:
## Summary

- the portal index drops private boards, so the Fusion Monsters entry challenge stops rendering as an empty card
- `/v1/benchmarks` is unchanged and still returns every board — `sf.leaderboards` needs the private ones so participants can submit
- rankings are untouched; `OME-894`'s access rule is exactly as it was

**Asana:** N/A — tracked in Linear as OME-1147

## Scope came from the owner's clarification

The original ticket asked for the full rankings to stay public over the API. The owner's comment of 2026-09-09 reverses that — *"none to see the rankings … outside of the product team"* — and asks for a web-only unlisting. The ticket description has been rewritten to match; this PR implements that version.

Two of the three requirements already held:

| Requirement | Mechanism | Status |
| --- | --- | --- |
| rankings stay private | `get_leaderboard` → `_private_leaderboard` | already correct (`OME-894`) |
| `sf.leaderboards` lists everything | `ScoreStore.list_benchmarks()` returns all boards | already correct |
| the index hides it | `initIndex()` rendered everything | **fixed here** |

No API, schema, migration or seed change: `BenchmarkSchema` already exposes `visibility`.

## The filter fails open, deliberately

Only the exact string `"private"` removes a card. Absent, null and unrecognised values keep one. This is a **cosmetic catalogue rule, not an access control** — the server already refuses private rankings — so guessing "hide it" buys no privacy and risks emptying the public catalogue on a serialisation change.

## Why the logic went into `leaderboard-logic.js`

`index.html` already loads it before `main.js`, it is already exercised by `tests/portal/leaderboard-logic.test.js`, and that file is already named in the scoreboard gate list. A helper in `main.js` would be untestable, and a new test file would mean editing the gate list by name (`OME-798`) for one function.

## Test plan

- [x] `uv run .claude/scripts/run_gates.py scoreboard --base origin/main --skip-append-only` → ALL GATES GREEN. `leaderboard-logic.test.js` alone: 43 passed.
- [x] both layers written first and confirmed RED
- [x] mutation-checked, every mutation reverted in a `finally` and all three files confirmed byte-identical afterwards:

| Mutation | Layer | Result |
| --- | --- | --- |
| filter made fail-closed | node | CAUGHT |
| predicate inverted | node | CAUGHT |
| helper defined but never called from `main.js` | python | CAUGHT |
| `index.html` script order reversed | python | CAUGHT |

The last two are why the Python layer exists. The node suite cannot see `main.js`, so a perfect helper that nothing calls — or one whose module loads too late — passes every behavioural test.

- [ ] screenshots — not attached; the visible change is one card no longer appearing

## ⚠️ The append-only gate reports a false positive here

`run_gates.py` without `--skip-append-only` flags `tests/portal/leaderboard-logic.test.js` with *"existing test artifact is unsupported or unparseable"*.

**This is not a rule-5 violation.** The change is `+44/−0`, and a diff filtered for removed lines returns nothing. The checker has no JavaScript parser, so it reports any modification to a `.js` test file rather than reading it. No prior assertion was touched.

Worth a follow-up: as it stands, every future portal JS test addition needs `--skip-append-only`, which trains us to skip the check that also guards the Python suites.

## Note for whoever picks up OME-1112

An `AIDEV-NOTE` at the filter records that `visibility` stands in for **provenance**. The owner's rule is "established versus ours", and the two coincide only while `healthbench-worst30` is both the only board we assembled and the only private one. They part company the first time we run a challenge on an academic benchmark, or publish one of our own.

## Cross-service notes

None. No API, schema, or client change.

### DIFF
--- a/apps/scoreboard/portal/leaderboard-logic.js
+++ b/apps/scoreboard/portal/leaderboard-logic.js
@@ -1,13 +1,15 @@
-/* Pure ranking / SOTA decisions for the leaderboard board (OME-769).
+/* Pure decisions the leaderboard portal makes about what it publishes (OME-769).
  *
  * FEATURE: the per-benchmark submissions board — ranked rows, score bars, and
- * the SOTA medal on the best reproducible result.
+ * the SOTA medal on the best reproducible result — and, since OME-1147, which
+ * boards the index catalogue lists at all.
  *
- * WHY this file exists separately from benchmark.js: these three functions decide
- * what the public board *claims* — which row (if any) is presented as
- * state-of-the-art, and how long each score bar reads. Keeping them free of
- * the DOM makes them assertable in `tests/portal/leaderboard-logic.test.js`
- * without a browser, which the rest of the portal's rendering is not.
+ * WHY this file exists separately from benchmark.js and main.js: these functions
+ * decide what the public portal *claims* — which row (if any) is presented as
+ * state-of-the-art, how long each score bar reads, and which benchmarks a
+ * visitor is shown. Keeping them free of the DOM makes them assertable in
+ * `tests/portal/leaderboard-logic.test.js` without a browser, which the rest of
+ * the portal's rendering is not.
  *
  * Loaded as a plain <script> in the browser (exposing window.SFLeaderboardLogic)
  * and via require() in tests. No build step, matching the rest of the portal.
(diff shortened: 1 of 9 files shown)


HupBaHa (approved):
Reviewed exact head `6931dd3f`. The portal filters only exact `visibility: "private"` values before board fetches, keeps `/v1/benchmarks` complete for participant discovery, and leaves the server-side private-ranking boundary unchanged. No blocking or material code findings. The branch currently conflicts with main in `apps/scoreboard/tests/unit/test_portal_static.py`; after rebase, preserve both the main Pareto-heading test and this PR’s `listedBenchmarks` wiring test, then rerun the Scoreboard gates.

PR #898, opened 2026-09-10
feat(aigateway): carry gateway_call_id on every log line (OME-938)

sergio-bershadsky:
Rung 3 of the correlation ladder. Phase 0 child (`OME-935`) that Phase 1 depends on.

## The problem

The id reached **exactly one** log line — `plugins/taxonomy/session.py` — so an operator holding a `gateway_call_id` from a response body could find the line announcing it and nothing about what the call did. Dispatch, cache, concurrency and error lines were all anonymous.

## The change

A ContextVar bound per request by **pure-ASGI middleware**, plus a log-record factory stamping the bound id onto every record created inside that scope, rendered through a `%(call_context)s` slot exactly as the Engine renders `%(run_context)s`.

Three decisions worth review:

**Pure ASGI, not `BaseHTTPMiddleware`.** The app's only existing middleware is `BaseHTTPMiddleware`, so that was the tempting model. It returns when the response *starts*, and this gateway streams SSE — every line emitted while the body streams (dispatch, retry, provider errors: the interesting ones) would lose its id.

**The id is minted in middleware and CONSUMED by the taxonomy plugin**, not the reverse. The plugin is gated by `AIGW_TAXONOMY_ENABLED`, so minting there made correlation a side effect of usage accounting — turning off accounting deleted the gateway's only correlation mechanism. One id per request now, shared by the response body and every log line; two ids for one call would be worse than none, because both look right.

**Middleware registration order is counter-intuitive** and is commented as such. `add_middleware` does `user_middleware.insert(0, ...)` and the stack wraps that list in reverse, so the **last** registration is the **outermost**. `CallIdMiddleware` is added after the auth guard so the guard's rejection lines are inside a bound scope. I had this backwards first and checked Starlette's source rather than trusting memory.

## A latent bug this exposed, and fixed

First full-suite run: **85 failed, 435 errors**, all `RecursionError: maximum recursion depth exceeded` out of `log_filter.py` — at *fixture setup*, nowhere near anything resembling logging.

Redaction and correlation both own `logging.setLogRecordFactory`, and each guard checked only whether the **outermost** factory was its own. Each installer therefore saw the other's wrapper, concluded it had not run, and wrapped again — every interleaved pair adding two layers. `create_app` installs both and the suite builds thousands of apps, so the chain grew until creating one log record blew the stack.

**The redaction installer had this flaw all along**; it was invisible while it was the only installer. Both guards now walk the wrapper chain via `__wrapped__` (`factory_chain_has`), with a regression test asserting depth stays <= 3 across 50 interleaved install rounds.

Baseline was verified green (4107 passed) before blaming the change. It was correctly mine.

## The security property is tested, not assumed

The injector must **wrap** the redaction factory, never replace it — a replacement silently stops scrubbing provisioning tokens while every observability assertion keeps passing. Both install orders are covered.

Worth flagging: the first version of that test was **vacuous**. It asserted a secret-looking token was absent from a message `RedactProvisioningTokenFilter` would never have matched — its pattern keys on `X-Aigw-Provisioning-Token` followed by `:`/`=`. It would have passed whether or not wrapping worked. Rewritten against a genuinely matching message, and every redaction assertion now also requires `[REDACTED]` to be **present**, proving redaction ran. **Mutation-verified**: replacing the wrap with `logging.LogRecord(...)` makes it fail.

## Test plan

- `run_gates.py aigateway --skip-append-only` — **ALL GREEN**: **4119 passed, 58 skipped** (baseline 4107, +12 new).
- `run_gates.py screamingface --skip-append-only` — **ALL GREEN**.
- Ladder against the real engine + real aigateway: **3 passed, 2 xfailed** — rung 3 green.

Observed on a real run, three loggers, one id:

```
aigateway.plugins.taxonomy.session   gateway_call_id=call_eba9... usage accounting started ...
aigateway.core.concurrency           gateway_call_id=call_eba9... provider concurrency limit applied ...
aigateway.routes.chat_cache_stage    gateway_call_id=call_eba9... global cache fill stored ...
```

## Two prior-test changes — both Confidence-Gate items

**1. Rung 3's assertion was unsatisfiable.** It required *every line in the gateway log* to carry an id. That log also holds startup lines (`loaded provider plugin: openrouter`) belonging to no request, and text that is not a log record at all (a pydantic `UserWarning` on stderr). No correct implementation can pass it — stamping startup lines means inventing an id, the same defect as a well-formed traceparent that joins nothing.

Rescoped to the window between the first and last identified line: full strength where request handling happens, which is where the defect it hunts lives, plus a floor of >= 2 identified lines so it cannot pass vacuously (before this change exactly one line carried an id). Counts: tests 5 -> 5, assertions 14 -> 15, xfail markers 10 -> 8.

**2. `test_a_valid_accounted_request_allocates_one_gateway_call_id` was retargeted, not weakened.** It patched the two former mint sites and asserted one call; the id is now minted once in middleware, so both patches observed zero while behaviour was perfectly correct. The claim is unchanged — one allocation per request, response carries that id. Counts: tests 46 -> 46, assertions 194 -> 194.

## Deliberately out of scope

**Whether logs become JSON.** `OME-1118` wants structured fields so SigNoz can *filter* rather than grep. With a plain-text formatter this delivers a searchable `key=value` — the shape the Engine already emits — not a parsed attribute. Changing the format is a cross-app infrastructure decision (engine, aigateway, scoreboard, plus collector config), not one this unit should settle alone. Worth its own item.

Per the issue: LiteLLM callbacks untouched, request bodies never logged.

Refs: OME-938

### DIFF
--- a/apps/aigateway/src/aigateway/call_context.py
+++ b/apps/aigateway/src/aigateway/call_context.py
@@ -1,0 +1,120 @@
+"""The request-scoped correlation context every aigateway log line is stamped with.
+
+FEATURE (OME-938): before this, `gateway_call_id` reached exactly ONE log line
+(`plugins/taxonomy/session.py`), so an operator holding an id from a response body could find
+the line announcing it and nothing about what the call actually did — dispatch, cache, retry
+and concurrency lines were all anonymous.
+
+The mechanism is a ContextVar bound per request by `middleware.call_id`, plus a log-record
+factory that stamps the bound id onto every record created inside that scope. A factory rather
+than a per-call argument, because the lines that most need attributing are the ones nobody will
+go back and edit: LiteLLM's, a library's, an exception handler's.
+
+INVARIANT: `install_call_context_injection` WRAPS whatever factory it finds; it never replaces
+one. `core.auth.log_filter.install_provisioning_token_redaction` owns the same single
+`logging.setLogRecordFactory` slot, and replacing it would silently stop scrubbing provisioning
+tokens — a security regression hiding inside an observability change, invisible to every
+observability assertion. Both install orders are covered by tests.
+"""
+
+from __future__ import annotations
+
+import contextvars
+import logging
+import secrets
+from collections.abc import Iterator
+from contextlib import contextmanager
+
+from aigateway.core.auth.log_filter import factory_chain_has
+
+_INSTALLED = "_aigw_call_context_injection"
+"""Marks the factory THIS module installed, so a second install is a no-op."""
+
+_RECORD_ATTR = "gateway_call_id"
+"""The attribute name the injector stamps on each record, and the field name in the output.
+
+One constant, because the renderer (`logs.CallContextFilter`) and every reader must agree with
+the writer. A drift here is silent: records carry the id, nothing renders it.
+"""
+
+
+def record_call_id(record: logging.LogRecord) -> str | None:
+    """The call id stamped on `record`, or None if it was created outside a request.
+
+    A typed accessor rather than `getattr(record, "gateway_call_id", None)` at each site:
+    `LogRecord` has no such attribute in its type, so every direct read is a pyright error, and
+    the string would be repeated in the renderer, the tests and any future consumer.
+    """
+
+    value = getattr(record, _RECORD_ATTR, None)
+    return value if isinstance(value, str) else None
+
+
+_call_id: contextvars.ContextVar[str | None] = contextvars.ContextVar(
+    "aigateway_gateway_call_id", default=None
+)
+
+
+def new_gateway_call_id() -> str:
+    """A fresh id in the shape the public contract already fixes.
+
+    `call_<32 hex>` is pinned by `plugins/taxonomy/usage_accounting.schema.json` and appears in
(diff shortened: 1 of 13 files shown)


keelancj (commented):
@sergio-bershadsky  Middleware-owned ids are a nice improvement + factory-chain fix and regression coverage is solid. Two gaps came up in review around error logging

1. `middleware/call_id.py:369-375`:
Unhandled exceptions won't show the correlation id. It gets cleared as the exception unwinds out of the with call_scope(call_id): block before Starlette/Uvicorn's own error logging runs. So the one log line reporting a real crash has no id.

2. `logs.py:88-91`:
Uvicorn and library loggers don't print the id. Only the aigateway logger's handler got a format string that renders gateway_call_id. Uvicorn's own loggers and any third-party ones keep their original formats, so they never show it.

3. `middleware/call_id.py:41` (no behavioural harm)
The comment says downstream consumers read the id from scope["state"], but no current consumer does. The actual readers all use current_call_id() or AccountingSession. Safe to delete, or the comment needs updating

PR #903, opened 2026-09-10
fix(aigateway): strip new LiteLLM 1.100.0 dynamic callback params

keelancj:
## Summary

The bundled gateway must remove caller-supplied LiteLLM callback controls before dispatch. LiteLLM 1.100 adds `langfuse_environment`, `newrelic_api_key`, and `newrelic_region`; the old gateway filter did not cover them.

Review also found an alternate input path: `litellm_trusted_callback_vars`. LiteLLM treats this internal container as proxy-stamped configuration. The gateway previously forwarded it unchanged, allowing a caller to supply New Relic settings through that trusted path even when ordinary callback fields were blocked.

This is a callback-configuration trust-boundary issue. Telemetry effects depend on an enabled integration; it is not evidence of unconditional data exfiltration or normal-chat failure.

## Scope and implemented fix

- Strip the three new callback fields from the request body and metadata.
- Strip the entire `litellm_trusted_callback_vars` and `user_api_key_auth_metadata` containers from request bodies and metadata, regardless of shape. The latter is consumed as trusted project/service routing configuration by LiteLLM 1.100 OTel integrations.
- Cover New Relic, Datadog, empty and malformed containers with regression tests, including the real LiteLLM callback initializer. Preserve ordinary metadata and do not mutate input.
- Extend the independent exact-set inventory without removing existing assertions.
- Upgrade the gateway lock to LiteLLM 1.100.1 and re-verify all 13 guarded globals before updating the two version assertions.

## Package relationship

[PR #858](https://github.com/ScreamingFace/screamingface/pull/858) upgrades the optional `screamingface[runtime]` dependency to LiteLLM 1.100.0. That extra bundles gateway source; ordinary client/notebook installations do not install LiteLLM. This fix protects the bundled source as well as the separately deployed gateway.

The two dependency pins remain 1.100.0 and 1.100.1; exact alignment remains separate work under OME-1173.

## Validation

The initial review passed 4,181 gateway unit tests on 1.100.1 and 31 hardening tests on the package's 1.100.0. The trusted-container follow-up reproduced six failing cases before the fix and passed all 37 focused hardening tests after it. All follow-up gates passed: lint, formatting, pyright, no-enterprise, and the full test suite with 93% coverage. All 37 focused hardening tests also passed against the package runtime's LiteLLM 1.100.0. Follow-up commit: `494f2a3a`.

PR: https://github.com/ScreamingFace/screamingface/pull/903
Follow-up ledger: `docs/work/2026-09-10-OME-1177-trusted-callback-envelope.md`.

## Telemetry auth-metadata follow-up

Commit `d9bd461f` blocks caller-supplied `user_api_key_auth_metadata` through the complete ingress/projection/preparation pipeline. Twelve new regression cases failed before the fix and passed afterward. All 50 focused hardening tests pass on both LiteLLM 1.100.1 and 1.100.0. Full gateway gates pass with 93% coverage.

Review correction: the complete gateway already rejects caller `rust` during parameter projection. The earlier direct-handler reproduction did not establish a caller-accessible bypass. This follow-up changes telemetry filtering only.

Ledger: `docs/work/2026-09-10-OME-1177-otel-metadata-ingress.md`.

## Test-preservation exception

Each follow-up appends one field to the independent exact-set inventory. No prior assertion is removed or weakened. The append-only gate flags that intentional contract extension; the remaining full gates run with `--skip-append-only`, documented in the ledger.

Refs: OME-1177

## Main refresh — 2026-09-16

Rebased onto main `0a260b4a`. All four existing patches are unchanged (`git range-diff`); the additional documentation commit records approval and validation. `uv lock --check` and regeneration pass without changes, retaining LiteLLM 1.100.1 and main's OpenTelemetry 1.44.0 dependencies. Full gateway gates pass with 92.8% coverage; 98 focused hardening/runtime-guard cases pass.

Live-provider validation remains outstanding: this worktree has no app-local `.env` and no live admin credentials, OpenAI/OpenRouter keys, or configured gateway connection. No live pass is claimed. The PR remains open for review/merge.

Ledger: `docs/work/2026-09-16-OME-1177-main-refresh.md`.


khoaguin:
**Review: approve with nits.** Strip completeness was verified against a live `litellm==1.100.1` install: `iter_client_callback_metadata_dicts` reads exactly `litellm_params.metadata` / `litellm_metadata` / `metadata` (all stripped or sanitized), `get_trusted_callback_params` reads only top-level `litellm_trusted_callback_vars`, `auth_metadata` reads only `user_api_key_auth_metadata` — no bypass, no aliasing/case hole, and malformed/None/list containers are dropped wholesale so no legitimate-request 500 path exists. Tests feed sanitized output to LiteLLM's real consumer functions, not shims. 

The remaining risk is process, not code — this is the second denylist-chasing-a-bump PR (Datadog fields in #478 were the first), and both trusted-container gaps here were found by human review. Two tripwires would make the next bump fail loudly instead of failing open:

1. **Pin the strip list to LiteLLM's own exports** — `core/request_hardening.py:29-33`: the dynamic-param class already has a subset tripwire (it caught the new trio); the trusted-container class has none. LiteLLM 1.100.1 exports constants to pin against: assert `initialize_dynamic_callback_params.TRUSTED_CALLBACK_VARS_FIELD in DISPATCH_CONTROL_FIELDS` and that the client-metadata slot set is covered — an upstream rename or new slot then fails the suite at the version bump. (`user_api_key_auth_metadata` has no upstream constant; the guarded version pin stays the alarm for it.)
2. **Run the hardening subset against the runtime package's own LiteLLM pin** — the coverage tripwire only runs against the gateway's lock (1.100.1), while `screamingface[runtime]` bundles this source under an independently resolved litellm (1.100.0) — exactly the PR #858 shape that triggered this ticket. The PR ran the 37 hardening tests against 1.100.0 manually; codifying that in the runtime package's CI fits the open OME-1173 alignment work.

Longer-term (own ticket, not this PR): the filter is a denylist over an unbounded upstream control plane; an allowlist of permitted top-level chat fields would fail closed for container-style additions the subset tripwire can't see.

Mechanical: the INVARIANT comment at `request_hardening.py:55-56` now understates the set's contract — it also carries two container names deliberately *outside* `_supported_callback_params`; one clause keeps the next editor from "cleaning them up" as stale. Ledger typos in `docs/work/2026-09-10-OME-1177-trusted-callback-envelope.md:24-26` (`GREEN:37`, `gateway1.100.1`, `and80%`).

khoaguin:
Follow-up after verifying against litellm 1.100.0 **and** 1.100.1 source (the callback-params module is byte-identical in both, so the 1.100.1 pin is fine). **The code is right; two comment claims need correcting.**

Confirmed accurate:
- `langfuse_environment` is new in `_supported_callback_params` and NOT in `_request_blocked_callback_params` → genuinely caller-injectable via body or metadata. Strip justified.
- Coverage invariant holds: all 29 names in 1.100's `_supported_callback_params` appear in the post-PR field set, so `test_litellm_dynamic_callback_parameter_set_is_covered` passes.
- `litellm_trusted_callback_vars` = `TRUSTED_CALLBACK_VARS_FIELD` (`types/utils.py:3516`); `get_trusted_callback_params` reads it straight off whatever mapping it's handed with no provenance check — stripping it at both ingress locations is sound.
- `user_api_key_auth_metadata`: the OTel metadata model reads it for project/service selection (`integrations/otel/model/metadata.py:263`), and the proxy *merges into* a caller-supplied dict rather than replacing it (`litellm_pre_call_utils.py:1421-1423`) — caller-planted keys would survive. Strip fully justified.

Two comment corrections:

1. **"New Relic trio" → pair.** Only `newrelic_api_key` and `newrelic_region` exist; the trio is the three names *added* (those two + `langfuse_environment`). The current wording attributes three params to New Relic.
2. **"`newrelic_api_key` is caller-injectable" is overstated.** Both `newrelic_*` params (and all `dd_*`) sit in litellm 1.100's `_request_blocked_callback_params`: `initialize_standard_callback_dynamic_params` skips them for both top-level kwargs and metadata, so request-supplied values only ever flow through the proxy-stamped trusted overlay. The gateway strip is correct defense-in-depth (and the coverage invariant demands the names), but litellm itself already blocks caller injection on the ordinary path — the real caller hole for these was the trusted container. Suggested rewording: "blocked upstream in 1.100's `_request_blocked_callback_params`; kept here for the coverage invariant + defense-in-depth."

HupBaHa (approved):
Reviewed exact head `d9bd461f`. The ingress filter closes the LiteLLM 1.100 callback and trusted-metadata paths at both the request root and metadata boundary, preserves ordinary metadata, and is covered against the real 1.100.0/1.100.1 consumers. No blocking or material findings. After rebasing onto current main, regenerate `apps/aigateway/uv.lock` and rerun CI so the LiteLLM update and main OTel dependencies are both retained.

PR #915, opened 2026-09-11
feat: connect optional observers to Engine execution

keelancj:
This PR connects the observation interfaces from [PR 899](https://github.com/ScreamingFace/screamingface/pull/899) to real Engine execution. When a model request needs a retry, an injected observer now receives the actual attempt number and selected delay. A future activity plugin can translate that into “Model request retrying in 2 seconds” for the researcher’s logs.

The default observer registration is empty, so this PR alone does not add researcher activity records. The next delivery supplies the complete activity plugin; the Client Logs tab remains a separate follow-up.

### Why this belongs in the Engine

The connector already decides when requests start, retry and finish. Reporting those facts here avoids asking benchmark code to infer them or duplicate retry logic. Core imports only its observation interfaces: the plugin remains responsible for its schema, privacy policy, occurrence IDs, admission limits and heartbeat resources.

### What changes and why

- Composition accepts observer factories with an empty default. The execution wrapper creates fresh observers per run, so concurrent evaluations cannot share telemetry state accidentally.
- Context binding surrounds each inner iterator advancement and cleanup, restoring context before outward yields. This permits another task to close the iterator safely. Observer cleanup also runs if iterator creation fails synchronously.
- Model calls capture the current URL4 log sink and report start, actual retries, completion or failure through the generic interfaces. Request decisions, results and accounting stay with existing execution code.
- The existing transport-loss warning can receive validated structured attributes from an observer. Its original message and empty-registration behavior remain intact.
- Existing operator heartbeat behavior remains independent. Its task is cancelled and joined when a call ends, including cancellation.

### Observer latency contract (OME-1201)

Observer hooks execute inline on the caller’s task. Both protocols require prompt callbacks and put slow network/disk I/O on observer-owned resources. Async cleanup cancels/joins owned tasks promptly without waiting for remote delivery. A characterization test demonstrates same-task, before-return dispatch; it does not enforce a latency limit. This is documentation and a test only, with no dispatch behavior change.

### Review boundary

**Diff: 491 changed lines (474 additions, 17 deletions), including tests and shared docs.**

PR 899 merged at `dcaba228` on 14 September. This draft is rebased onto `main` and targets `main`; the temporary stack is removed. The integration commit rebased without conflicts or code changes, preserving intervening changes including the terminal diagnostics from PR 901.

No activity implementation, deployment configuration, Helm, URL4 protocol or Client changes are included. The intended next boundary is the complete activity plugin, with its size reviewed separately. The integration retains the 500 changed-line maximum, including tests and updates to the same four shared documents.

### Validation

- Tests first: five integration regressions failed before wiring; an additional regression reproduced skipped observer cleanup on synchronous iterator-creation failure.
- 31 focused cases pass, including the unchanged 899 regressions, actual retry/outcome delivery, concurrent run ownership, cross-task close, and structured bridge-loss decoration.
- A copied installation without activity exercises real connector requests/retries, results/accounting, provider failure, cancellation, operator diagnostics and top-level composition using fake HTTP providers.
- Full local Engine gates passed: append-only against updated main, Ruff lint/format, Pyright, layering, and the full pytest/coverage gate. Both independent reviews have no remaining findings after the cleanup fix.

Refs: OME-1161, OME-1201


IonesioJunior (approved):
LGTM!

PR #919, opened 2026-09-11
feat(aigateway): ship the OTLP endpoint and credential through the chart

sergio-bershadsky:
`OME-1132` made aigateway emit spans and left it **inert** — nothing set `OTEL_EXPORTER_OTLP_ENDPOINT` and the chart had no way to. This turns it on, mirroring `OME-1131`'s engine shape deliberately: two charts configuring one feature two different ways is a tax paid at every future edit.

Off by default, so every existing installation is unaffected.

## This PR also fixes its own gate — and that matters more than the chart

The issue asked where the assertion should live: a pytest render test like the engine's, or `verify_chart_wiring.py`. Investigating that turned up something that **changes the answer and reveals a defect in already-merged work**.

`aigateway-tests.yml` has **no helm**. A pytest chart test would therefore carry `skipif(shutil.which("helm") is None)` and **skip silently in CI** — passing locally, asserting nothing on the merge gate.

That is not hypothetical. It is exactly what happened to:

- **`OME-1131`'s 12 engine chart tests** (merged in #908), and
- the pre-existing **`test_chart_render_runner_pool.py`** — including its assertion that deploy-time credentials never appear as literals in a rendered manifest, **which has therefore never once run in CI.**

The engine's `test` job runs pytest with no helm; its `chart` job installs helm but runs no pytest. The two never meet. Filed as **`OME-1189`** (High).

So this PR adds `azure/setup-helm@v5` to the `test` job. **A test that cannot run is not a deliverable**, so it belongs in this unit rather than a follow-up.

## Also: I recorded a wrong fact in `OME-1131`, now corrected

That ledger/mirror/PR claims the engine chart is *"gated by the engine's own workflow, **not by `charts.yml` at all**"*. False — `charts.yml` lists `apps/screamingface-engine/deploy/helm/**` explicitly in both path filters. I read the first few entries of the list rather than all of them. Corrected on the issue.

## The chart

| value | lands as | where |
|---|---|---|
| `tracing.endpoint` | `OTEL_EXPORTER_OTLP_ENDPOINT` | ConfigMap |
| `tracing.serviceName` | `OTEL_SERVICE_NAME` | ConfigMap (omitted when empty) |
| `tracing.resourceAttributes` | `OTEL_RESOURCE_ATTRIBUTES` | ConfigMap (omitted when empty) |
| `tracing.headers` | `OTEL_EXPORTER_OTLP_HEADERS` | **Secret only** |

Three deliberate divergences from the engine chart, each because this chart differs:

- **No schema rule.** This chart has no `values.schema.json`, so the template `fail` on `enabled && !endpoint` is the ONLY guard rather than a backstop — hence a dedicated test for it.
- **The credential is OPTIONAL**, unlike every other one here. The collector is in-cluster and needs none, so the pod references the Secret only when one exists; an `envFrom` naming a Secret that is never created stops the pod from starting.
- **No `lookup` reuse in the Secret.** `OME-1131` wrote that pattern and removed it as a bug: the template renders only when headers are supplied, so the reuse branch fires exactly where the operator *did* supply a value, silently reverting a rotation to the stale cluster copy. Documented in the template so it is not re-introduced.

## Test plan

- `run_gates.py aigateway` **ALL GREEN**; full suite **4250 passed, 58 skipped**. 11 new tests.
- `verify_chart_wiring.py` **112/112**; `helm lint` clean.
- **Five mutations, five killed** — endpoint dropped from the ConfigMap; headers leaked *into* the ConfigMap; Secret referenced unconditionally; the `fail` guard disabled; unset `serviceName` rendered as `""`.
- **Rendered and read by eye** in the real production shape: the three OTLP vars in the ConfigMap, headers absent. With a credential set, `grep -c` finds it exactly **once** in the whole manifest — in the Secret.

## What turns it on

Deployment is **ArgoCD**; `screamingface-sf-aigw` (dev) tracks this repo's chart with values from `$values/kubernetes/apps/sf-aigw/values/dev.yaml` in `OpenMined/infrastructure`. Four lines there:

```yaml
tracing:
  enabled: true
  endpoint: http://signoz-otel-collector.signoz.svc.cluster.local:4318
  serviceName: aigateway
  resourceAttributes: deployment.environment=dev
```

⚠️ **Never point it at the SigNoz UI hostname.** `https://signoz.pulse.dev.openmined.org/v1/traces` answers `200 text/html` (SPA catch-all); OTel treats any 2xx as success, so that hostname discards every span while reporting healthy. Warned beside the field in `values.yaml`.

Preview environments carry inline values in their ArgoCD Application rather than a values file, so they stay off — whether previews should trace at all is an owner call.

Ledger: `docs/work/2026-09-11-OME-1184-aigw-otlp-chart.md`

Refs: OME-1184

### DIFF
--- a/.github/workflows/aigateway-tests.yml
+++ b/.github/workflows/aigateway-tests.yml
@@ -31,19 +31,6 @@ jobs:
 
       - uses: astral-sh/setup-uv@v7
 
-      # WHY helm is in the TEST job (OME-1184): `tests/unit/test_chart_render_tracing.py`
-      # renders the chart and asserts against the rendered manifest, because `helm lint`
-      # reports "0 chart(s) failed" for a chart that cannot render at all.
-      #
-      # Those tests carry `skipif(shutil.which("helm") is None)`. Without this step they would
-      # SKIP here and pass while asserting nothing — not hypothetical: it is exactly what
-      # happened to the engine's equivalent chart tests, whose `test` job has no helm and whose
-      # `chart` job runs no pytest, so they have never once executed on the merge gate
-      # (OME-1189). A conditional skip reads as healthy locally and is invisible in CI output.
-      #
-      # SECURITY: a pinned third-party action, no `github.event.*` interpolation anywhere.
-      - uses: azure/setup-helm@v5
-
       - uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.python-version }}
--- a/.github/workflows/aigateway-tests.yml
+++ b/.github/workflows/aigateway-tests.yml
@@ -31,6 +31,19 @@ jobs:
 
       - uses: astral-sh/setup-uv@v7
 
+      # WHY helm is in the TEST job (OME-1184): `tests/unit/test_chart_render_tracing.py`
+      # renders the chart and asserts against the rendered manifest, because `helm lint`
+      # reports "0 chart(s) failed" for a chart that cannot render at all.
+      #
+      # Those tests carry `skipif(shutil.which("helm") is None)`. Without this step they would
+      # SKIP here and pass while asserting nothing — not hypothetical: it is exactly what
+      # happened to the engine's equivalent chart tests, whose `test` job has no helm and whose
+      # `chart` job runs no pytest, so they have never once executed on the merge gate
+      # (OME-1189). A conditional skip reads as healthy locally and is invisible in CI output.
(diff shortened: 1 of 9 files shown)


bitsofsteve (commented):
LGTM. Might be worth updating the PR body since it's dated.

PR #921, opened 2026-09-11
feat(engine): default tracing on so ArgoCD ships it

sergio-bershadsky:
**This is the PR that unblocks Phase 2 without infra-repo access.**

`OME-1131` shipped the engine's `tracing` stanza off by default. Turning it on needs four lines in `kubernetes/apps/sf-url4-cloud/values/dev.yaml` inside `git@github.com:OpenMined/infrastructure` — which this account cannot reach (full `repo` scope, but `OpenMined` is not among its orgs).

ArgoCD merges chart defaults **under** that values file. So defaulting here ships tracing with no infra change, and an operator still overrides it normally.

```yaml
tracing:
  enabled: true
  endpoint: http://signoz-otel-collector.signoz.svc.cluster.local:4318
```

Collector confirmed from its **own startup log** (`otelcol.component.id=otlp kind=receiver endpoint=[::]:4318`), namespace `signoz`, cluster `aks-dev`. No credential — no Secret renders.

## ⚠️ This has a sunset condition, not an indefinite life

Defaulting a **cluster-specific** endpoint is defensible only because neither chart in this repo is published: the engine's OCI publishing is deferred until the NATS dependency is pinned and vendored (`release-screamingface-engine.yml`, **`OME-567`**), and aigateway's `chart:` job only lints and renders. These charts are consumed solely by ArgoCD from this repo.

**When `OME-567` lands, both `enabled: true` and the hardcoded endpoint must go back to opt-in.** Stated at the field in `values.yaml`, in the ticket, and in the ledger.

## Previews are tagged, not excluded

Defaulting on means `sf-preview-pr-*` namespaces export too, and their spans would otherwise be **indistinguishable from dev's** — a direct OTLP export carries the app's own Resource, and the k8s collector's namespace enrichment applies to the LOGS it scrapes, not to spans an app posts itself.

So `OTEL_RESOURCE_ATTRIBUTES` defaults to `deployment.environment=<release namespace>`. Tagging beats excluding: a preview exporting under its own environment name is *useful*, and a namespace-prefix check would put cleverness in the chart that nothing else there does.

**The default lives in the TEMPLATE, not `values.yaml`** — Helm does not template values files, so `{{ .Release.Namespace }}` written there renders as that literal string and every service would report the template source as its environment. A test pins exactly that (`assert "{{" not in preview`).

## ⚠️ Modifies prior tests — rule 5 Confidence-Gate, owner-approved

The append-only gate fired, which is the point of it. The decision was named **before** the work (in `OME-1190`'s description and the ledger), the owner chose the behaviour that forces it, and the override is the sanctioned `--skip-append-only` — not a weakened assertion.

**Net assertions went UP: 12 → 17.**

| prior test | what happened |
|---|---|
| `test_tracing_is_off_by_default_…` | inverted to `…is_ON_by_default_…`; the old property preserved as a NEW `test_the_off_switch_still_works` under explicit `enabled=false` |
| `test_unset_descriptive_values_are_ABSENT_rather_than_empty` | **split** — `serviceName` half kept verbatim, `resourceAttributes` half replaced by three stronger tests |
| the two refusal tests | now pass `tracing.endpoint=` explicitly, since `enabled=true` alone no longer produces the empty-endpoint state they test |

`serviceName` deliberately keeps its absent-when-unset rule: the code already has the right default, and rendering `""` would report a blank `service.name` — indistinguishable from unconfigured. `resourceAttributes` differs because the chart knows something the code cannot: its namespace.

## Test plan

- `run_gates.py screamingface-engine --skip-append-only` **ALL GREEN**. 17 chart tests.
- **Four mutations, four killed** — namespace attribute hardcoded instead of derived; the default made unconditional so an override is ignored; default endpoint swapped to the SigNoz **UI** hostname; the `enabled` guard forced true so the off switch breaks.
- Rendered by eye: bare `helm template` emits the in-cluster endpoint; `-n sf-fusion` and `-n sf-preview-pr-999` produce different `deployment.environment` values.

## Note

aigateway's equivalent is **not** here — it needs #919 (`OME-1184`) merged first, since that is what adds the stanza this would default. No stacked PRs in a squash repo.

Ledger: `docs/work/2026-09-11-OME-1190-engine-tracing-default.md`

Refs: OME-1190

### DIFF
--- a/apps/screamingface-engine/deploy/helm/templates/configmap-runner-env.yaml
+++ b/apps/screamingface-engine/deploy/helm/templates/configmap-runner-env.yaml
@@ -77,9 +77,19 @@ data:
   # the documented default apply.
   OTEL_SERVICE_NAME: {{ . | quote }}
   {{- end }}
-  {{- with .Values.tracing.resourceAttributes }}
-  OTEL_RESOURCE_ATTRIBUTES: {{ . | quote }}
-  {{- end }}
+  # FEATURE (OME-1190): defaults to the release NAMESPACE, so a preview environment's spans are
+  # distinguishable from dev's. They would otherwise be identical — a direct OTLP export carries
+  # the app's OWN Resource, and the k8s collector's namespace enrichment applies to the LOGS it
+  # scrapes, not to spans an app posts itself.
+  #
+  # WHY here and not in `values.yaml`: Helm does not template values files, so
+  # `{{ `{{ .Release.Namespace }}` }}` written there renders as that literal string and every
+  # service reports the template source as its environment.
+  #
+  # Unlike OTEL_SERVICE_NAME above, this is NOT gated on a non-empty value: the chart knows
+  # something the code cannot — which namespace it is being installed into — so it has a
+  # meaningful default to supply rather than an empty one to avoid.
+  OTEL_RESOURCE_ATTRIBUTES: {{ .Values.tracing.resourceAttributes | default (printf "deployment.environment=%s" .Release.Namespace) | quote }}
   {{- end }}
   # NOTE: OTEL_EXPORTER_OTLP_HEADERS is deliberately ABSENT here, for the same reason as the S3
   # secret key below — it carries the collector credential.
--- a/apps/screamingface-engine/deploy/helm/values.yaml
+++ b/apps/screamingface-engine/deploy/helm/values.yaml
@@ -243,7 +243,18 @@ tavily:
 # mounted in the run mode), so configuring it there would wire up an exporter that does not
 # exist. aigateway is instrumented separately.
 tracing:
-  enabled: false
+  # ON by default (OME-1190) — deliberately, and conditionally.
+  #
+  # WHY the chart rather than the deployment's values file: ArgoCD merges chart defaults UNDER
+  # `$values/kubernetes/apps/sf-url4-cloud/values/dev.yaml` in `OpenMined/infrastructure`, so
+  # defaulting here turns tracing on with no change to that repo. An operator overrides it the
+  # normal way.
+  #
+  # ⚠️ THIS IS ONLY DEFENSIBLE WHILE THE CHART IS UNPUBLISHED, and that is a SUNSET condition.
+  # `release-screamingface-engine.yml` defers OCI Helm publishing until the NATS dependency is
+  # pinned and vendored (OME-567). The endpoint below is specific to `aks-dev`, so the day this
+  # chart ships to anyone else, BOTH lines must go back to being opt-in. Revisit with OME-567.
+  enabled: true
   # Base OTLP/HTTP endpoint, e.g. http://signoz-otel-collector.platform.svc.cluster.local:4318
   # The exporter appends `/v1/traces`. Required when enabled.
   #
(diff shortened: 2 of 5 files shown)


bitsofsteve:
Thanks for this — the chart mechanics are sound, and the in-cluster endpoint and the UI-hostname warning are exactly right. We'd like to turn tracing on from our side instead of defaulting it on in the chart, and we're asking to close this PR (or trim it, see below).

## What's happening

We reviewed #921 against the live cluster. Defaulting the endpoint on in the chart ships it to places where the collector is not reachable today, and the engine's exporter then pays for that on every run:

- **Dev.** The network boundary around your dev namespace allows only DNS, your own services, and the public internet on 443. The collector is inside the cluster, so every export packet is dropped. The engine's sink flushes and shuts down at the end of each run, so each run waits out the exporter timeout on a connection that never answers, and every span is discarded. Empty trace view, slower exit per run, nothing in the logs that says why.
- **Staging.** There is no in-cluster collector there at all. The hostname does not resolve, so every run logs an export failure.
- **Previews.** Same public-only boundary as dev. Their spans would be dropped too, so the namespace tag does not buy anything yet.

Turning it on is our job: the values file for your dev deployment is ours, and we open the network path alongside it. That is what we're doing for aigateway already, and the engine gets the same treatment.

## What changes for you

- Keep `tracing.enabled: false` and `endpoint: ""` as the chart defaults, as OME-1131 shipped them. No sunset condition needed.
- Nothing else in your flow changes. Merge to main, and dev picks up the new chart and image as it does today.
- Previews stay off for now. Staging stays off until it has a collector.
- If you keep the namespace-derived attribute default, please name it `deployment.environment.name`. The edge already stamps that name on the spans it emits for your routes, so the two line up in SigNoz.

## What we need from you

1. Close #921, or trim it to the attribute default and its tests with `enabled` and `endpoint` back to their previous values.
2. Confirm our reading that only the runner pods export spans (the runner env carries the OTLP variables, the app pod does not). We scope the network opening to those pods.
3. Same for aigateway after #919 merges: please leave its default off. We set the values.

## What's next

We open the network path for the engine's runner pods now, and set the values once the next pin lands. You'll see `otlp span export enabled` in the runner logs and `screamingface-engine` appear as a service in SigNoz. We'll post the first trace ID here.

One thing on our side this surfaced: the collector does not yet stamp the source namespace onto spans the way it does for logs. We'll fix that, so your namespace tag becomes a convenience rather than the only identity on the span.

sergio-bershadsky:
Thanks — this is the review I needed, and the dev/staging/preview breakdown corrects a real error rather than a stylistic one.

I reasoned "in-cluster Service DNS ⇒ reachable" and never considered the NetworkPolicy. Worse, the failure mode you describe is one my own design makes silent: the sink swallows exporter errors by contract (a collector outage must never fail a run), and it flushes on shutdown. So each run would have paid the exporter timeout on a connection that never answers, with nothing in the logs saying why — degraded runs plus an empty trace view, which is precisely the "tracing works and the backend is empty" state this phase exists to avoid. Turning it on is opening the network path *and* setting the values; I could only see half of that, so the half I could reach looked like the whole job.

**Closing #921** (not trimming). You'll make the collector stamp the source namespace itself, which retires the chart-side attribute — no point shipping something you're about to make redundant. Both charts keep `tracing.enabled: false` and `endpoint: ""` exactly as `OME-1131`/`OME-1184` shipped them, and I'm not adding the aigateway equivalent. If you do want the attribute later it's a small PR and I'll name it `deployment.environment.name`.

## Your question 2 — confirmed, with one precision that matters for the NetworkPolicy

Your reading is right: only the runner pods carry the OTLP variables, and the App pod does not. Verified against `main`:

| | OTLP env | constructs an exporter |
|---|---|---|
| App / control-plane pod | ✗ — `deployment.yaml`'s `envFrom` takes only its own ConfigMap, which has **0** `OTEL_` keys | ✗ |
| **Runner pool pod** (`url4-runner`) | ✓ via `…-runner-env` | ✓ |
| `serve --local` | — | ✗ |

`OTEL_` appears in exactly one template, `configmap-runner-env.yaml`. The only sink construction in the tree is `runner/main.py:379` (`sink_from_env`) and `:545` (`SpanRelay`).

**The precision:** that code is the `screamingface-engine run` entrypoint, which the worker **forks as a child process inside the runner pool pod** (`worker/exec_wrapper.py` execs it). The worker process itself never builds an exporter — the layering gate forbids it importing the run half at all. So if you scope egress to "the thing that exports" you'd be scoping to a subprocess; the right target is the **runner pool Deployment's pods**, because the children share that pod's network namespace.

Also worth knowing, since it narrows the scope further: **`serve --local` exports nothing at all.** The in-process adapter calls `lifecycle.run` directly and wraps no relay, so a local-mode process needs no egress even though it can execute runs. Only the deployed worker-pool path exports.

## Your question 3

Confirmed — aigateway's default stays off. #919 merged with `enabled: false` and no endpoint.

## One thing back to you

When you set the values, the engine's runner logs `otlp span export enabled service=…` at boot — that's the cheap confirmation the config landed, before any span is emitted. If you see runs but never that line, the values didn't reach the runner env.

Happy to run the deployed acceptance (`e2e/failor/verify_phase.py`) against the first trace ID when you post it.

PR #922, opened 2026-09-11
feat(scoreboard): accept and classify declared model identities

FilipBolt:
Scoreboard half of `OME-1179`. Gives the board the inputs to classify openness by **model** rather than by the reseller that carried the request. Publishes no new number — `OME-1145` owns the metric and the response.

## The problem

The Client truncates every declared route to its first path segment, so `openrouter/deepseek/deepseek-v4-pro` arrives as `"openrouter"` — which is on the closed marker list. All seven live `draco-3pass` entries store `ran_with_providers = ["openrouter"]`, and an entry named `best_open_source` (deepseek, kimi, qwen) is published as **closed**.

## What lands

| | |
|---|---|
| **Accept** | optional `models` on `ScoreSubmission`, capped at 32 routes / 255 chars / 4096 bytes |
| **Store** | nullable `models` JSON column + migration `0013_score_models` |
| **Derive** | stored `ran_with_providers` computed from the routes, so the two fields cannot contradict each other |
| **Classify** | `classify_model(route) -> open \| closed \| unknown` — **structural**: the owner segment is matched exactly, family exceptions are scoped to their owner |
| **Replay** | a same-owner republish may fill in routes on a row that lacks them |
| **Read** | `models_for_score_ids()`, chunked, projecting two columns |

## Decisions this encodes

* **OME-1179 D1** — any-closed-wins stays; the unit of the statistic is the entry, not the model. So there is no new aggregation, only better input.
* **Q1** — "open" means weights you can download and run, whatever the licence says about commercial use. `mistralai` is therefore an open **owner**, so Mistral Large 2411 classifies correctly and the earlier "tighten mistral" instruction is withdrawn. (An earlier revision of this description said `mistral` stays a bare substring; matching is now structural, not substring — see below.)
* **Q2** — `models` lands on `ScoreSchema` and stays off `LeaderboardEntry`, the ranked-board payload. It is **not** internal, though: `ScoreSchema` is itself a response model, so the field carries `exclude_if` and an absent value never reaches the wire. (This bullet originally called the schema internal. It is not — see round 2 below.)
* **Q3** — a same-owner replay may fill in routes, reusing the existing anti-hijack guard and visibility lock. `models` stays out of `_content_hash`.

## Two invariants worth review attention

**`_content_hash` reads the wire `ran_with_providers`, never the derived value.** Storage may be corrected; identity may not be rewritten underneath existing rows. Hashing the derived value would recompute identity for every row whose client-sent providers differ from its routes, and each would stop deduplicating to its stored twin.

**`unknown` is a third verdict, not a synonym for `closed`.** Both close an entry under D1, but D4 needs the count of unrecognised models reportable — a stale registry must be observable rather than looking like a genuinely closed board. On the live board `moonshotai/kimi-k2.6` alone is the difference between 0% and 29%.

## Test plan

`run_gates.py scoreboard --base origin/main` — **ALL GREEN**: append-only, ruff check, ruff format, pyright, pytest at 88% coverage, three portal suites. 688 passing overall, of which **66** are the two new modules here; no existing test edited, so no append-only exception.

Mutation-checked — dropping the same-owner guard, wiping routes on an omitted replay, deleting the routing-prefix strip, and skipping the provider derivation on either path are each caught. The classifier's own precedence is now pinned directly rather than by mutation, since the rules are structural: a family exception only fires for its own owner, and crafted names are covered by explicit cases.

**Migration verified by hand**, because nothing in this repo applies migrations: the fixture builds the schema from the models, and neither the gate runner nor CI runs one. The full chain through `0013` was applied to a fresh SQLite database and the column lands.

## ⚠️ Deploy before releasing OME-1180

`ScoreSubmission` is `extra="forbid"`, so a released Client sending `models` to an un-upgraded board **422s every submission**. Merging this is not sufficient — it must be deployed and confirmed live first.

## One thing in the history

`5b10f5f6` committed its tests without their implementation and was red. The mutation harness restores with a checkout and ran against uncommitted work, so its trap reverted `store.py` before I staged. Restored unchanged in `77a426ce`, and recorded in the ledger's deviations. Harmless on `main` since this squash-merges, but worth knowing if you read the commits individually.

---

## Review round 2 — four findings, all fixed

Three P1s and one P2, all confirmed empirically before fixing. Two invalidated claims I had made without checking.

**`models` was not internal.** `ScoreSchema` is the response model for `POST /scores` and `GET /scores/{id}`, and it feeds the private JSONL export whose bytes `purge_private_benchmark.export_sha256` hashes to authorize a purge. `"models": null` changed every previously saved export, so a certified export could no longer authorize its own purge. Now carries `exclude_if`, like `ranking_notice` beside it. Verified byte-identical to `origin/main` for a legacy row: `ac966efd…b8a3`, 582 bytes both sides.

**Substring matching was exploitable.** Routes are client-submitted, so `openai/not-gemma-proprietary` classified **open** — along with `anthropic/kimi-wrapper` and `google/not-qwen-api`. Matching is now structural: owner matched exactly, family exceptions prefix-matched and scoped to their owner. No owner segment fails closed as `unknown`.

**Replay could replace a declaration.** `_content_hash` excludes `models`, so two submissions differing only in their routes share one identity, and the update was unconditional — a replay could swap what an entry is made of and flip its openness without changing its identity. Now fills only a null value.

**Three registered providers read as unknown.** `gemini-cli`, `codex` and `antigravity` were missing from the closed-owner set. No percentage changes, but it corrupts the D4 staleness count. Added, along with a test covering all seven owner-shaped registered providers. Also dropped `cohere` and `xai`, both of which I had added speculatively — neither is registered, and `xai` is not OpenRouter's namespace either.

One bookkeeping fix went to Linear rather than here: OME-1181's acceptance demanded the unrecognised-model count be *exposed*, contradicting its own scope table. Moved to OME-1145.

### DIFF
--- a/apps/scoreboard/src/scoreboard/classification/openness.py
+++ b/apps/scoreboard/src/scoreboard/classification/openness.py
@@ -46,37 +46,51 @@ _OPEN_BASELINE_MARKERS: tuple[str, ...] = ("llama", "mistral", "qwen", "deepseek
 # describe PROVIDERS and keep serving `classify_providers` unchanged.
 ModelOpenness = Literal["open", "closed", "unknown"]
 
-# WHY a separate list rather than more entries in _OPEN_PROVIDER_MARKERS: these are the
-# vendors that ship BOTH. Every marker here names a model family whose OWNER is on the closed
-# list, so it must be consulted before the owner rule or the owner wins and an open-weights
-# model is published as closed. That is the OME-1145 defect, and on the live draco-3pass board
-# `moonshotai`/`kimi` alone is the difference between 0% and 29%.
+# INVARIANT: model routes are CLIENT-SUBMITTED, so this is an adversarial surface and the rules
+# below are STRUCTURAL, never substring. An earlier version matched any open marker anywhere in
+# the route, which let a submitter name a proprietary model `not-gemma-proprietary` and be
+# published as open — the fail-closed contract inverted by choosing a string (review of PR #922).
 #
-# INVARIANT (OME-1179 Q1): membership here means "weights are downloadable and locally
-# runnable", NOT "permissively licensed". Gemma carries use restrictions and gpt-oss a usage
-# policy; both are still open under the chosen definition, because the board's claim is
-# reproducibility — can someone else run this and get your number.
-_OPEN_MODEL_MARKERS: tuple[str, ...] = (
-    "gpt-oss",
-    "gemma",
-    "moonshotai",
-    "kimi",
+# A route is `owner/model`. The OWNER segment decides, matched exactly.
+_OPEN_OWNERS: frozenset[str] = frozenset(
+    {
+        "meta-llama",
+        "mistralai",
+        "qwen",
+        "deepseek",
+        "moonshotai",
+        "huggingface",
+        # A locally-run model is the definition of downloadable-and-runnable under Q1.
+        "ollama",
+    }
 )
+_CLOSED_OWNERS: frozenset[str] = frozenset(
+    {
+        "openai",
+        "anthropic",
+        "google",
+        "gemini",
+        "cohere",
+        "xai",
+    }
+)
+
+# The vendors that ship BOTH. A family exception is SCOPED TO ITS OWNER and matched as a prefix
+# of the model segment, so `google/gemma-*` is open while `openai/gemma-anything` is not —
+# Google ships Gemma's weights, OpenAI does not.
+#
+# INVARIANT (OME-1179 Q1): membership means "weights are downloadable and locally runnable", NOT
+# "permissively licensed". Gemma carries use restrictions and gpt-oss a usage policy; both are
+# still open under the chosen definition, because the board's claim is reproducibility.
+_OPEN_FAMILIES_BY_OWNER: dict[str, tuple[str, ...]] = {
+    "openai": ("gpt-oss",),
+    "google": ("gemma",),
+}
 
 # WHY: a routing prefix says who carried the request, not what ran. Every live draco-3pass
-# route is `openrouter/`-prefixed and `openrouter` is a closed PROVIDER marker, so without
-# stripping this every model on the board classifies closed however open its weights are.
-_ROUTING_PREFIXES: tuple[str, ...] = ("openrouter",)
-
-# INVARIANT: ORDER IS THE RULE. `classify_model` returns on the first match, so the specific
-# model markers must sit above the owner markers — `gpt-oss` before `openai`, `gemma` before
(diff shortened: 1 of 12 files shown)


HupBaHa (changes requested):
Two blocking issues remain:

1. `scores/store.py`: replay enrichment checks `existing.models is None` before the transaction, but the eventual update neither re-reads/locks the Score nor requires `models IS NULL`. Two stale same-owner replays can therefore overwrite the first fill, violating Q3's fill-only contract.

2. `classification/openness.py`: family exceptions use unrestricted `startswith("gemma")` / `startswith("gpt-oss")`. Valid caller-controlled names such as `gemmalicious-proprietary` and `gpt-ossification-api` classify open. Match the required `gemma-*` / `gpt-oss-*` token boundary.

FilipBolt:
Both fixed.

**1. `store.py`** — `_apply_replay_updates` (`772e85e0`) re-reads the row under `select_for_update()` inside the transaction and recomputes `_replay_updates` against it, so the fill-only decision is taken on the locked row rather than a pre-transaction read. `test_the_replay_row_read_really_locks_the_row` renders the query on the asyncpg dialect, so the `FOR UPDATE` claim is checked rather than asserted in prose.

I deliberately did not add `models IS NULL` to the update filter. The three correctable fields share one statement and one `updated != 1` refusal, so that predicate would also reject a legitimate `authors`/`metadata` correction riding the same request. The method returns the settled values, so the response reports what the row holds rather than the claim this request lost.

**2. `openness.py`** — `_in_family` (`f01a1dca`) matches the family token exactly, or followed by `-`. Both names you gave classify closed now; `gemma-2-27b-it` and `gpt-oss-120b` are unchanged.

Four tests added, no prior test modified.

One call worth your eye: adding `replay_row_query` tripped `test_no_visibility_dependent_exit_is_unguarded_without_a_recorded_reason`, but only because its docstring named `visibility_query` and that guard selects on the string "visibility" appearing anywhere in the body. The function queries `Score` by identity and reads no visibility at all, so I removed the false positive instead of recording an exemption for it. Reasoning is in `docs/work/2026-09-11-OME-1181-model-identities.md` — push back if you'd rather see it on the list.

Rebased onto main, CI green.

HupBaHa (approved):
Re-reviewed at `c9212dce` against current `main` (`a78d58da`). Both prior blockers are resolved:

- replay enrichment now re-reads and locks the Score row inside the transaction, then recomputes the fill-only update against the locked value;
- Gemma and gpt-oss exceptions now require an exact family token or a `-` boundary, so crafted prefixes remain closed.

The synthetic merge is clean. The full Scoreboard suite passes (697 passed, 3 skipped), as do formatting, lint, type, and portal checks. No further findings.

PR #923, opened 2026-09-11
feat(screamingface): send the declared model routes on a submission

FilipBolt:
Client half of `OME-1179`. The submission payload gains a `models` array carrying `CandidateResult.models` verbatim.

## The problem

`_submission()` sends `_providers(candidate_result.models)`, and `_providers` keeps only the first path segment of each route. `openrouter/deepseek/deepseek-v4-pro` leaves as `"openrouter"`.

All seven live `draco-3pass` entries store `ran_with_providers = ["openrouter"]`, so an entry named **`best_open_source`** — deepseek, kimi, qwen — is published as **closed**, and the board reads 0% open.

## The change

One key:

```python
"models": list(candidate_result.models),
```

`ran_with_providers` is deliberately unchanged. It is required on `ScoreSubmission`, the portal's Backends column reads it, and the Scoreboard's `_content_hash` hashes the **wire** value — altering it would rewrite recipe identity for every existing row and break dedup across the board.

## Guards worth reviewing

**A fusion declares its synthesizer, not just its members.** The answer passes through the synthesizer, so omitting it would let a closed synthesizer hide behind open members — and under `OME-1179` D1 that flips the entry's published verdict from closed to open. Asserted at the source, `CandidateResult.models`, because that is where an upstream change would silently drop it.

**`models` and `ran_with_providers` describe the same set.** `OME-1181` derives the stored providers from these routes server-side, which only holds while the Client sends a consistent pair.

**Routes are sent whole.** A regression to `_providers`-style truncation is the entire bug this fixes, and it is invisible to any assertion that only checks the key exists.

## Approved Confidence-Gate exception

`test_the_submission_payload_gains_only_the_cost_key` asserts the payload's exact key set, so adding one breaks it. Owner approved adding the single string `"models"` (2026-09-11).

That guard's own comment says it exists so *"an accidental change to a neighbouring field should fail loudly rather than ship"*. This change is deliberate, so the guard worked as designed and recording the new key is the intended response — nothing removed, no assertion loosened, the set stays exhaustive.

Blast radius was measured before asking rather than guessed: exactly **one** failing test out of 1466.

Rejected alternative: nest `models` inside `metadata`, as `benchmark_revision` already is. That touches no existing test and removes the deploy-ordering constraint entirely, since any Scoreboard version ignores unknown metadata keys — but it leaves the field untyped on the wire, which is the opposite of why a typed field was chosen, and would mean reopening `OME-1181` after its review.

## Test plan

`run_gates.py screamingface --base origin/main --skip-append-only` — **ALL GREEN**: ruff check, ruff format, pyright, pytest at **95%** coverage, the deterministic notebook check, `uv build`, and the distribution check. Full suite **1472 passed / 23 skipped**.

The plain run fails the append-only check on `tests/test_leaderboards.py` — that is the recorded exception above and nothing else.

The public surface snapshot is **untouched**, as expected: `_submission` is private, and this changes a wire payload rather than the Python API.

## 🚨 Do not release before OME-1181 is deployed

`ScoreSubmission` is `extra="forbid"`, so a released Client sending `models` to an un-upgraded Scoreboard **422s every submission**.

`packages/screamingface` is on release-please, so merging this branch opens a release PR — and merging **that** publishes to PyPI. The gate is on the release PR, not on this one. `OME-1181` (#922) must be merged **and deployed and confirmed live** first.

### DIFF
--- a/docs/plan/2026-09-11-OME-1180-send-models.md
+++ b/docs/plan/2026-09-11-OME-1180-send-models.md
@@ -1,0 +1,41 @@
+---
+title: Send the declared model identities — plan
+ticket: OME-1180
+spec: docs/spec/2026-09-11-OME-1180-send-models.md
+status: approved
+date: 2026-09-11
+---
+
+# OME-1180 — implementation plan
+
+One phase. The change is a single payload key; the work is in the guards around it and in the
+one approved edit to an existing test.
+
+## Steps
+
+1. **RED.** Add to `packages/screamingface/tests/test_leaderboards.py`:
+   - the full routes are sent verbatim — a regression to provider prefixes fails
+   - `models` and `ran_with_providers` describe the same set, so the two cannot drift
+   - a **fusion** sends its members *and* its synthesizer, not just the members
+   - a solo sends exactly one route
+   - the payload still survives `json.dumps`, as the cost tests require of theirs
+2. **GREEN.** `"models": list(candidate_result.models)` in `_submission()`. Already written.
+3. **Rule-5 edit (approved).** Add `"models"` to the expected key set in
+   `test_the_submission_payload_gains_only_the_cost_key`. One string; nothing removed, nothing
+   loosened, the set stays exhaustive.
+4. **CHANGELOG.** An entry under `## Unreleased`.
+5. **Gates.** `run_gates.py screamingface --base origin/main`.
+
+## Gates for this stack differ from the scoreboard's
+
+Coverage is **95%**, not 80%, and there are three extra gates: the deterministic notebook check,
+`uv build`, and the distribution check. `uv sync --all-extras` is needed first or the suite
+fails to collect on a missing `ipywidgets`.
+
+## Stop conditions
+
+Return to the owner rather than working around, if:
+
+- any existing test other than the one approved in the spec needs to change
+- `CandidateResult.models` turns out not to include a fusion's synthesizer
+- the public surface snapshot moves (it should not — `_submission` is private)
--- a/docs/spec/2026-09-11-OME-1180-send-models.md
+++ b/docs/spec/2026-09-11-OME-1180-send-models.md
@@ -1,0 +1,78 @@
+---
+title: Send the declared model identities on a leaderboard submission — spec
+ticket: OME-1180
+status: approved — scope inherited from OME-1179; rule-5 exception approved 2026-09-11
+date: 2026-09-11
+parent: OME-1179
+---
+
+# OME-1180 — send the declared model routes
+
+## 1. Problem
+
+`_submission()` (`_scoreboard/leaderboards.py:443`) sends `_providers(candidate_result.models)`,
+and `_providers` (line 502) keeps only the first path segment of each route. The identities the
+recipe declares never reach the board.
+
+## 2. Change
+
+One key on the payload:
+
+```python
+"models": list(candidate_result.models),
+```
+
+`CandidateResult.models` is required, ordered, unique and non-empty, so it is always available
+and needs no fallback.
+
+`ran_with_providers` stays exactly as it is. It is required on `ScoreSubmission`, the portal's
+Backends column reads it, and `_content_hash` on the Scoreboard hashes the wire value — changing
+or removing it would rewrite recipe identity for every existing row.
+
+## 3. Declared, not observed
+
+These are the routes the recipe composes. `_evaluation/results.py:141` builds
+`CandidateResult(models=candidate.models)` from the composed candidate, never from the run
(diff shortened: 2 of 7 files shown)


HupBaHa (approved):
Implementation matches OME-1180 and preserves the existing provider and identity contracts. Approved for code quality; do not release until PR #922 is fixed, deployed, and confirmed live.

FilipBolt:
## What changed since the approval

Approved at `bff90ebf` on 2026-09-16. Head is now `17ab979a`. Two things happened, one of them adding a ticket.

| | |
|---|---|
| Rebase | onto current `main` — `origin/main` had moved 176 commits |
| New commit | `17ab979a` — **OME-1247**, the bounds guard |

The first two commits are unchanged from what you approved. **Only `17ab979a` is new code.**

## The new commit, in one paragraph

The Scoreboard caps `models` at 32 routes / 255 characters each / 4096 bytes serialized. The Client enforced none of them. Once this PR starts sending the field, a candidate over any cap turns a previously valid submission into a **422 on the whole submission** — `models` fails validation and takes `ScoreSubmission` with it.

`_submission_models()` mirrors `_submission_authors()` sitting directly beside it: same class of field, same class of cap, same `ValueError`. Three constants next to `_MAX_AUTHORS` / `_MAX_AUTHOR_LENGTH`.

## Why it is on this PR rather than its own

A guard and the field it guards are one change. Both alternatives were worse:

- **guard on `main` first** → regression window. A candidate with 33 distinct routes is accepted *today* precisely because the field is not transmitted, and would start being rejected for a field still not sent.
- **this PR first, guard after** → ships the defect, and merging this opens the SDK release PR whose merge publishes to PyPI.

## Three decisions worth your eye

**It raises; it does not drop the field.** Omitting `models` when over-cap would let the submission succeed and the entry be classified from `ran_with_providers` — which is the OME-1145 bug this whole chain exists to fix. A visible local failure beats an invisible wrong answer on the public board.

**It sits at the submission boundary, not on `Pipeline`.** The caps are the leaderboard's, not the toolkit's. Composing a 40-model ensemble locally stays legal; only publishing it to a board that will refuse it does not.

**`import json` became `from json import dumps as _json_dumps`.** Not required by the change. `_sync_json` and `_async_json` both take a parameter named `json`, so adding the module import created a shadow inside exactly the two functions most likely to want it. Flagging it because it is the one edit that is not about bounds.

## The subtle test

`test_the_byte_cap_is_measured_the_way_the_board_measures_it` pins **4096 exactly** rather than comparing the two serializations.

My first draft asserted a payload under the cap by the board's spelling and over it by the default one. That fixture cannot exist for a uniform payload: the gap between the spellings is `len(models) - 2` bytes while one character of route moves the total by `len(models)`, so no length straddles them. Pinning the boundary is stronger anyway — the same payload measures 4115 under default separators, so a wrong-separator implementation fails the test.

## Scale, measured rather than assumed

`CandidateResult.models` is deduplicated at every accumulation path (`_ordered_unique` at `candidate.py:294`, `:326`, `:359`), so the cap is on **distinct** routes. The live maximum on any board is 4. Reaching 33 means a candidate eight times larger than anything that has run — real, remote, a correctness gap rather than an outage.

## Verification

`run_gates.py screamingface --base origin/main --skip-append-only` — **ALL GREEN**: ruff, format, pyright, pytest at **95%** coverage, the deterministic notebook check, `uv build`, the distribution check. Full suite **1656 passed / 26 skipped**. CI on this head: **39 checks, 0 failures**.

Six tests added, all appends. Every cap pinned on both sides — 32 passes / 33 raises, 255 passes / 256 raises, 4096 passes / 4097 raises.

The `--skip-append-only` is **OME-1180's existing recorded exception** (the single string `"models"` added to `test_the_submission_payload_gains_only_the_cost_key` on 2026-09-11), not a new one. Verified against the diff: every line this unit added is a pure append.

## Release gate is unchanged

Do not release before OME-1181 is deployed. It **is** live on dev — verified against the deployed OpenAPI, which now lists `models` on `ScoreSubmission` with `additionalProperties: false`. Production ships separately through release-please PR #511, which is open and unmerged.

HupBaHa (approved):
Re-reviewed the current head after the rebase and added OME-1247 scope. The previously approved OME-1180 commits are semantically unchanged, and the new client-side guard mirrors the Scoreboard contract exactly at 32 routes, 255 characters per route, and 4096 compact UTF-8 JSON bytes without transforming valid payloads. Synthetic merge with current main is clean and the full screamingface gate suite passes on Python 3.12. No code findings. Release gate remains: do not publish the SDK until Scoreboard release PR #511 is merged, deployed, and confirmed live; dev already advertises the models field, but production is not established.

PR #924, opened 2026-09-11
feat(engine): retain Client version in ephemeral run evidence

keelancj:
The Client now identifies itself through User-Agent, but the Engine previously discarded that version when starting a run. Capture one bounded ScreamingFace product/version token and emit it as an ordinary sequenced INFO log event immediately after the root Started event.

The event carries `screamingface.client.version` and follows existing run-stream replay and cleanup. Both local execution and queued workers carry the value per run; workers clear any inherited ambient version. Missing, malformed, ambiguous, or oversized headers remain compatible and produce no version evidence. No raw header, new database, typed Report field, or response schema is introduced.

Validation:
- 25 new parser, lifecycle, REST admission, isolation, queue/worker and cleanup cases.
- Full Engine suite: 2,831 passed, 9 skipped; 93.23% coverage against an 80% requirement.
- Final gate runner passes lint, format, pyright, layering and full tests/coverage.
- Actual Engine Started/Log frames decode through the Client merged in PR #918.

Approved process exception: the owner approved adding only the optional `client_version` argument to four existing test doubles. Exact comparison confirms their assertions and bodies are unchanged. The final gate run uses `--skip-append-only` solely for this approved signature exception.

Retention remains backend-specific and ephemeral. Client-held copies and separately retained diagnostic data have independent lifetimes; this change does not add version data to Python logs or exported spans. Typed Report provenance remains separate follow-up work under OME-416.

Refs: OME-1191

## Review refresh — 2026-09-16

Rebased onto main `f31084a5`, resolving nine conflict files while preserving both answer_seed and client_version. Added combined REST/local and queue/worker regression checks. Main's new answer-seed test double also needed the optional client_version argument: two existing tests reproduced the incompatibility before the signature-only correction. No existing assertions were changed.

Addressed Khoa's small notes: clarified that a directly launched runner trusts its process environment, and moved RecordingJobRunner into the existing module-level import. UA comment behavior is unchanged; shared helper extraction and OME-416 remain separate work.

Validation: 50 focused provenance/answer-seed tests passed; full Engine gates passed with 93.35% coverage. The documented append-only exception also covers the approved import relocation and fifth test-double signature adjustment. Independent integration review found no material issues.

Head: `6e09d9b3`. Ledger: `docs/work/2026-09-16-OME-1191-review-refresh.md`. PR remains open; refreshed CI runs after push.

### DIFF
--- a/apps/screamingface-engine/src/screamingface_engine/adapters/inprocess.py
+++ b/apps/screamingface-engine/src/screamingface_engine/adapters/inprocess.py
@@ -18,6 +18,7 @@ from collections.abc import Callable, Mapping, Sequence
 from functools import partial
 
 from screamingface_engine import job_env
+from screamingface_engine.client_provenance import ProvenanceExecutor
 from screamingface_engine.ports import IdentityAwareJobRunner
 from screamingface_engine.run_evidence import (
     TerminalWatch,
@@ -223,6 +224,7 @@ class InProcessJobRunner(IdentityAwareJobRunner):
         profile: str | None = None,
         identity: Mapping[str, str] | None = None,
         cache: CachePolicy | None = None,
+        client_version: str | None = None,
     ) -> str:
         """Spawn the run as a task and return its job name.
 
@@ -250,7 +252,7 @@ class InProcessJobRunner(IdentityAwareJobRunner):
         task = asyncio.get_running_loop().create_task(
             lifecycle_run(
                 watch,
-                executor,
+                ProvenanceExecutor(executor, client_version),
                 topic,
                 url4,
                 traceparent=run_traceparent,
--- a/apps/screamingface-engine/src/screamingface_engine/adapters/queue_runner.py
+++ b/apps/screamingface-engine/src/screamingface_engine/adapters/queue_runner.py
@@ -278,6 +278,7 @@ class QueueJobRunner(IdentityAwareJobRunner):
         profile: str | None = None,
         identity: Mapping[str, str] | None = None,
         cache: CachePolicy | None = None,
+        client_version: str | None = None,
     ) -> str:
         """Publish the run to the queue and return its job name.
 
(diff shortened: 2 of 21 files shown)


khoaguin:
**Review: approve.** No blockers found on any dimension I checked: worker isolation is correct (`worker/supervisor.py:264` pops the inherited env key before overlay, pinned by a test asserting absence, not just value); the parser is ReDoS-safe (512-char length gate before a single linear character-class `fullmatch`); injection is impossible (value restricted to `[A-Za-z0-9._+-]{1,128}`, raw header never retained, event body constant); no new store/schema (the attribute rides `LogData.attributes`); and `client_version` is threaded routes → port → both adapters → queue env → runner with no accepted-but-dropped parameter anywhere in production code.

Two forward-looking notes, neither blocking:

1. **Whole-header rejection on UA comments** — `client_provenance.py:80`: any `(`/`)` in the header discards the entire thing, so a future Client or intermediary appending a standard UA comment (`screamingface/1.4.0 (darwin; arm64)`) silently loses provenance for every run — and nothing surfaces the regression because "no evidence" is also the legitimate old-client outcome. Spec owns this choice; for OME-416, consider stripping comments instead of discarding, or a Client-side test pinning "our UA never grows a comment".
2. **Runner-side ambient read** — `runner/main.py:188`: `params_from_env` reads `URL4_CLOUD_CLIENT_VERSION` from whatever env the runner has; only the queue-worker spawn path clears ambient values, so a runner launched directly with the var exported would stamp it onto every run. Theoretical today (namespaced var, single spawner) — one sentence in the spec's isolation paragraph would cover it.

Mechanical: `tests/unit/test_client_provenance_wiring.py:21-28` is the first module importing private helpers from five sibling test files — a rename in `test_rest._cap` during an unrelated refactor would collect-error four provenance tests pointing at the wrong ticket; promote the shared helpers into a conftest/`_fakes.py` in a follow-up. Line 485 has an in-body `from _fakes import ...` the module already does at top level.

PR #927, opened 2026-09-12
feat(screamingface-engine): let a run declare an answer seed

khoaguin:
A run can now declare an answer seed, so a score can be published as mean ± CI over N labelled sittings and any sitting replayed exactly. Spec + implementation for OME-1038 (owner approved seed-per-run).

## Problem / Solution

**Problem:** Run the same exam twice and the student may score differently — models are nondeterministic. Every leaderboard number was one sitting shown as the truth: nobody could say how much it wobbles, and nobody could re-create the run behind it.

**Solution:** A run declares an answer seed per run. Every answer call carries it, identical sittings produce identical requests byte-for-byte, and N seeded runs are N honest samples.

answer-seed-lifecycle drawio

## Before / After

The before/after mechanism diagram lives on [OME-1038](https://linear.app/openmined/issue/OME-1038/a-run-cant-declare-answer-seeds-so-score-variance-cant-be-measured-or) (red TODAY lane / green AFTER lane).

### Before
- Trigger: a researcher wants to publish a score as mean ± CI by running the same cases N times.
- Today: no answer-seed exists; two "identical" runs can't be told apart or reproduced.
- Cost: every published number is a single sample; a carelessly-added seed field would invalidate every request-keyed replay fixture.

### After
- Same trigger.
- Now: `X-Answer-Seed: 7` on run start; the run's env carries it; the connector stamps `seed=7` onto every answer call; same seed ⇒ byte-identical requests ⇒ one reproducible cache slot per sitting.
- Win: honest variance sampling and exact replay — with **zero fixtures re-recorded**, because an undeclared run's egress is byte-identical to today's.

### Don't regress
- A run declaring nothing changes nothing: asserted as `set(body) == {"model", "messages"}` and byte-compared request content.
- A call pinning its own `seed` param always wins — the draco judge's per-pass seeds are never re-keyed (asserted end-to-end).
- `benchmarks/spine/` is untouched (the spec's deletion test — 0 files in the diff).

## New runtime behavior (review attention here)

The seed is an **ambient per-run value**, not an expression change — the same journey `profile` and the cache policy make (REST edge → job env → connector). Two deliberate rules carry the correctness:
1. **Absence is the default.** `None` is a no-op at every hop (renders no env var, adds no param), which is what keeps replay fixtures valid.
2. **An explicit `seed` param wins.** Merged after the retrieval ceiling, only onto calls that state no seed of their own.
A malformed seed **refuses the run** (400 at the edge; `RunnerConfigError` at the runner) — executed unseeded, it would publish a score claiming a sitting it never had.

## Review order

1. `apps/screamingface-engine/src/screamingface_engine/runner/request_parameters.py` — the core idea: the one merge function and its two invariants (no-op absence, explicit-seed-wins). Verify the judge-protection rule reads on `params`, not just the request.
2. `apps/screamingface-engine/src/screamingface_engine/runner/connector.py` — the payoff: where the merge lands on every answer call, after the retrieval ceiling, with call identity still computed pre-policy (attribution unaffected). Verify the per-run field sits on the endpoint, never on the shared world config.
3. `apps/screamingface-engine/src/screamingface_engine/job_env.py` + `rest/routes.py` — the run contract: the env pair (why the reader raises where cache/io readers are total) and the `X-Answer-Seed` header validation.
4. `ports.py`, `adapters/inprocess.py` (note the ambient-reset invariant), `adapters/queue_runner.py`, `runner_queue.py`, `runner/main.py` — the mechanical port ripple; each mirrors `profile`/`cache` exactly.
5. `tests/unit/test_answer_seed_threading.py` — the contract spec: 21 tests covering byte-identity, judge protection, env round-trip, ambient-leak reset, header validation, and per-seed cache-slot separation.
6. Prior test files (`_fakes.py`, `test_cache_policy_threading.py`, `test_queue_admission.py`, `test_e2e_compose_flow.py`) — signature-only widenings of fakes to the new port kwarg (owner-approved); confirm no assertion changed.
7. `docs/spec/2026-09-12-answer-seeds-spec.md` + ledger + mirror — the decision record; skim.

**Also in this PR — OME-1193 (SDK half, commit `9d389103`):** `evaluate(answer_seed=…)` sends `X-Answer-Seed` on run start (Recipe evaluations and url4 replays alike) and `report.json` names the sitting per candidate record (`answer_seed`, always present, null when unseeded). The seed rides the compiled Candidate the run-transport already receives, so the transport protocol and every existing fake are untouched. Public surface widened deliberately: changelog entry + regenerated surface snapshot. Read `packages/screamingface/tests/test_answer_seed_report.py` as that half's contract spec.

Out of scope (follow-up per the cross-cutting rule): scoreboard mean ± CI aggregation over sibling reports.


keelancj (commented):
@khoaguin  2 things flagged from review

- `connector.py:29`) adds seed unconditionally. Anthropic’s gateway contract rejects it. Adding seed=7 raises UnsupportedParametersError, the mocked gateway hides this failure

- The same injection affects HealthBench and GDPVal judges, whose parameters contain no explicit seed. Changing the answer seed therefore changes judge requests and cache keys under unchanged benchmark revisions

khoaguin:
Both review findings confirmed and fixed in `a9159f1b` — thanks, they were real:

1. **Judge re-keying** (the deeper one): HealthBench/GDPVal judges pin params with no `seed`, so the ambient stamp re-keyed their calls per sitting under an unchanged revision. The seed is now scoped to the **Candidate invocation** — a task-local flag the Candidate adapter raises beside `retrieval_scope`, read by the connector — so benchmark-authored grading calls can never carry it, on any board. Pinned by `test_the_seed_reaches_candidate_calls_and_never_judge_calls`, which reproduced your finding red before the fix.
2. **Anthropic `UnsupportedParamsError` behind mocks**: fixed pre-spend on the SDK side — when a seed is declared, the existing model-parameters preflight now validates `seed` for every candidate model and refuses with `unsupported_model_parameter` before any token is minted. Unseeded evaluations never demand seed support (no regression for seed-less routes).

Consequence worth knowing: the seeded-egress tests now run through `/benchmarks/candidate`, and a bare (non-candidate) model call is deliberately never ambient-seeded — that's the judge's shape. Both stacks' gates green (engine 2898, SDK 1488).

khoaguin:
Thanks @keelancj — both findings were real, and both would have hurt in production. Both are fixed in a9159f1b (`fix(screamingface-engine): scope the answer seed to the Candidate invocation`); writing them back in plain words to make sure I understood them right:

**Finding 1 — the seed was leaking onto the judge.** A run is an exam: the candidate writes answers, a benchmark-hired judge grades them. Both talk through the same connector, and the connector stamped the declared `seed` on *every* call — it had no way to tell a student from an examiner. Since the gateway caches on the whole request body, a seeded run would silently re-key every HealthBench/GDPVal judge call per sitting under an unchanged benchmark revision: grading identity varying with the answer seed.

*Fix (in a9159f1b):* a task-local flag — new `candidate_scope.py` (ContextVar, so concurrent local-mode runs can't read each other's scope), raised by `benchmarks/candidate_adapter.py` around the answering expression only, checked by `runner/connector.py` before applying the ambient seed. Pinned by `test_the_seed_reaches_candidate_calls_and_never_judge_calls`, which runs one candidate call and one judge call in the same seeded run — candidate body gets `seed: 7`, judge body has no `seed` key at all.

**Finding 2 — Anthropic rejects `seed`, and the mock gateway was too polite to say so.** The real gateway's own contract already knows this (the anthropic provider deliberately omits `seed` — honest support only), but the mocked gateway 200s everything, so `evaluate(answer_seed=7)` on a Claude candidate would have started a run, spent real money, then died mid-run at the provider.

*Fix (same commit, SDK side):* the seed now rides the model-parameters preflight the explicit params already use — `_evaluation/model_parameters.py` validates an enabled `seed` in every candidate model's contract whenever a seed is declared, refusing the whole evaluation with `PlanningError(unsupported_model_parameter)` before any token is minted: loud and pre-spend, never a silently unseeded sample. No seed declared → no new check, so today's seed-less models keep evaluating exactly as before (pinned by `test_a_seeded_evaluation_is_refused_when_a_candidate_model_lacks_seed` and its proceed/unseeded twins).

Both invariants now live in tests, so a refactor that regresses either fails CI instead of a leaderboard. Genuinely good catches — the second one especially, since the tests were structurally incapable of seeing it.

PR #930, opened 2026-09-14
feat: standard metadata (cost, latency, tokens) on every cached response

IonesioJunior:
## What

The global response cache stored only the provider body, so a hit could not say what the call would have cost and a deployment could not report what the cache saved. This adds a standard metadata block beside every cached row, reads it back on a hit through `_aigw`, and totals it into a run-level saved-cost figure.

Migration `0011` appends a **nullable `metadata_json`** column (~400 bytes: canonical token usage, direct cost, provider latency, response model). On Postgres it is a catalog-only change — no rewrite, no long lock, no backfill.

## Load-bearing decisions

- **A column, not a field in `response_json`.** The cached row must stay provider-compatible for every future replay, and `response_size_bytes` measures what the store is given — embedding the block could push an otherwise cacheable response over the cap and silently cost a cache entry.
- **The block is built from the `AccountingSession`, not a rendered `_aigw`.** The store still runs *before* `attach_success_metadata`; that ordering is locked and a test pins it.
- **The cache key is unchanged.** Cost is an output, so it is never part of the key — otherwise the corpus would silently partition by price.
- **Unknown is not free.** A provider that reports no cost still gets its row, with an explicit `absent`/`unavailable` status and a null amount — never `0`.
- **Two saved-cost totals, never one.** `cache.saved_cost_usd` holds provider-authored money; `cache.saved_cost_archive_usd` holds `archive_matched` money (real, but not provably that row's). No code path combines them, and a test asserts the absence of a third accumulator. Neither changes `cost_usd`, and neither reaches the leaderboard.

## Deliberate override of a documented rule

`docs/usage-accounting.md` said cached monetary evidence can never prove raw-JSON provenance. That is true of `response_json` and **not** true of a block captured at write time from the raw provider response, before any conversion. The doc now carries that carve-out explicitly, and records the still-open OpenAI/HuggingFace question (OME-884 / OME-791) that a truthful stored block reopens.

## Compatibility

Snapshot archives stay restorable both ways: the import path accepts the legacy 12-column COPY header and loads `NULL`, the export writes 13, and the merge carries `metadata_json` through `DO UPDATE SET` so a snapshot cannot silently drop it. Legacy rows, seed rows with no block, Tavily rows and unreadable blocks all fall back to the provider's existing mapper — no cost is ever inferred from the cached body.

Every step is best-effort: a build, serialize or parse failure writes `NULL` and still serves the answer, warning once per key with the key prefix alone.

## Also fixed here

`cache_saved_cost_usd` is the first `Decimal` ever carried on `SpanData`, and OTel attributes admit no such type — the OTLP encoder rejected it and only logged, so the attribute vanished from every exported span. Money is now rendered as an exact decimal string.

## Verification

All gates green on every touched stack:

| stack | ruff | format | pyright | tests |
|---|---|---|---|---|
| aigateway | ✅ | ✅ | ✅ | 4358 passed, 92.35% |
| screamingface-engine | ✅ | ✅ | ✅ (+layering) | 2851 passed, 93.23% |
| url4 | ✅ | ✅ | ✅ | 1250 passed, 96.53% |

`AIGW_TEST_PG=1 pytest -m needs_postgres` — **25 passed against real Postgres**, including the real `pg_dump` round-trip on the new 13-column layout.

## Notes for the reviewer

- The `archive_matched` status and the separate archive total are gateway-side support for a corpus that is priced and loaded **out of band**. No seed package, generator or corpus is included in this PR, and `draco-cache-seed-*/` is now gitignored so the local working data cannot be committed by accident.
- PRD test 23 is specified at `level: property`. It ships **example-based** — the repo has no property-testing dependency and this change did not seem the right place to introduce one. The matrix was broadened instead (sub-cent amounts, more significant digits than a binary float holds, the latency boundaries).


HupBaHa (changes requested):
Reviewed exact head f06e9ecf against current main c9c6761f. Requesting changes for the following correctness issues:

1. **High — rolling upgrade/rollback can pair stale metadata with a replaced response** (`apps/aigateway/src/aigateway/core/request_cache/bulk_loader.py:172`). Migration 0011 leaves old binaries operational against the widened table, but an old loader still updates `response_json` without updating or clearing `metadata_json`. During pod overlap or rollback, loading a legacy snapshot can therefore replace the body while preserving the previous body metadata. A PostgreSQL reproduction ended with `response_json=new-body` beside `metadata_json={cost_for:old-body}`. Fence legacy loaders/schema versions or make old writes fail safely; the new-loader-only staging repair cannot protect this path.

2. **High — migration 0011 can queue cache traffic behind a snapshot export** (`apps/aigateway/src/aigateway/migrations/0011_cache_entry_metadata.py:92`). `ADD COLUMN` is catalog-only but still needs `ACCESS EXCLUSIVE`. The exporter can hold a conflicting relation lock for up to its 600-second COPY timeout (`snapshot_export.py:273-275`); once the DDL waits, later cache SELECTs queue behind it. A PostgreSQL 16 probe confirmed a waiting AccessExclusiveLock and a subsequent SELECT timing out. Add a bounded `lock_timeout`/retry or quiesce exports, and remove the current “no long lock” runbook claim.

3. **High — a transport retry can report zero spend plus savings after a billed first attempt** (`apps/screamingface-engine/src/screamingface_engine/runner/connector.py:146`). `_post_completion` retries a lost response, but only the final response reaches accounting. If attempt 1 was processed and billed, then attempt 2 hits the row it populated, the run reports current cost `0` and `cache.saved_cost_usd` equal to the first charge. A direct MockTransport probe produced 2 requests, simulated actual spend `0.25`, reported spend `0`, and saved cost `0.25`. Carry transport-retry uncertainty into accounting instead of treating the final hit as the whole operation.

4. **Medium — `metadata_degraded` races with concurrent cache fills** (`apps/aigateway/src/aigateway/core/request_cache/bulk_loader.py:207`). The count and upsert are separate statements under READ COMMITTED. A priced row inserted after the count is omitted from the reported total and then overwritten with NULL by the merge. A PostgreSQL interleaving reproduced `metadata_degraded_count=0` with final `metadata_json=NULL`. Derive the count atomically from changed rows or use appropriate locking/isolation with retry.

5. **Medium — valid maximum-precision saved costs are silently rounded while accumulating** (`apps/screamingface-engine/src/screamingface_engine/runner/cache_counters.py:104`). Single-value conversion deliberately uses a 53-digit local context, but `total + amount` uses ambient Decimal precision (28 by default). Adding two valid boundary amounts produced `1999999999999999998.000000000` instead of `1999999999999999998.000000000000000000000000000000002`. Both run and span totals share this helper, so their reconciliation tests agree on the same wrong value.

6. **Medium — the public saved-cost seam accepts invalid values and can total them** (`packages/url4/src/url4/observe.py:177`). `ModelResponse.__post_init__` checks only amount/provenance pairing; it accepts negative amounts and arbitrary provenance strings. A negative amount with `reported` is accumulated into negative savings, while a positive amount with `bogus` is silently counted as unpriced and discarded. Validate the exact provenance vocabulary and non-negative Decimal domain at the observation/wire boundaries (`signals.py:111-124`).

7. **Low — the tracked TypeScript API schema is stale** (`apps/aigateway/src/aigateway/core/admin_schemas.py:166`). The response adds `AdminCacheJobOut.metadata_degraded`, but `apps/aigateway-ui/src/lib/aigateway/schema.d.ts:840-885` omits it, so typed admin clients cannot consume the new signal. Regenerate and commit the API declaration.

The PR is CLEAN/MERGEABLE and all GitHub checks pass. Independent suites also passed (AIGateway focused 179, Engine 2,859, url4 1,255), and the merge tree is clean; the findings above are uncovered edge cases, not existing CI failures.

IonesioJunior:
All seven findings are addressed. Tracked as **OME-1203**; 12 commits on top of `f06e9ec`, head `44be27d7`.

Thank you for the review — three of these were genuinely load-bearing, and two of them changed the design rather than just the code.

## Finding by finding

| # | What changed | Pinned by |
|---|---|---|
| 1 | A Postgres `BEFORE UPDATE OF response_json` trigger clears `metadata_json` when the body changes without it. Closed in the database rather than in the loader, because the ordinary cache-fill upsert has the same hole — a loader-side fence would have closed the instance and left the class. Degrades to unknown, never to a wrong number. | `test_replacing_a_response_alone_clears_its_metadata_block`, plus a Postgres forward→reverse round trip asserting `pg_trigger` **and** `pg_proc` are both empty after rollback and the table stays writable |
| 2 | `SET LOCAL lock_timeout = '3000ms'` before the DDL. **No retry** — see below. | `test_the_postgres_ddl_runs_under_a_bounded_lock_timeout`, plus an integration test driving the real migration via Tortoise's `migrate()` against a held lock and asserting SQLSTATE `55P03` |
| 3 | `avoided_usd_for_outcome(aigw, cache)` in `runner/accounting.py` withdraws the savings claim when the round trip was retried. The hit is still a hit and lands in `unpriced_hits`; no amount rides with it. | `test_a_hit_after_a_transport_retry_is_not_priced` and its no-retry contrast, plus `MockTransport` tests pinning `_post_completion`'s own `attempt > 0` production |
| 4 | `LOCK TABLE … IN SHARE ROW EXCLUSIVE` at the top of the merge transaction, so the count and the merge see one state. | `test_a_racing_write_to_an_unrelated_row_blocks_the_merge` — deliberately holds a key the archive does **not** mention, so the merge's own row contention cannot produce the block |
| 5 | `_accumulated` now adds under the same `AMOUNT_PRECISION` context the single-value conversion uses; `_AMOUNT_PRECISION` promoted to public so both name one constant. | `test_accumulating_two_maximum_precision_amounts_rounds_nothing` — asserts a literal, never `bound + bound`, since recomputing would use the same rounding it tests |
| 6 | Provenance vocabulary and a finite non-negative domain enforced on `ModelResponse.__post_init__` and on `SpanData` via `@model_validator`. The vocabulary is derived from the `SavedCostProvenance` alias with `get_args`, not restated. | Five tests, each `pytest.raises(..., match=)` pinned to the validator's own message |
| 7 | Regenerated. Additive only — 171 insertions, 0 deletions. | `tsc --noEmit` clean, 240/240 UI tests |

## Where we deviated from the suggestion, and why

**Finding 2 ships a bounded timeout with no retry.** Our first attempt followed the suggestion literally: `Migration.atomic = False` plus five attempts. Our own review caught that this silently costs the **downgrade** its all-or-nothing property — Tortoise wraps operations individually only when the migration is not atomic (`migrations/migration.py:140-180`), so the three reverse operations went from one transaction to three. An interruption after the column drop but before `_restore_sqlite_indexes` completes would leave the column gone *and* the indexes missing, where it previously rolled back whole. Neither downgrade test can see it, since both exercise only the uninterrupted path.

We reverted it. The retry was also weak on its own terms: five attempts at 3s with 5s backoff is ~40s against an exporter that sets `lock_timeout = 0` and can hold for its full 600s COPY — it would nearly always exhaust anyway. A bounded timeout alone satisfies the actual requirement (don't queue readers); failing loudly and rerunning is the safe direction. `SET LOCAL` rather than `SET`, so it cannot leak into the session.

## Two things our own review found that your findings surfaced

Both landed on the **cache-hit read path**, which no finding's fix touched directly — `record_hit_metadata` bumps `hit_count` inline before the cached body is returned, so *serving a hit is a write*.

- **The finding-1 trigger would have made every cache hit expensive.** A bare `BEFORE UPDATE … FOR EACH ROW` evaluates its `WHEN` qual per candidate row, so the hit-count bump ran `NEW.response_json IS DISTINCT FROM OLD.response_json`. Same value, same length, so `texteq`'s length shortcut doesn't apply — it detoasts both sides and `memcmp`s them, on values capped at 1 MB with `EXTENDED` storage. Fixed with the `OF response_json` column list: the bump doesn't name that column, so the trigger is skipped entirely rather than evaluated cheaply.
- **The finding-4 lock blocks cache hits, not just cache writes** — again because a hit is a write. It blocks rather than errors, so the `except Exception` that exists to serve the hit anyway never engages. We bounded the merge's *acquisition* (`LockNotAvailableError` → a `merge_lock_timeout` refusal rather than a generic failure), and documented plainly that hits stall for as long as the merge *holds*, which no timeout can fix. `DEPLOYMENT.md` now calls a merge against live traffic a maintenance-window operation.

Also corrected while in those sections: migration 0011 was **not idempotent on Postgres** — `CREATE OR REPLACE FUNCTION` but a bare `CREATE TRIGGER`. `test_0011_is_idempotent` passed only because it runs on SQLite, where the whole guard no-ops. Now `CREATE OR REPLACE TRIGGER`, pinned by a test that re-executes the SQL against a live migrated database.

## What is NOT closed

**Finding 3 removes the false claim, not the underlying gap.** A billed-then-lost attempt still reports `0` spend — we withdraw the savings claim because we cannot know the true cost, we do not reconstruct it. Properly fixing attribution needs gateway-side idempotency keys so the second attempt is recognised as the same operation. That is a new gateway contract and a separate ticket, not a review-round change.

Two accepted limitations, both documented in the code:

- The trigger has a deliberate false positive: a writer replacing the body with a byte-identical block loses it. The block carries provider latency, so collision is vanishingly rare, and the cost is one row degraded to "unknown" — the safe direction.
- `metadata_degraded` is exact *modulo* that false positive: a merge row whose incoming block is byte-identical but whose body differs is cleared by the trigger and not counted by `_DEGRADED_COUNT_SQL`. Noted on ERD invariant E11.

**PRD test 23** still ships example-based against a `level: property` requirement — unchanged this round, still recorded as open.

The schema regeneration also picked up two previously-unregenerated Tavily retrieval-cache endpoints. Pre-existing drift in the committed output rather than new API surface; included because hand-patching one field would have left a file no regeneration reproduces. Additive only.

## Verification

| stack | result |
|---|---|
| url4 | 1261 passed, 97.62% |
| screamingface-engine | 2877 passed, 93.54% |
| aigateway | 4380 passed, 92.35% |
| aigateway-ui | typecheck clean, 240/240 |

`AIGW_TEST_PG=1 pytest -m needs_postgres` — 32 passed, 5 failed. The five need a `pg_dump` binary absent from this machine and fail identically on the bare branch.

`run_gates.py` green on all three Python stacks, append-only check included.

Docs and spec updated to match: `DEPLOYMENT.md`'s "no long lock" claim is gone, the PRD records the retry demotion, and the ERD gains the trigger, the merge lock and invariants E10/E11 — each marked as added or corrected in review round 2 so a later reader can tell a revision from a recovered fact.

HupBaHa (changes requested):
Follow-up review on exact head `44be27d7`. The earlier fixes close several findings, but four correctness issues remain:

1. **High — a hit after a transport retry still reports an affirmative zero spend** (`apps/screamingface-engine/src/screamingface_engine/runner/connector.py:448-483,511-513`). `avoided_usd_for_outcome` correctly withdraws the saved-cost claim, but `_report_usage` still routes every hit through `_report_served_from_cache`, which publishes zero tokens and `cost_usd=0`. If attempt 1 reached the provider and was billed but its response was lost, and attempt 2 hits the row it wrote, the run now reports no saving but still reports zero spend. Carry the retry uncertainty into call and run cost accounting so the total is unknown/incomplete rather than a false zero. Gateway idempotency can later recover an exact amount, but it is not required to stop making the false claim.

2. **High — making `metadata_degraded` exact blocks cache-hit serving for the whole merge** (`apps/aigateway/src/aigateway/core/request_cache/bulk_loader.py:94-102,241-250`; `store.py:207-216`). `SHARE ROW EXCLUSIVE` conflicts with the `ROW EXCLUSIVE` lock taken by every awaited `hit_count`/`last_hit_at` update. The response row can be read under MVCC, but `get()` does not return it until that update finishes or fails, so hits stall for the duration of the merge. This contradicts the still-current snapshot contract at `docs/spec/2026-08-22-OME-951-admin-cache-snapshot-upload.md:129-130`: “The load never blocks serving.” Preserve serving and derive a non-blocking signal; if an exact count cannot be obtained without a global lock, an explicitly named lower bound is safer than changing the primary availability contract. If maintenance-window semantics are intentional, that needs an explicit approved contract change rather than being introduced by a telemetry fix.

3. **Medium — fixed precision still rounds valid totals once enough contributions accumulate** (`apps/screamingface-engine/src/screamingface_engine/runner/cache_counters.py:98-114`). Reusing `AMOUNT_PRECISION = 53` fixes the two-value example, but a total can require more significant digits than one contribution. Adding the valid boundary amount `999999999999999999.000000000000000000000000000000001` 1000 times under the shipped accumulator produces `999999999999999999000.00000000000000000000000000000010`; the exact result is `999999999999999999000.000000000000000000000000000001000`. Derive enough precision from the operands and exponent span, or enforce and test a total-domain bound; a fixed per-contribution precision cannot make an unbounded sum exact.

4. **Medium — `metadata_degraded` can still report zero when this load clears a block** (`bulk_loader.py:84-91`; `migrations/0011_cache_entry_metadata.py:110-115`; `admin_schemas.py:166-170`). The count includes only rows whose staged `metadata_json` is NULL. If a staged row has metadata byte-identical to the live block but a different `response_json`, the trigger sees unchanged metadata beside a changed response and clears it to NULL, while the pre-count remains zero. That contradicts the API description that this field counts live rows the load turned from priced to unknown. Include the trigger-cleared case in the signal or narrow the public contract so zero cannot be read as “this load degraded nothing.”

The stale-old-writer guard, bounded forward migration lock wait, saved-cost domain/provenance validation, and generated TypeScript field appear addressed on this head. All current checks are green; these remaining paths are not covered by them.

HupBaHa (approved):
The PR-specific blockers are resolved and the OME-1155 implementation is ready. Remaining pre-existing/adjacent concurrency gaps are tracked separately in OME-1220 and OME-1221.

PR #931, opened 2026-09-14
feat(engine): implement removable model-call activity plugin

keelancj:
## What this enables

Implements the removable Engine plugin that turns existing model-call observations into structured URL4 Logs. With the deployment wiring in the dependent draft, consumers can display:

- Model call started.
- Still waiting — elapsed time, refreshed every 60 seconds.
- Retrying — attempt and the actual selected delay.
- Completed, failed, refused or cancelled — elapsed time and safe outcome fields.

These are illustrative Client labels, not a new UI. Records have occurrence IDs, revisions and producer timestamps; the caller supplies the node-associated sink. A heartbeat means the Engine is still waiting, not that it sees provider-internal progress.

**This PR implements and tests the plugin; it does not register or enable it.** Deployment configuration and integration verification are in [stacked draft #935](https://github.com/ScreamingFace/screamingface/pull/935).

## Why this design

The generic observation ports and execution wiring already merged in #899 and #915. The plugin owns activity schema, validation, sessions, operation identity, rolling admission and heartbeat resources. Core execution remains independent of activity policy. This PR's runtime changes are entirely inside `activity/`; it adds no URL4 changes or deployment wiring.

One session shares a budget across a run's calls, preventing fan-out from multiplying the allowance. Each call still has its own occurrence. URL4 continues carrying generic structured Logs without knowing activity kinds or Client presentation.

## Safety and lifecycle

- Rolling admission: 100 records/second, burst 200, reserving 40 tokens from routine starts/heartbeats for retries and outcomes. Capacity refills throughout days-long runs; no lifetime logging cutoff.
- Safe scalar attributes, 4 KiB payloads and 256-character messages exclude prompts, responses, credentials and arbitrary exception text.
- Delivery is best-effort, including terminal records. Producer-suppression snapshots are cumulative; sink exceptions are contained without being counted as invalid records. Pending snapshots survive uncertain delivery. Bridge-loss attributes use a separate structured schema.
- Cleanup revokes activity, cancels all owned timers before yielding and joins them collectively despite repeated cleanup cancellation; caller cancellation propagates afterward.
- Disabled calls reuse stateless callbacks with no per-call activity operation allocation or tracking. Run-level binding still masks inherited activity.
- Callbacks perform no remote I/O. The merged inline-hook contract has no runtime latency enforcement, and record-rate admission does not cap heartbeat-task count. No server-side archive is introduced.

## Review boundary and follow-ups

Review size: **1,263 additions and 40 deletions across 12 files**, including tests and shared docs.

Review `contract.py` / `session.py`, then `scope.py` / `observer.py`, alongside their direct tests. The dependent draft owns registration, local/worker policy, Helm and run/stream/Client-decoding/removal tests. It must land after this PR.

Client Logs-tab work (OME-1135), benchmark-stage producers, exact case/member attribution (OME-699), aggregate privacy policy and provisional scores (OME-932) remain separate. Declaring case-loading/grading/aggregation kinds does not instrument those stages.

## Validation

- Full Engine gates pass on the standalone plugin branch: append-only checks, Ruff, Pyright, layering and pytest/coverage.
- Direct tests retain field/admission/sink-fault coverage, long-run simulations, fixed heartbeats, cancellation/revocation and disabled-mode regressions.
- Both Standards and Spec split reviews found no actionable issues.
- At the split, combined runtime matched `f418f53e` exactly and moved test bodies were unchanged. Subsequent deployment-policy corrections and their validation are tracked in #935; this plugin implementation is unchanged.
- Simulated durations and fake providers do not establish production-load or widget-UI behavior.

Shared design, plan, task mirror and ledger remain under the existing OME-1161 document paths.

Refs: OME-1161

## Review cleanup — 2026-09-16

Commit `831f3bf0` removes the unused on_heartbeat callback, names the unchanged burst/refill/reserve constants, and clamps bridge-loss counts to the nonnegative safe-integer range. The contract docstring now explains reserved vocabulary; the standard-library-only Emitter boundary is preserved.

Five appended loss-boundary cases reproduced two failures before the clamp. All 60 direct activity tests pass; full Engine gates pass, including append-only, with 93.26% coverage. Independent review found no material issues. No existing test was modified and no deployment behavior changed.

Ledger: `docs/work/2026-09-16-OME-1161-plugin-cleanup.md`.


khoaguin:
**Review: approve with nits.** The plugin is hexagonally clean (core has zero activity imports), admission/teardown invariants are traced and pinned by tests calling production paths. Nothing blocks merge. Three items worth fixing before or shortly after:

1. **Dead hook parameter** — `activity/scope.py:202`: `operation(on_heartbeat=...)` is accepted, stored, and invoked at `scope.py:130`, but no caller in this PR, #935, or any test ever passes it — it is permanently `None`, an exercised-nowhere path. Drop it, or land it with its first caller.
2. **Producer-less wire vocabulary** — `activity/contract.py:22-29`: `ActivityKind.CASE_LOADING/GRADING_PREPARE/GRADING_CHECK/GRADING_REDUCE/AGGREGATION`, the `_COUNTS` facts, and identifiers `benchmark_id`/`case_id`/`provider` have no producer in this PR or #935; the only production producer is `MODEL_CALL` + `model_id`. If the intent is freezing the `screamingface.activity.v1` wire schema before stage producers land, say so in the contract module docstring; otherwise trim to what `MODEL_CALL` emits. Same question for `current_operation()` (`scope.py:215`, test-only consumer) and the never-raised `provider_timeout`/`provider_rate_limit`/`provider_error` failure codes.
3. **Coupled magic numbers** — `activity/session.py:37,60-63`: `200.0` (burst), `100.0` (refill/s), and `41` are bare literals, and `41` secretly encodes "burst − reserve of 40 + the token being spent" — only decodable via the PR body. Name them (`BURST`, `REFILL_PER_S`, `RESERVED_FOR_OUTCOMES`) so a future tuner can't change one without the other.

Mechanical: `activity/observer.py:41` clamps `dropped` only from above — add `max(0, ...)` so a negative internal counter can never emit a negative `loss.dropped_total`. `contract.py:31-35` `Emitter` is a byte-identical duplicate of `observations.LogEmitter`; since `observer.py` already bridges to core, the copy buys no independence and the two protocols can drift.

Also flagging cross-stack: the stacked #935 has a precedence hole (ambient `URL4_CLOUD_ACTIVITY_LEVEL=full` can defeat injected `off` through a caller-built `Settings()`) — doesn't block this PR, but worth fixing before the stack is marked ready. Details on #935.

PR #932, opened 2026-09-14
feat(aigateway): expose execution access in model discovery

keelancj:
## Summary

Model discovery can return a valid datasheet even when the caller has no provider access. Add `context.execution_access: configured | missing` to the existing Gateway model-parameters response so evaluation preflight can distinguish those cases without a new endpoint.

The field uses the already resolved account/profile target and the provider's existing profileless/no-auth declarations, including Gemini environment-key access. It reports configuration, not credential validity, and preserves credential-free datasheets, typed selection errors, and private/no-store responses. No inference or credential validation is added.

This draft implements only the Gateway portion. Engine already forwards the response bytes and permits unknown fields, so no Engine change is required. OME-1042 will consume the field in Client before Candidate dispatch. Planned Engine unit OME-1196 was canceled. Older responses without the field must be treated as unknown. The approach narrows the separate-preflight proposal in #928 to an additive discovery field.

## Validation

- New behavioral tests cover missing access, API-key/OAuth profiles, active connections and ambiguous selection, Gemini environment-key presence/absence, no-auth execution, caller isolation, secret non-disclosure, and existing typed errors.
- 28 focused tests passed and exited cleanly.
- Full Gateway gate runner passed: append-only checks, lint, format, typecheck, no-enterprise guard, and full pytest/coverage.
- An earlier complete run had 4,249 passing tests and 92.61% coverage, with one existing login-timing assertion failure. The unchanged full rerun passed; no tests or thresholds were weakened.

Refs: OME-1195
Related: OME-1194, OME-1042

## Review cleanup — 2026-09-16

Commit `87d98763` corrects the contract-identity comments to explicitly exclude route-added execution_access and freshness, adds the configured bool annotation, and adds two identity-stability regression cases. No runtime semantics changed.

The test keeps auth mode, account and target fixed while actual Gemini environment-key availability changes missing → configured → missing. It asserts unchanged contract_id, context.revision and parameter semantics for both GEMINI_API_KEY and GOOGLE_API_KEY. An uncontrolled Gemini environment flip would also change auth mode, which legitimately changes identity.

All 13 access tests and full Gateway gates pass, including append-only; coverage 92.61%. Independent review found no material issues. No existing tests modified. Ledger: `docs/work/2026-09-16-OME-1195-identity-review.md`.


khoaguin:
**Review: approve with nits.** The 13-line production delta is correctly wired, adds zero provider round-trips, reports configuration presence only (never credential validity), and the tests pin the security invariants (no injection, no inference, no secret echo, caller isolation). Two things to fix:

1. **Now-false invariant comment** — `core/model_parameter_contract.py:205-208`: the OME-648 comment says "Every other field in this block is a digest input", but `routes/model_parameters.py:200` injects `execution_access` into `context` *after* the digest is built. A future dev trusting the comment either pins access state by `context.revision` (stale `missing` served under an unchanged id when the env key appears) or "fixes" it by folding access into `digest_inputs`, breaking the deliberate outside-identity decision. Update the comment to name `execution_access` as the second stated exclusion (like `freshness`), or route the field through the composer as an explicitly-excluded parameter.
2. **Missing identity-stability test** — the load-bearing claim "changing access does not itself change parameter-contract identity" has no test; nothing goes red if someone adds `execution_access` to `digest_inputs`. In the existing Gemini env-flip test, assert `contract_id` and `context.revision` are byte-identical across the missing→configured flip.

Nit: `routes/model_parameters.py:195` — `configured` lacks a local annotation in a file that annotates locals.

Design calls verified and agreed: the two-value `configured|missing` enum with client-side "absent = unknown" is the right YAGNI shape (the gateway never fabricates an "unknown" it can't observe); the gate mirrors the chat route's own chatless-permission check, so discovery `configured` tracks what dispatch would actually accept; `("none",)` comparison copies the existing `chat_credentials.py` idiom byte-for-byte.

HupBaHa (approved):
Reviewed exact head `f5fc2544`. The execution-access signal reuses the dispatch credential target, preserves account/profile isolation and typed errors, handles current profileless and no-auth providers consistently, and keeps the response private and non-cacheable. No blocking or material findings.

PR #933, opened 2026-09-14
fix(screamingface): check provider access before evaluation

keelancj:
## Summary

A parameter-free `evaluate()` can currently dispatch with no provider access and return an unscored report. Both Clients now inspect Gateway-reported `context.execution_access` for every required Candidate Model before observers or Candidate dispatch. An authoritative `missing` raises `ProviderConnectionError` with provider/model context and a connection hint.

The Client fetches one details document per required Model, reusing any model-admission response. Configured hosted/BYOK/profileless access proceeds normally. Missing metadata from older Gateways remains unknown, preserving existing behavior; malformed metadata and auth/transport failures keep their discovery errors. Ordinary model-details lookup remains available even when access is missing. Doctor is unchanged.

## Compatibility and delivery

This consumes Gateway #932 through the existing Engine passthrough; no Engine change is needed. This PR can merge before #932 because the new field is optional. The early-error behavior requires the new Gateway field at runtime. Merge Gateway first before cutting the Client release that bundles the local runtime.

`ModelDetails.execution_access` is additive and defaults to `None`; README, changelog and public API snapshot document it. Request-count tests now require every Candidate Model to be checked. Existing parameter-validation and dispatch assertions remain intact.

## Validation

- 41 new access regression cases passed; the earlier focused preflight/admission run passed all 48 cases before adding the local/hosted matrix.
- Full screamingface gate runner passed: lint, format, typecheck, all tests with 95.16% coverage (95% floor), notebook checks, build and distribution verification.
- The append-only guard was explicitly waived for the approved request-count changes and additive public API snapshot; all other gates ran unchanged.
- No inference, credential validation, or live-provider calls added.

Refs: OME-1042
Related: OME-1194, OME-1195

## Semantic rebase — 2026-09-16

Rebased onto main `0c0abfcf`, preserving answer-seed validation alongside all-model access preflight and admission-document reuse. Both preflight functions now accept keyword-only prefetched alongside answer_seed. Sync reuse checks None explicitly. Existing seed, parameter and access checks remain before observer creation and dispatch. Shared test-helper extraction remains separate.

Validation: 83 existing combined tests passed, then 91 with eight additional sync/async, listed/admitted, missing/configured cases. The new cases assert the expected pre-spend access or seed error, one details fetch, no dispatch and no events. Full Client gates passed, including append-only against the rebased head, 95.23% coverage, notebook checks, build and distribution verification. Independent integration review found no material regressions.

Head `90ac30c8`; ledger `docs/work/2026-09-16-OME-1042-semantic-rebase.md`. Refreshed CI runs on the pushed branch; PR remains open.

## Replay CI follow-up — 2026-09-17

Commit `1f07ce3f` fixes keyless cache replay being rejected by live-provider access preflight. A test-only Gateway adapter omits only optional `context.execution_access` from successful model-parameter discovery, exercising supported legacy/unknown compatibility. It never fabricates configured credentials. Production preflight, all other discovery fields/errors, chat/cache handling and credential-free cache-miss failures remain unchanged.

Validation: actual-stack discovery regression failed before the fix; five adapter cases pass. Docker replay passed 135 tests, including trace correlation and unchanged cache-miss assertions. Separate board runs match the draco-3pass, ifeval and healthbench-worst30 goldens; GDPval assets are unavailable locally and remain covered by CI. Client lint, format, types, full tests/95.24% coverage, notebook checks, build and distribution checks pass. The authorized replay-launcher and README edits require the documented narrow append-only exception; existing test assertions, goldens and snapshots are unchanged.

The unrelated Gateway `test_validation_session_shares_one_absolute_deadline` timing failure repeated on Python 3.13 after rerun; this commit does not alter that Gateway test or production code.


khoaguin:
**Review: request changes — semantic rebase needed before anything else.** The implementation itself is sound and well-tested (error UX, wire-boundary validation, version skew all verified clean), but the branch forked before OME-1193/OME-1038 (PRs #913/#927) rewrote the exact module this PR rewrites, and GitHub now reports the PR CONFLICTING.

**The blocker** — `_evaluation/model_parameters.py`: main's version is now `preflight_sync(candidates, load, *, answer_seed)` with `_seed_models`/`_preflight_models`/`_validate`, and `runner.py` calls it with `selected_candidates` + `answer_seed=answer_seed`. A naive textual conflict resolution has two silent-failure shapes: keep this branch's positional-`prefetched` signature and drop `answer_seed` → a seeded evaluation on an Anthropic route loses its pre-spend seed gate and fails mid-run after spend; or keep main's `_preflight_models` and lose this branch's "seed the map with every `candidate.models`" line → parameter-free models silently skip the access check and the ticket's guarantee evaporates while all diff-visible code still looks correct. Please rebase by re-weaving: `prefetched` as a keyword arg alongside `answer_seed`, the all-models seeding folded into `_preflight_models`, `_validate_access` called from `_validate` — then run the 41-case access matrix and the OME-1038 seed tests together.

Minor, while rebasing:
- `model_parameters.py:29` (sync path): `known.get(model) or load(model)` relies on `ModelDetails` always being truthy — true today, but the idiom silently re-fetches if the class ever gains container semantics; `(details := known.get(model)) is None` states the intent (the async path already does the explicit check).
- Sync path fetches details serially — a 10-model Fusion does 10 serial round-trips pre-dispatch where async gathers. Matches the pre-existing sync preflight pattern and it's all pre-spend, so acceptable; flagging because parameter-free `evaluate()` previously made zero detail calls.
- `tests/e2e/test_fake_gateway_discovery.py:5` imports helpers from sibling `test_failures` — a rename there now breaks discovery tests.

Security dimension verified clean: `ProviderConnectionError` carries only gateway-reported provider/model ids, never key material; `execution_access` is validated to the closed literal set at the wire boundary with malformed values raising the typed discovery error; absent field → `None` → old behavior, pinned by the omitted-field test; the client talks only to Engine passthrough routes.

PR #935, opened 2026-09-15
feat(engine): enable model-call activity in local and deployed runs

keelancj:
## What this enables

Connects the model-call activity plugin from #931 to local and deployed Engine runs. Deployments can select structured progress in the existing URL4 Logs stream: call start, fixed 60-second waiting updates, retries with actual delay, and terminal outcomes.

**Rebased onto main after #931 merged.** This review contains deployment/composition and integration verification, preserving the merged plugin cleanup.

The deployment commits were rebased separately from the merged plugin implementation; review this PR against main.

## Why this is separate

The parent defines the plugin's behavior and direct contracts. This PR makes it selectable at the entry points and proves it survives the real execution/stream boundaries. Reviewers can assess deployment policy and integration without reviewing the plugin internals again.

The registration module is the composition point. Generic observation interfaces, connector and executor remain unchanged; the generic executor still defaults to no observers. URL4 remains a generic structured-Log carrier.

## Deployment policy

- `off` / `full`, default `off`. Select locally through `URL4_CLOUD_ACTIVITY_LEVEL=full` or configured Settings; in Helm use `config.activityLevel: "full"`.
- A configured activity field on caller-supplied Settings wins. Otherwise an injected environment mapping takes precedence over the process environment, including an empty mapping defaulting activity to off. Only when env is omitted is process policy used. Automatically created Settings receive the selected activity policy, so ambient full (or an invalid ambient value) cannot override injected off.
- Worker policy is deployment-owned; queued run inputs cannot escalate it.
- Unsupported policies are rejected. Aggregate mode remains deferred.
- Off mode preserves existing operator diagnostics and masks inherited activity; the parent's stateless callbacks avoid per-call activity bookkeeping.
- Activity remains ephemeral and best-effort through the existing bridge. No additional archive or remote I/O inside callbacks is added.

## Integration evidence

Tests cover policy precedence in both directions, worker enforcement, nested/cross-task run scope and revocation, real model success/retry/failure/refusal/cancellation streams, WebSocket delivery and independent Client decoding. Structured bridge-loss totals remain separate from producer suppression.

The removal test omits the activity package and empties registration in a copied installation, then verifies execution without editing generic core code. Requests, retries, results/accounting, cancellation and operator diagnostics retain their behavior.

## Validation and limits

- Full Engine gates pass on the combined branch: append-only checks, Ruff, Pyright, layering and pytest/coverage.
- The original split preserved pre-split runtime and test bodies. Follow-up `769d3a29` corrects injected-versus-ambient activity precedence, with six added cases. All 19 deployment-policy tests pass; direct cleanup/off regressions remain unchanged in the parent.
- Standards and Spec split reviews found no actionable issues.
- Tests use fake providers and simulated durations. Inline hooks have no enforced latency bound; the record budget does not cap heartbeat-task count. This does not establish production-load or widget-UI behavior.

This pair completes the model-call producer. The Client Logs tab (OME-1135), benchmark-stage instrumentation, semantic attribution, aggregate privacy policy and provisional scores remain separate.

Refs: OME-1161

## Review follow-up — 2026-09-17

`ec3d6a1f` fixes the remaining ambient-versus-injected policy issue for caller-supplied Settings. Constructor and model-copy overrides retain precedence; automatically environment-loaded fields do not override injected policy. Effective app settings now match observer policy without mutating caller settings, and registration reuses the shared environment-key constant.

Eleven appended regression cases cover ambient full versus injected off/empty, reverse precedence, invalid injected policy, explicit values equal to ambient, copied settings and process fallback. Six regressions failed before the fix. All 46 focused deployment/integration tests pass with ambient full; all Engine gates, Helm lint and 112 chart-wiring checks pass. Standards and Spec follow-up reviews found no blockers. Worker boot-time policy validation remains a separate optional improvement.


khoaguin:
**Review: request changes** — the injected-vs-ambient precedence fix is only half-landed.

**1. Ambient env can still override injected `off`** — `local.py:150-160` (`_local_activity_configuration`). The helper decides "did the caller set `activity_level`?" via pydantic's `model_fields_set`, but a plain `Settings()` constructed under ambient `URL4_CLOUD_ACTIVITY_LEVEL=full` absorbs that value from the env and marks the field as set. Reproduced: ambient `full` + `create_local_app(settings=Settings(), env={"URL4_CLOUD_ACTIVITY_LEVEL": "off"})` → level `full`. That is exactly the escalation the follow-up commit and PR body claim is closed — it is closed only for auto-created Settings, and passing `Settings()` is the natural way to set any *other* field (the PR's own tests do it). Fix direction: derive precedence from whether the *caller* set the field (compare against a `Settings` built without env sources, or thread the level separately) rather than from `model_fields_set`.

**2. The same mechanism makes a shipped test environment-sensitive** — `tests/unit/test_activity_deployment.py:298-302`: `test_local_invalid_activity_policy_is_rejected_before_startup` builds `Settings()` without clearing the ambient var; on any machine with `URL4_CLOUD_ACTIVITY_LEVEL=full` exported, `Settings()` absorbs `full` as "explicit", the injected `"limited"` is never consulted, no `ValueError` fires, and the test fails. Add the `monkeypatch.delenv` guard the neighboring test (~line 343) already has — and add a test asserting injected `off` beats ambient `full` through a caller-built `Settings()`, which pins finding 1.

Secondary (judgment calls):
- `local.py:154-159`: when supplied Settings carries an explicit level, an *invalid* injected value is silently discarded instead of rejected — the misconfiguration a deployer meant to express is never diagnosed.
- `local.py:160`: supplied `Settings()` (field unset) + injected `full` → observers run `full` while `app`'s settings object still reads `off`; nothing reads the field downstream today, but the first future reader inherits a lie. Consider `model_copy(update=...)` on the supplied object too.
- `worker/supervisor.py:743`: an invalid worker-env value is forwarded verbatim to every child and dies per-run instead of at worker boot; a one-time validation at supervisor startup fails faster (fine to defer since Helm validates).

Mechanical: `observation_plugins.py:12` hardcodes `"URL4_CLOUD_ACTIVITY_LEVEL"` while `job_env.ACTIVITY_LEVEL` is the declared constant for the key — use the constant. A one-line docstring on `_local_activity_configuration` stating "supplied-configured > injected env > process env > off" would have made finding 1 visible at review time.

Everything else checked out: composition confined to the registry module, generic executor stays observer-free by default, off-mode cost effectively zero per call, queue/expression escalation and secret non-leakage each pinned by a dedicated test.

PR #955, opened 2026-09-16
feat(engine): register the imported gsm8k proof board with its check surface (5/6)

khoaguin:
The first real imported benchmark goes live: **inspect-gsm8k**, graded by inspect's own `match(numeric=True)` scorer, with its scorer doubling as a check surface — which is what lets `corrective_loop` run on an imported board with zero adapter changes.

> **Stack 5/6** for OME-1115 (spec §4 + §7). Base: #954. Also carries #954's deferred test suite and the two owner-approved roster-test amendments the first registered plugin board forces (the third — the url4 import-inventory allowlist — rode #954).

## Problem / Solution

**Problem.** Everything so far is machinery with no passenger: no imported board actually registers, so the discovery loop, the shim, the snapshots, and the assembly line have never been proven end-to-end on a real eval.

**Solution.** Register gsm8k — grade-school math, free-text numeric answers — as the proof board. Its declaration is ~90 lines: the pinned dataset, their scorer, the prepare/install hooks, and `with_check_surface=True` (a numeric-answer board can safely expose its grader as a practice surface; the sealed feedback never leaks the key).

## Before / After

```mermaid
flowchart TB
  subgraph TODAY["TODAY — the import lane is an empty conveyor"]
    direction LR
    T1["discovery (#951) · shim (#952) ·\nsnapshots (#953) · assembly (#954)"] -.->|"no board registered;\nplugin still contributes ()"| T2["/v1/benchmarks:\n7 home-grown boards only"]
  end
  subgraph AFTER["AFTER this PR — first imported board on the catalogue"]
    direction LR
    A1["gsm8k.py declaration\n(pins · match(numeric=True) ·\ncheck surface ON)"] --> A2["single_shot_board"]
    A2 -->|"plugin entry point"| A3["/v1/benchmarks:\n7 + inspect-gsm8k"]
    A4["candidate probes the\ncheck surface"] -->|"sealed pass/fail\nfeedback — never the key"| A2
  end
  TODAY ~~~ AFTER
  style T2 fill:#7f1d1d,color:#ffffff
  style A3 fill:#14532d,color:#ffffff
```

### Before
- Trigger: an operator with the `inspect` extra installed lists the catalogue.
- Today: seven home-grown boards; the plugin contributes nothing.
- Cost: the whole OME-1115 stack is unproven plumbing.

### After
- Same trigger.
- Now: `inspect-gsm8k` appears with its own 16-hex revision; its grading route and check route both run their real scorer through the shim.
- Win: the corrective_loop recipe works on an imported board for free — the check surface IS the grading route, exactly the spec §4 promise.

### Don't regress
- Extra-less installs still see exactly seven boards, byte-identical (the #951 pin keeps holding).
- Check-surface feedback stays sealed: the target never enters the response — pinned by test.
- Per-Case failure names (`missing_case_row`, `missing_target_asset`) surface in the aggregate rather than aborting it — pinned by test.

## Roster-test amendments (owner-approved — the riskiest diff here, review first among tests)

The first plugin-registered board breaks roster photographs that previously assumed "every registered board is home-grown." The two in THIS diff were amended to state the real invariant rather than weakened (the url4 import-inventory amendment landed with #954):

- `test_benchmark_deployment.py` — the family guard now iterates `STATIC_REGISTRATIONS` (the home-grown seven) instead of the whole deployment: plugin boards legitimately live in another package family.
- `test_benchmark_declaration.py` — the declaration table gains the imported boards' rows, with a subset rule so the pin holds at every stack stage (gsm8k registered here, mmlu only after #956).

## Review order

1. `apps/screamingface-engine/src/screamingface_engine_inspect/gsm8k.py` — the passenger: the whole board as one declaration. Verify every value traces to `pins.py`, the scorer factory is lazy, and `with_check_surface=True` is a deliberate choice (free-text board — the elimination attack that bans MCQ check surfaces doesn't apply).
2. `apps/screamingface-engine/src/screamingface_engine_inspect/single_shot.py` + `boards.py` — small wiring deltas: the per-board installer conformance (`ASSET_BUNDLE_ID` + `install_gsm8k` beside it) and `board_registrations()` now returning the board.
3. `apps/screamingface-engine/tests/unit/test_inspect_gsm8k_board.py` — the board's contract: definition pins, the check route's sealed feedback, aggregate accuracy, and both missing-material failure names.
4. `apps/screamingface-engine/tests/unit/test_inspect_single_shot.py` — #954's deferred factory suite (identity shape, revision determinism + pin sensitivity, registration completeness) with dummy hooks — proves the factory independent of any real eval.
5. The two amended roster tests (listed above) — verify each amendment narrows to the true invariant and nothing else in each file moved.

Refs: OME-1115


khoaguin:
Closing unmerged by owner decision: the per-board-module design (one Python file per benchmark) doesn't scale to the ten-board import. gsm8k lands instead as a **BoardSpec row** in the board-as-row design on OME-1116, together with mmlu and the ten new boards. This branch stays as the donor for the board metadata and the full test suite (check-surface sealed feedback, refuse-early, factory identity tests), which carry over. Infra from this stack (#951–#954) is merged and unaffected.

PR #956, opened 2026-09-16
feat(engine): register the imported mmlu proof board and close the ticket docs (6/6)

khoaguin:
The second proof board lands: **inspect-mmlu**, an MCQ board graded by inspect's real `choice()` scorer — and deliberately **without** a check surface, proving the assembly line handles both sides of the OME-796 elimination-attack rule. Closes out the OME-1115 stack.

> **Stack 6/6** for OME-1115. Base: #955. After this merges, the whole spec §3+§4 execution is on main; the paid `limit=50` acceptance run on inspect-gsm8k is the owner's hand-over step.

## Problem / Solution

**Problem.** One proof board (gsm8k, #955) shows the happy path, but it exercises only half the design: a free-text board *with* a check surface. The owner decision on OME-1115 requires a second, structurally different board — MCQ, where a check surface is forbidden (a candidate could eliminate wrong letters by probing) — before the lane counts as proven.

**Solution.** Register mmlu: 4-choice questions across 57 subjects, their `choice()` scorer through the same shim, `with_check_surface=False`. Same factory, opposite flags — zero new assembly code, which is the point.

## Before / After

```mermaid
flowchart TB
  subgraph TODAY["TODAY — one proof board, half the design exercised"]
    direction LR
    T1["inspect-gsm8k\n(free text · match scorer ·\ncheck surface ON)"] --> T2["catalogue"]
    T3["MCQ lane:\nchoice() scorer · surface\nrefusal · choices in targets"] -.->|"unproven"| T2
  end
  subgraph AFTER["AFTER this PR — both sides of the check-surface rule live"]
    direction LR
    A1["inspect-gsm8k\ncheck surface ON\n(free text — safe)"] --> A3["catalogue:\n7 + 2 imported boards"]
    A2["inspect-mmlu\ncheck surface OFF\n(MCQ — elimination attack,\nOME-796)"] --> A3
    A4["candidate probes\nmmlu check route"] -->|"refused by name"| A2
  end
  TODAY ~~~ AFTER
  style T3 fill:#7f1d1d,color:#ffffff
  style A3 fill:#14532d,color:#ffffff
```

### Before
- Trigger: a reviewer asks whether the import lane generalizes past its first board.
- Today: gsm8k alone can't answer — MCQ grading (choices riding the private targets, the shim's answer-marking replay) and the check-surface *refusal* path have never run against a registered board.
- Cost: the "zero adapter code per board" claim is untested where it's hardest.

### After
- Same trigger.
- Now: two boards from one factory, opposite check-surface flags, distinct revisions; mmlu's aggregate runs their real `choice()` scorer (an unparseable reply scores 0, never crashes).
- Win: the next imported eval is a declaration + tests, nothing else. Ticket docs (spec §7 board table, work ledger outcome) close in this same PR per house rule.

### Don't regress
- gsm8k's registration, revision, and check surface are untouched — the registration order and both revisions are pinned by test.
- The MCQ surface refusal is loud and named, not a silent 404 — pinned by test.
- Extra-less installs remain byte-identical to pre-stack main (the #951 pin holds through the full stack).

## Review order

1. `apps/screamingface-engine/src/screamingface_engine_inspect/mmlu.py` — the second passenger: same declaration shape as gsm8k with the opposite surface flag and the shuffle seed among its revision pins (the baked order is exam identity, #953). Verify no assembly logic leaked in — it must stay a pure declaration.
2. `apps/screamingface-engine/src/screamingface_engine_inspect/boards.py` — `board_registrations()` now returns both boards, gsm8k first; verify order (registration order is catalogue order).
3. `apps/screamingface-engine/tests/unit/test_inspect_mmlu_board.py` — the genuinely new behavior under test: the surface refusal, real-`choice()` aggregate, the unparseable-reply-scores-0 path, and both-boards/distinct-revisions pins.
4. `docs/spec/2026-09-09-OME-1113-inspect-evals-import.md` — §7 amended from one proof board to the two the owner decided; verify it records the decision, not new design.
5. `docs/work/2026-09-15-OME-1115-inspect-adapter-plugin.md` — ledger outcome (deviations include the two-session collision resolution); mechanical skim.

Refs: OME-1115


khoaguin:
Closing unmerged by owner decision — same reason as #955: mmlu lands as a BoardSpec row on OME-1116 (board-as-row design). The OME-1115 spec §7 amendment and ledger/mirror close that rode this PR move to OME-1116's first PR. Branch kept as test/metadata donor.

PR #964, opened 2026-09-16
feat(py-screamingface): export a report in inspect's .eval log format

khoaguin:
A completed report can now export a one-way copy of itself in inspect's `.eval` log format, so researchers in the inspect ecosystem open our runs in the tooling they already use.

## Problem / Solution

**Problem:** Our run results are written in a language only our own tools read. Researchers who live in the inspect ecosystem cannot open them in the viewers and scripts they already use — exactly the audience the imported benchmarks are meant to attract.

**Solution:** Teach the report to export a one-way copy in their log format, clearly stamped as ours, so their tooling opens our results with nothing new to install.

## Before / After

```mermaid
flowchart TB
  subgraph TODAY["TODAY — results locked in our formats"]
    direction LR
    R1["completed report"] -->|"export()"| F1["report.json only"]
    F1 --> X1["inspect view / their
analysis scripts:
cannot open"]
  end
  subgraph AFTER["AFTER this PR — one-way export into their ecosystem"]
    direction LR
    R2["completed report"] -->|"report.export(format='inspect')"| L2[".eval log — provenance-labeled,
real metered cost"]
    L2 -->|"one-way: never re-imported,
never feeds the leaderboard"| V2["inspect view opens it"]
  end
  TODAY ~~~ AFTER
  style X1 fill:#7f1d1d,color:#ffffff
  style L2 fill:#14532d,color:#ffffff
```

### Before

- Trigger: a researcher who lives in the inspect ecosystem wants to open one of our runs in `inspect view` or their analysis scripts.
- Today: the report exports only our own JSON document; results are locked out of their ecosystem's viewers.
- Cost: friction for exactly the researchers the imported boards attract.

### After

- Same trigger.
- Now: `report.export(format="inspect")` writes a `.eval` file that inspect's own reader opens — stamped `produced_by: ScreamingFace engine`, carrying the run's real metered cost as `total_cost`, one log per candidate (their own unit: one task × one model).
- Win: our runs open in their viewer with zero new tooling on their side; a one-line `pip install "screamingface[inspect]"` on ours.

### Don't regress

- One-way only: no reader API, no ingest surface, no `expression_sha` anywhere in the log — nothing in the leaderboard path can consume it.
- `export()` with no arguments (and every existing JSON call) behaves byte-identically; the existing export tests are untouched and green.
- The metered cost is passed through, never recomputed — an unpriced run exports with **no** cost rather than a fabricated zero.
- Works for home-grown-board reports, not only imported boards (all tests run on plain draco-shaped fixtures).

## How it's built

- **The mapping is pure, the dependency is quarantined.** `screamingface._inspect_log.payload` turns one candidate's run into a plain-JSON document with zero inspect imports — fully covered by the extra-less CI install. `screamingface._inspect_log.write` is a mail slot: it checks the `.eval` address and the extra is installed, then hands the document to inspect's own `EvalLog.model_validate` + `write_eval_log`, so the zip layout stays *their* contract.
- **Pinned to the engine's dialect.** New `inspect` extra pins `inspect-ai==0.3.263` — the exact version the engine's inspect plugin pins, because the log contract is version-shaped.
- **Genuinely new dependency fact:** `inspect-ai==0.3.263` and the runtime extra's `litellm==1.98.0` cannot co-install on *any* platform (their `boto3` windows never intersect — verified empirically). Declared as a `[tool.uv] conflicts` pair rather than loosening either pin; the hosted-notebook persona is unaffected, a local-runtime env exports from a separate env.
- **Surface change (owner-approved):** `Report.export` grows `format=` / `candidate=` keywords; the public-surface snapshot regen is exactly the two additive signature rows.

Verified live: `uv run --extra inspect pytest tests/test_inspect_log_live.py` round-trips an export through inspect's real reader (the same one `inspect view` uses) — status, samples, provenance label, and cost all intact.

Spec: `docs/spec/2026-09-16-OME-1117-inspect-log-export.md` · Plan: `docs/plan/2026-09-16-OME-1117-inspect-log-export.md` · Ledger: `docs/work/2026-09-16-OME-1117-inspect-log-export.md`

Refs: OME-1117

## Review order

1. **`docs/spec/2026-09-16-OME-1117-inspect-log-export.md`** — the contract in one page: what we take from inspect (their writer) and what we refuse (any read path, hand-rolled zips). Verify the Out-of-scope list matches what you'd expect a one-way export to refuse.
2. **`packages/screamingface/src/screamingface/_inspect_log/payload.py`** — the heart: the translator that turns one candidate's run into inspect's document, with no inspect installed. Verify the provenance metadata block and that the cost is a pass-through, never arithmetic.
3. **`packages/screamingface/src/screamingface/_inspect_log/write.py`** — the mail slot around their writer. Verify the guarded import is the only place `inspect_ai` appears in the SDK and the missing-dep error names the fix.
4. **`packages/screamingface/src/screamingface/report.py`** — the export seam only: format dispatch plus the refusals (`candidate=` with JSON, unknown formats). Verify the JSON branch is the old body unchanged.
5. **`packages/screamingface/tests/test_inspect_log_payload.py`** — the mapping's spec in test form (17 tests). Riskiest judgment call to check: a graded refusal exports with the refusal text as the scored answer (the OME-1037 output/refusal exclusivity).
6. **`packages/screamingface/tests/test_inspect_log_write.py` + `test_inspect_log_live.py`** — the stubbed mail-slot proof and the real round-trip (opt-in, skipped in CI). Verify the live test asserts the acceptance items: reader opens it, provenance label, metered cost.
7. **`packages/screamingface/tests/test_report.py`** — four appended seam tests; the pre-existing tests are pure untouched insertions above them.
8. **Mechanical skims:** `pyproject.toml` (the extra + the `[tool.uv] conflicts` pair with its WHY comment), `uv.lock` (regenerated), `public_surface_snapshot.json` (two signature rows), and the four SDLC docs.


keelancj (commented):
@khoaguin 

[P2] Avoid double-counting ensemble usage
_stats_payload puts the candidate’s aggregate usage and its members’ usage into the same stats.model_usage map. Inspect sums these entries, so the existing fusion fixture displays $0.22 for a $0.12 run ($0.12 + $0.05 + $0.05). Token totals are similarly inflated whenever member usage is present.

PR #965, opened 2026-09-16
feat(engine): read an inspect task's exam facts for the pin generator

khoaguin:
Milestone B begins: the pin generator's **reading half** — one command-to-be that turns an inspect eval into row diffs, this PR landing the part that reads the facts (introspection + capture), with the part that writes them following in the stack.

> **Stack 1/2 for Milestone B** (base: main; owner decisions 2026-09-16: rows are written in place so `git diff` is the review artifact · facts are auto-introspected from the task name with dev review as the gate · an uncleared license warns but emits). Split by the ≤500-line rule: this PR reads, the next writes.

## Problem / Solution

**Problem.** After Milestone A, importing a benchmark is two hand-written rows — but every value in them (dataset path, split, revision sha, row count, license, scorer, template) is still copied by a human from the eval's source and the Hub, which is exactly the transcription work the ten-board ticket exists to mechanize.

**Solution.** The importer reads instead of retypes: it calls the eval's own task function against a recording stub (so nothing downloads), harvesting everything the task declares, then captures the three facts no eval file carries — revision sha, row count, license — as named observations. Verified live: introspecting the real gsm8k and mmlu tasks reproduces our hand-written rows exactly.

## Before / After

```mermaid
flowchart TB
  subgraph TODAY["TODAY — row values are transcribed by hand"]
    direction LR
    T1["eval source +\nHF dataset page"] -->|"a human copies\npath/split/sha/count\nBY HAND"| T2["pins + SnapshotSpec +\nBoardSpec rows"]
  end
  subgraph AFTER["AFTER this PR — the facts are read, never retyped"]
    direction LR
    A1["inspect task ref\n(e.g. gsm8k:gsm8k)"] --> A2["introspect_task:\nhf_dataset recorder +\nregistry metadata\n(NO network)"]
    A2 --> A3["TaskFacts: dataset · split ·\nrecord_to_sample · template ·\nscorer(+kwargs) · mcq"]
    A4["capture_observations:\nHub sha · row count ·\nlicense (network, build-side)"] --> A5["Observations"]
  end
  TODAY ~~~ AFTER
  style T1 fill:#7f1d1d,color:#ffffff
  style A2 fill:#14532d,color:#ffffff
```

### Before
- Trigger: a dev imports inspect eval #3.
- Today: they open the eval's source and the dataset's Hub page and transcribe ~8 values into the three row files — the copy-typo class of bugs, invisible until a bake fails (or doesn't).
- Cost: import stays a careful hand process; ten boards means ~80 hand-copied values.

### After
- Same trigger.
- Now: `introspect_task("inspect_evals.gsm8k.gsm8k:gsm8k", {"fewshot": 0})` returns every declared fact by running the eval's own task against a recorder (its `hf_dataset` binding is swapped for the duration — zero downloads); `capture_observations` records the sha/count/license the way a lockfile records resolved versions (an upstream revision pin, when the eval has one, wins over Hub HEAD).
- Win: the write stage (next PR) turns these two values into the reviewed diff; humans review, never transcribe.

### Don't regress
- No runtime surface changes at all: the module is new and imported by nothing — catalogues, boards, and the extra-less install are byte-identical to main.
- The importer refuses rather than guesses: a task that never loads from the Hub, an ambiguous multi-load task, an unresolvable scorer or template each raise a named `ImporterError` — pinned by test.
- Prompt settings the bake would silently drop are *flagged*, never guessed at: a custom solver (like mmlu's own multiple-choice wrapper), a `system_message` instruction, and a custom choice template each earn a review flag in the facts (review round 2026-09-17) — pinned by test.
- An upstream revision pin is resolved through the Hub to the commit sha it names, so a mutable ref like `main` can never land in pins.py, and the license is read at that same revision (review round 2026-09-17) — pinned by test.

## Architecture / Design

This PR is the importer's reading half — two functions, two value shapes. `introspect_task` runs the eval's own task function against a swapped-in `hf_dataset` recorder (restored in a `finally`), so every declared fact is read from the recorded load kwargs + registry metadata with zero network and zero re-typing; anything it cannot resolve raises a named `ImporterError`, and anything the bake would silently drop (custom solver, `system_message`, choice template) earns a review flag instead of a guess. The result is **TaskFacts** — what the eval DECLARES. `capture_observations` then talks to the Hub (build-side only) for the facts no eval file carries: the revision resolved to a commit sha (an upstream pin wins; a mutable ref like `main` can never land), the row count at that sha, and the license read at that same revision — **Observations**, what was OBSERVED. The write stage (#966) consumes both to emit the three row diffs; nothing imports this module yet, so the runtime surface is byte-identical to main.

## Review order

1. `apps/screamingface-engine/src/screamingface_engine_inspect/importer.py` — the whole idea, top to bottom: the customs-officer docstring (declared facts vs captured observations), `TaskFacts`/`Observations` (the two value shapes), `introspect_task` (the `hf_dataset` recorder trick — verify the original binding is restored in a `finally`), the exam-vs-fewshot dataset disambiguation, scorer/template resolution to dotted references, and `capture_observations` (upstream pin wins; both network loaders injectable). Genuinely new, but build-side only — nothing imports it yet.
2. `apps/screamingface-engine/tests/unit/test_inspect_importer.py` — the contract via a fabricated eval module (no network): free-text and MCQ fact-reading, the fewshot two-load case, the refusals, and capture's pin-precedence. Verify the fabricated module exercises the same shapes the real evals showed live (gsm8k / mmlu introspection reproduced our merged rows exactly).
3. `docs/work/2026-09-16-OME-1116-import-ten-inspect-boards.md` — the Milestone B plan updated to this two-PR split; mechanical skim.

Refs: OME-1116


keelancj:
@khoaguin  found in review:

- **Lost prompt settings**: system instructions and custom MCQ templates disappear without a review flag, changing the imported exam. Preserve them or flag unsupported settings. importer.py:205-209

- **Mutable revision pins**: an upstream revision like main is stored unchanged, so later builds can fetch different data. Resolve it to a commit SHA and read the license at that revision. importer.py:258-260

khoaguin:
@keelancj both findings were real — fixed and pinned, thanks.

**1. Lost prompt settings** — `system_message` was on the "understood" solver list, so its instruction would silently vanish from the baked exam, and a custom `multiple_choice` template was dropped the same way. The list is now `_FULLY_BAKED_SOLVERS` (only the renders prepare.py actually reproduces); a `system_message` is **always** flagged (`…(system instructions are not baked)`) and `multiple_choice` is flagged whenever it carries a non-default template — both land as `TODO(review)` lines in the generated SnapshotSpec entry (#966). Pinned by `test_introspect_flags_a_system_message_the_bake_would_drop` and `test_introspect_flags_a_custom_choice_template`; the default-template MCQ case is asserted flag-free.

**2. Mutable revision pins** — `pinned_revision or info.sha` stored a branch/tag ref verbatim and read the license at HEAD. `capture_observations` now passes the upstream pin through `HfApi.dataset_info(dataset, revision=pin)` and stores the **resolved commit sha** the Hub returns, so the license is read at that same revision and `main` can never land in pins.py. Pinned by `test_capture_resolves_a_mutable_upstream_pin_to_its_commit_sha` (asserts the pin names *which* revision to resolve, the sha is what's stored).

Both branches force-pushed (#966 rebased on top); all gates green.

PR #966, opened 2026-09-16
feat(engine): emit generated board rows in place — the importer's write stage

khoaguin:
Milestone B completes: the pin generator's **writing half** — the read facts become the three row fragments and land in place at anchor comments, so `git diff` is the import's review artifact.

> **Stack 2/2 for Milestone B** (base: the reading half). After this merges, `uv run python -m screamingface_engine_inspect.importer <task_ref> --key <k>` is the whole import command, and Milestone C's ten boards arrive as generated, human-reviewed diffs.

## Problem / Solution

**Problem.** The previous PR reads every fact an import needs, but they still end as Python objects in memory — a dev would print and paste them, and hand-pasting is the same transcription risk the importer exists to remove.

**Solution.** The write stage: render the pin constants, the `SnapshotSpec` entry, and the `BoardSpec` entry in the target files' own style, and insert each immediately above an anchor comment in its file. The diff a reviewer reads before merge IS the trust window pins.py's lockfile docstring promises; catalogue prose is emitted as TODOs for the importing **agent** to write — onboarding is AI-first: the AI finds every fact and writes everything, and the human's whole job is reviewing/verifying the diff.

## Before / After

```mermaid
flowchart TB
  subgraph TODAY["TODAY — facts read, still pasted by hand"]
    direction LR
    T1["TaskFacts +\nObservations"] -->|"a human pastes\nBY HAND"| T2["pins.py / prepare.py /\nboards.py"]
  end
  subgraph AFTER["AFTER this PR — one command, one reviewable diff"]
    direction LR
    A1["python -m ….importer\ntask_ref --key k"] --> A2["render_fragments +\ngenerate_rows:\ninsert at anchors,\nrefuse existing keys,\nwarn uncleared license"]
    A2 -->|"in-place edit"| A3["git diff: pin rows ·\nSnapshotSpec entry ·\nBoardSpec entry (prose TODOs)"]
    A3 -->|"HUMAN reviews the diff —\nthe only trust window"| A4["PR merges;\nboard on the catalogue"]
  end
  TODAY ~~~ AFTER
  style T2 fill:#7f1d1d,color:#ffffff
  style A2 fill:#14532d,color:#ffffff
  style A3 fill:#78350f,color:#ffffff
```

### Before
- Trigger: a dev imports inspect eval #3.
- Today: the reading half hands them the facts; getting them into the three files is still manual editing in three places (including keeping prepare.py's pins import block in sync).
- Cost: paste drift between the three files; the import block silently missing a constant.

### After
- Same trigger.
- Now (AI-first onboarding): the agent runs one command that edits all three files at their anchor comments — pin rows (with the HF permalink and the captured license in the comment), the `SnapshotSpec` entry referencing those constants (the import block updated, kept sorted), the `BoardSpec` entry with `with_check_surface=True` only for free-text boards (MCQ never gets it — OME-796) and TODO placeholders for title/description/focus, which the agent then fills from the eval's own docs.
- Win: importing benchmark #13 = the agent runs one command and writes the prose; the human reviews one diff.

### Don't regress
- Catalogues and runtime behavior unchanged: the anchor comments in pins.py / prepare.py / boards.py are comments — extra-less and extra-installed installs behave byte-identically to the base PR.
- An already-imported key is refused (never doubled) — and so is a key whose derived constant stem collides with an existing board's (`foo-bar` vs `foo_bar` both derive `FOO_BAR_*`) or isn't a valid Python identifier (`2wikimultihop` would emit unloadable pins) — all pinned by test (review round 2026-09-17).
- Writes are all-or-nothing: every anchor and the import block are validated while building the new file texts, before any file is written — a broken anchor refuses the whole import and a retry after restoring it succeeds cleanly, never a half-imported tree — pinned by test (review round 2026-09-17).
- The bake and board assembly now mechanically refuse a mutable `dataset_revision` (anything not a 40-hex commit sha), the backstop behind the importer's sha resolution — pinned by test.
- An uncleared dataset license WARNS but emits (owner decision 2026-09-16) and is written into the generated pins comment, so the diff reviewer cannot miss it — pinned by test.
- Every generated file still parses and the generated pins execute to the captured values — pinned by round-trip tests against copies of the real files.

## Architecture / Design

One pass through the importer, left to right: the eval's own task module (from the pinned inspect_evals package) plus one operator command go into **Stage 1 — introspect** (no network: the task's `hf_dataset` binding is swapped for a recorder, the task function runs, and every declared fact — dataset/config/split, `record_to_sample`, scorer, MCQ-ness — is read, never re-typed). **Stage 2 — capture** adds the facts no eval file carries, from the Hub at build time only: resolved revision sha (an upstream pin wins), row count at that sha, dataset license. **Stage 3 — emit** renders three code fragments and inserts each above its anchor comment in place: the pins.py pin row (the imported exams' lockfile), the prepare.py `SnapshotSpec` row (the frozen-snapshot bake), and the boards.py `BoardSpec` row (scorer, check surface iff not MCQ, catalogue prose left as `TODO(review)`). The resulting `git diff` IS the review artifact — the agent fills the TODO(review) prose, a human verifies pins/license/prompt-render, and import time is the only trust window. After merge, the build bakes the snapshot and the board factory stamps routes + revision from the pins; the board appears on `/v1/benchmarks` as `inspect-<key>`.

## Review order

1. `apps/screamingface-engine/src/screamingface_engine_inspect/importer.py` (the added stage-3 half) — `render_fragments` (the three fragments in the files' own idiom; MCQ vs free-text check-surface rule; custom solvers surfaced as TODO(review)), `generate_rows` (refuse-existing-key → license warn → four insertions), the anchor/import-block insertion helpers, and `main` (the CLI; both network loaders injectable). This is the PR's only genuinely new behavior — and it runs at import time on a dev machine, never in the engine's runtime.
2. `apps/screamingface-engine/src/screamingface_engine_inspect/pins.py` / `prepare.py` / `boards.py` — one anchor comment each (the insertion contract). Verify they are pure comments placed inside the structures the fragments extend.
3. `apps/screamingface-engine/tests/unit/test_inspect_importer.py` (the added sections) — renderer goldens, insertion round-trips on copies of the real three files (files still parse; generated pins execute), duplicate-key refusal, license-warn path, and the CLI wiring test that proves `--task-arg fewshot=0` reaches the task function.

Refs: OME-1116


keelancj:
@khoaguin some review findings

- **Pin collisions**: importing foo-bar, then foo_bar, generates identical constants. The second import silently overwrites the first board’s dataset, revision, and case count. Reject collisions before writing. importer.py:335

- **Partial writes**: a missing marker in boards.py causes failure after pins and snapshots are already written. Restoring the marker and retrying then fails with “key already exists.” Validate every insertion point before writing. importer.py:443-446

- **Invalid Python**: key 2wikimultihop generates 2WIKIMULTIHOP_DATASET, which is invalid Python. The importer reports success, but the generated files cannot load. Generate valid identifiers and validate syntax before writing. importer.py:335

khoaguin:
@keelancj all three confirmed and fixed in `8770471a`, thanks — good catches, all the same failure family (the write stage trusted its inputs/filesystem more than it should).

**1. Pin collisions** — the dedup only checked the literal key strings, so `foo-bar` then `foo_bar` passed and both derived `FOO_BAR_*`. `_refuse_existing_rows` now also refuses when the derived constant stem already exists in pins.py, naming the stem. Pinned by `test_generate_rows_refuses_a_colliding_pin_prefix`.

**2. Partial writes** — the four insertions each did read→insert→write, so a missing boards.py anchor failed *after* pins/prepare were already modified, and the retry hit "already exists". `generate_rows` is now compute-then-write: all three new file texts are built (every anchor and the import block validated) before ANY file is written — a broken anchor refuses the whole import and a retry after restoring it succeeds cleanly. Pinned by `test_generate_rows_writes_nothing_when_an_anchor_is_missing` (asserts pins/prepare byte-unchanged after the refusal AND the restored retry works).

**3. Invalid identifiers** — `2wikimultihop` → `2WIKIMULTIHOP_DATASET =` is a SyntaxError the tool reported as success. `_pin_prefix` now validates the derived stem with `.isidentifier()` and refuses before anything is rendered or written, with a suggestion in the message. Pinned by `test_generate_rows_refuses_a_key_that_is_not_an_identifier_stem`.

All gates green; branch pushed.

PR #968, opened 2026-09-16
feat(OME-1183): serve the node's config catalog over .well-known, enforce x-scope, migrate to url4.json

IonesioJunior:
Implements **[OME-1183](https://linear.app/openmined/issue/OME-1183)** — the node's layered configuration catalog, served as scoped JSON Schema over `.well-known`, plus the `url4.toml` → `url4.json` migration the spec locks.

## Summary

A url4 node now **tells a caller what it can be configured with, and then holds them to it.**

- **Three `.well-known` documents**, served per caller: the card, the compound bundle, and each mount's schema alone. `$id` is the only thing the node changes; the body is byte-for-byte what the endpoint served.
- **`x-scope` enforced** at rest (a node file) and in flight (a request). The ladder is a *floor* — a `node` item is the deployer's, a `user` item is the caller's.
- **D1 decided: both carriers are valid.** A `config` member on the WS attach frame, or `URL4-Config-*` headers. Both parse into one payload, so the transports cannot disagree.
- **A catalog linter and its CI gate**, because `x-scope` is not a JSON Schema keyword and [2020-12 §6.5](https://json-schema.org/draft/2020-12/json-schema-core#section-6.5) *requires* a validator to ignore it. Every validator calls a catalog with a scopeless leaf valid.
- **`url4.toml` is deleted.** No dual-format period, per the spec.

Design decisions that weren't in the parent spec are recorded in `docs/spec/2026-09-16-OME-1183-config-carriers-and-enforcement.md`.

## Commits

| | |
|---|---|
| `feat(url4): add the config discovery package` | 8 stdlib-only modules, nothing imports them yet |
| `feat(url4)!: read serve config from url4.json, and serve discovery` | the format layer + the wiring |
| `feat(engine)!: move the declared world to url4.json` | **atomic** — see below |
| `ci: gate every config catalog on the x-scope linter` | its own workflow, like `charts.yml` |
| `docs(OME-1183): record the carrier, enforcement and linter decisions` | spec, ledger, task mirror |
| `fix(url4): narrow Applied.problem before subscripting it in tests` | pyright, whole-package |
| `docs(OME-1183): record the append-only Confidence-Gate decision` | see below |

## Breaking changes

Two commits are marked `!`. `url4 serve` and the Runner both stop reading TOML, and **an argv is now an array** — TOML's bare-string form went through `shlex.split`, which is exactly the ambiguity an array removes. The error hands the operator the array to paste.

## Why the engine commit is one commit

Seven sites name the config file: the Dockerfile, the wheel build hook, `check_distribution.py`, the wheel resource lookup, the **checkout-detection marker**, `preview_contract.py`, and the e2e harness. Split them and the build breaks far from its cause — and **two fail silently**: miss the build hook and the wheel ships with no runner config; miss `source.py` and checkout detection quietly reports installed mode. OME-1001 already guarded a related bug in that exact code.

Both were verified after the change: the built wheel carries `screamingface/_runtime/resources/url4.json`, and the local-mode fallback resolves and parses 117 routes.

## Test plan

| gate | result |
|---|---|
| `packages/url4` | **1411 passed** |
| `apps/screamingface-engine` | **2879 passed**, 8 skipped |
| `packages/screamingface` | **1467 passed**, 23 skipped |
| wheel distribution check | pass — ships `_runtime/resources/url4.json` |
| `verify_chart_wiring.py` | 112/112 |
| `preview_contract.py classify` | a `url4.json` change triggers the engine preview |
| `url4 serve` over real HTTP (the CI heredoc) | 8/8 assertions |
| catalog gate | red on a scopeless leaf; exit 2 when the linter is sabotaged |
| ruff · ruff format · pyright | clean, all four packages |

**Migration parity was checked by diffing collected test NAMES, not counts**: 66 → 73 for `url4 serve` (four renames, one deliberate split), and every engine module matched exactly. No test was dropped.

## Reviewer notes

**Valid user-scope config returns `501`, not `200`.** The values are the caller's to set and the node validates them — but forwarding config to a mounted endpoint doesn't exist yet (OME-1187 owns the composer). Answering `200` would claim they were applied. Owner-confirmed; when forwarding lands, that branch becomes the forward.

**The catalogs are not in this PR.** The spec's home for them is `docs/spec/2026-09-11-OME-1183-config-schemas/`. They exist locally and were the corpus this was built against; the gate's tests are self-contained and skip cleanly without them.

**This unit was not decomposed.** The parent spec names four sub-issues after approval and this spans three of them. Linear's landing group is single-select, so this branch can't carry one correct leaf. Flagged in `docs/tasks/2026-09-11-OME-1183-config-schema-scopes.md` as an owner decision.

**This push used one authorised `--no-verify`.** The `url4` stack's append-only check
refuses a branch that modifies prior tests (sdlc rule 5), and the migration rewrote five
existing test modules — they wrote `url4.toml` fixtures, and that file no longer exists.
Every *other* gate passes: ruff, ruff format, pyright over the whole package, 1411 tests,
and `aigateway` + `screamingface-engine` green in full. The owner approved the
modification; the decision and its evidence are recorded in the work ledger.

**The parent spec is still unpushed** (`docs/spec/2026-09-11-OME-1183-config-schema-scopes.md`, branch `OME-1183-config-schema-scopes`). Everything here was built from the Linear issue body plus the worked-set README — a contradiction in that file would land on this work.

Refs: OME-1183


keelancj (commented):
@IonesioJunior some review findings:

- A bare URL4-Config- header causes an uncaught error instead of HTTP 400 `_serve.py:739`

- Config groups like auth, queue, and worker are accepted but silently ignored; only world is read `world_config.py:77`

- Discovery calls the node’s /claude/.well-known/ instead of the endpoint’s own /.well-known/, marking healthy endpoints unavailable `resolver.py:173`

- Caller identity isn’t forwarded when fetching schemas, so allowed values can’t be tailored to the caller ` resolver.py:146`

- Startup doesn’t validate the whole config, so misplaced fields like "port":9999 are ignored and defaults used ` _serve.py:265`

- Request validation skips constraints like maximum, required and additionalProperties, treating invalid values as valid `scope.py:114`

PR #970, opened 2026-09-17
feat(engine): land eight generated boards and the choice-template render

khoaguin:
Eight importer-generated boards land in the catalogue — it now holds **ten imported benchmarks** — plus the one renderer the new boards forced (an eval's own multiple-choice template, baked verbatim).

## Problem / Solution

**Problem:** the catalogue holds two imported benchmarks, and every further one has been a hand-crafted effort — days of transcribing values and writing wiring code per board.

**Solution:** run the importer built in the previous two PRs for real: eight boards arrive as generated data rows, an agent fills the catalogue prose, and the only human work left is reading this diff.

## Before / After

ome1116-ten-boards drawio

### Before
- Trigger: a researcher wants breadth — ARC, MMLU-Pro, BoolQ and friends next to gsm8k and mmlu.
- Today: each import is a bespoke project; the catalogue has exactly the two proof boards.
- Cost: breadth arrives one multi-day effort at a time; hand-typed pins and guessed licenses are the error surface.

### After
- Same trigger.
- Now: one importer command per eval emits the board as reviewable rows; eight new boards (ARC-Easy, ARC-Challenge, CommonsenseQA, MMLU-Pro, WinoGrande, RACE-H, PAWS, BoolQ) land in this single diff, every dataset pinned to a commit sha with a verified case count and a recorded license.
- Win: board #11 is one importer run plus one diff review — hours, not days.

### Don't regress
- A board stays two data rows + pins, zero new modules, zero spine edits.
- MCQ boards are refused the mid-run check surface (elimination attack, OME-796); the free-text boards (PAWS, BoolQ) carry it.
- Gated or drifting upstream datasets cannot change a published board: build fetches by pinned sha and refuses a wrong row count.
- Extra-less installs are untouched — everything ships behind the `inspect` extra.

## Architecture / Design

The pipeline is the one in the diagram's AFTER lane; what is genuinely new in this PR:

- **The choice-template render.** Three of the new boards (MMLU-Pro, WinoGrande, RACE-H) pass their own template to inspect's `multiple_choice` solver. `SnapshotSpec` gains a `choice_template` dotted reference, the MCQ formatter accepts it, and the importer captures a template it can resolve to one module attribute as a fact instead of a review flag — so those exams bake through the eval's own words, never a lookalike.
- **Two importer refusals hardened by this batch.** A `record_to_sample` defined *inside* the task function (TruthfulQA's shape) can never be resolved by the row's dotted reference — the importer now refuses at import time instead of emitting a row that fails at image build. And every generated comment line now wraps under the 100-column lint gate regardless of how long the eval's reference is.
- **Everything else is data.** Pins are captured observations (sha, count, license); the two seeded shuffles (MMLU-Pro, RACE-H) exist because those splits arrive grouped by discipline/passage — verified empirically — and the seed rides the revision hash as exam identity.

## What's in the diff

| Board | Family | Cases | License |
|---|---|---|---|
| ARC-Easy / ARC-Challenge | MCQ | 2,376 / 1,172 | cc-by-sa-4.0 |
| CommonsenseQA | MCQ | 1,221 | MIT |
| MMLU-Pro | MCQ (own CoT template, seeded shuffle) | 12,032 | MIT |
| WinoGrande | MCQ (own template, 2 options) | 1,267 | none on card ⚠ |
| RACE-H | MCQ (own template, seeded shuffle) | 3,498 | "other" ⚠ |
| PAWS | free-text (includes), check surface ON | 8,000 | "other" ⚠ |
| BoolQ | free-text (pattern), check surface ON | 3,270 | cc-by-sa-3.0 ⚠ |

⚠ = not on the cleared-license list; the rows carry the license note and emitting them was an owner decision (2026-09-17) — this review is the gate. TruthfulQA was in the approved list but was **dropped**: its row rule is a task-local closure the row reference can never resolve (the new importer refusal is its regression pin). Every board above was verified by a **full offline bake** from its pinned revision — counts match the pins.

Owner-approved prior-test changes (append-only gate skipped with approval, 2026-09-17): the mmlu-board registration test now asserts the proof boards in catalogue order (the new suite owns the exact set), and the declaration roster grows its eight designed rows.

## Review order

1. **`apps/screamingface-engine/src/screamingface_engine_inspect/pins.py`** — the lockfile: eight generated pin blocks. This is the trust window — check each sha link, count, and license note reads sanely.
2. **`.../prepare.py`** — the snapshot rows (what each bake points at) plus the one real mechanism change: `choice_template` on the spec and in the MCQ formatter. Verify the render falls back to inspect's default when no template is set.
3. **`.../boards.py`** — the catalogue rows: prose a person will read on the leaderboard, and the check-surface flags (ON only for PAWS and BoolQ). Verify the prose against the datasets it describes.
4. **`.../importer.py`** — the two refusal/formatting hardenings; confirm nothing else about the emit path moved.
5. **`tests/unit/test_inspect_imported_boards.py`** — the new definition suite: the 10-key set, pin shape, family flags, reference resolution, no-TODO prose.
6. **`tests/unit/test_inspect_snapshots.py` + `test_inspect_importer.py`** — appended renderer and refusal tests (contract tests for the new mechanism).
7. **`tests/unit/test_inspect_mmlu_board.py` + `test_benchmark_declaration.py`** — the two owner-approved prior-test touches; confirm they only widen (order assertion, +8 roster rows).
8. **`docs/work/2026-09-16-OME-1116-import-ten-inspect-boards.md`** — the ledger's milestone-C plan + outcome, including the truthfulqa deviation.

Refs: OME-1116


khoaguin:
**Review round 2026-09-17 applied** (blocker + should-fixes 1–4) in 04b13ef3:

- **Blocker — dataset-kwarg conservation.** Every `hf_dataset` kwarg the importer reads is now reproduced, benign, or a **named refusal** (`limit`, `shuffle_choices`, `data_dir`, unknowns). `shuffle=True` demands a pinned seed (upstream seed reproduced; else `--shuffle-seed`). Probe fallout: **commonsense_qa, paws, boolq** shuffle upstream and had been imported without a pinned order — their rows now carry policy seeds (20260917) + updated prose, pinned by `test_boards_whose_eval_shuffles_carry_a_pinned_seed`.
- **1** positional `hf_dataset` args signature-bound in the recorder.
- **2** `data_dir` no longer aliased to config — refused (gsm8k's merged row was hand-verified; docstring example moved to arc_easy).
- **3** single-call fallback requires the Task's dataset to still hold the recorder's stub — a local exam beside HF fewshots refuses.
- **4** injection guards: charset refusal on Hub-controlled strings, 40-hex sha required at the tool, `ast.parse` before every write.

Deferred as flags-only follow-ups (per review triage): non-literal scorer kwargs, `multiple_choice` params (`multiple_correct`/`cot`), epochs/generation config, stage-2 network-error wrapping, the CI inspect-extra lane.

PR #972, opened 2026-09-17
test(engine): emitted importer rows must construct the real specs

khoaguin:
Round-trip tests for the inspect importer: the emitted `SnapshotSpec`/`BoardSpec` rows must construct the real dataclasses, so a spec-shape change fails in its own PR instead of at the next import session.

Fixes OME-1214

## Problem / Solution

**Problem.** The importer writes new benchmark entries by filling in a code template that must match the shape the engine expects. Nothing checked the two stay in sync: rename or add a required field on either spec and every test stays green, while the importer quietly starts producing entries that no longer fit — the break surfaces weeks later, as a confusing `TypeError` inside a generated file, far from the offending diff.

**Solution.** Make the tests *use* what the importer produces: run the real generation into a copy of the package, then execute the emitted rows with the real dataclasses in the namespace. The emitted kwargs must construct the real spec — the dataclasses ARE the schema, no parallel copy to drift.

## Before / After

*(No fresh diagram: this is a test-only change and OME-1214 already carries the identical sf-dark mechanism diagram.)*

### Before
- Trigger: a dev renames or adds a required field on `BoardSpec` or `SnapshotSpec`.
- Today: the importer suite checks emitted fragments with `ast.parse` (syntax) and execs only pins.py's constants — nothing ever constructs the real dataclasses from an emitted row. Every test stays green.
- Cost: the break surfaces at the next import session as a `TypeError` inside a generated file, with no pointer back to the spec change that caused it.

### After
- Same trigger.
- Now: two round-trip tests exec the emitted prepare.py and boards.py rows with the real `SnapshotSpec`/`BoardSpec` in the namespace; the snapshot test uses the maximal row (prompt template + choice template + shuffle seed) so every kwarg the template can emit is checked.
- Win: the spec-changer sees the failure in their own PR, with a test name and docstring that say why. Probe-verified: renaming `SnapshotSpec.case_count` or `BoardSpec.scorer` locally fails exactly the matching test.

### Don't regress
- The existing `ast.parse` checks stay untouched — they catch template syntax bugs the exec path would blur.
- No parallel schema (JSON/pydantic) was introduced; the real dataclasses are the only shape authority.
- The ticket's optional in-importer `dataclasses.fields` guard was **deliberately skipped**: the importer would have to import prepare/boards (boards pulls `url4.peer.server`), a heavy import edge for a check these tests already pin. Flagging for review — say the word and it becomes a follow-up.

## Review order

1. `apps/screamingface-engine/tests/unit/test_inspect_importer.py` (the only code change, +4 append-only tests: maximal + minimal construction per fragment at the "round-trip" banner) — the customs-office suite gains the step where the stamped paperwork is actually filed: emitted rows are executed against the real specs. Verify the tests construct the REAL imported dataclasses (not local copies) and that the snapshot test's fact set emits every optional kwarg the template can render.
2. `docs/work/…` + `docs/tasks/…` — the unit's ledger and mirror; confirm the probe results and the skipped-guard deviation are recorded.


keelancj:
Reviewed with the seven-lane playbook from #979 (`.claude/agents/sf-code-review.md`). Context pulled per §0: the `docs/tasks/` mirror, the work ledger, `importer.py`'s `render_fragments` / `_board_lines` / `generate_rows`, both real dataclasses, and the pre-existing test helpers.

**Verdict: no Action-required findings.** The round-2 fix already closed the one blocking-class gap (maximal-only rows couldn't see an *optional* field losing its default). Both `exec` paths fail loudly if the fragment comes back empty (`KeyError` / unpack error), and a renamed or newly-required field lands as a `TypeError` right here — which is the ticket's whole point.

---

```
[Lane 7 · Review recommended] The round-trip stops at the fragment string, not the file the importer wrote
  bug:   the test execs `fragments.snapshot` into a namespace seeded with ALL of pins.py,
         so every `SUMS_*` reference resolves regardless of whether generate_rows added
         it to prepare.py's import block. The written file is a different artifact.
  hurts: if `Fragments.import_names` ever drifts from the constants `snapshot_lines`
         emits (importer.py:472-497 — two independent lists, both conditional on
         shuffle_seed), the generated prepare.py raises NameError at the next import
         session while these tests stay green. That is the exact silent-inert shape this
         PR exists to kill, just moved one layer out. No live bug: today both lists
         cover the same six constants.
  fix:   cheap pin, no heavy import — assert every name in `fragments.import_names`
         appears in the written prepare.py import block, and that every `{PREFIX}_`
         token in `fragments.snapshot` is in `import_names`.
  pin:   extend the maximal test; it is the only one that passes shuffle_seed, i.e. the
         only one exercising the conditional arm of both lists.
  evidence: tests/unit/test_inspect_importer.py:38-40 (new) · importer.py:472-497 ·
            rule: guide Lane 1, "a guard scoped to one layer just moves the silent gap"
```

**Auto-fixable (follow-up, not a block)**

- `test_generated_snapshot_row_resolves_against_the_real_spec` (test_inspect_importer.py:399) says "must construct a real SnapshotSpec when the file executes" — it execs pins.py and asserts two constants; it never touches `SnapshotSpec`. The ledger cites it correctly as the *pins* coverage, but the name now sits next to `test_emitted_snapshot_row_constructs_the_real_snapshot_spec`, which does the thing it claims. Rename + fix the docstring (a rename, not an assertion edit — append-only is fine). Pre-existing, so per the guide's noise list this is a follow-up, not a block.

**On the deviation flagged for review** — the ticket's optional in-importer `dataclasses.fields` guard: agree with skipping it, and don't file a follow-up. It would force `importer.py` to import `boards` (pulling `url4.peer.server` into a CLI tool) to check the same property at a *later* moment than these tests already do. The finding above is the better use of that budget.

**Coverage gaps** — suite not run (and per the guide, green gates are not evidence anyway; the probe results in the ledger are). Linear issue not read directly, only the `docs/tasks/` mirror. Non-findings I checked and cleared: `engine_src_copy` is `tmp_path`-backed, so the four new tests can't collide on the `sums`/`quiz` keys; `BoardSpec.multiple_correct` is never emitted by the template but *is* covered — both board tests omit it, so dropping its default fails them; `scorer_kwargs` round-trips typed (`True`, not `"True"`).

PR #974, opened 2026-09-17
fix(engine): parent aigateway spans to the calling node, not the run root

sergio-bershadsky:
Closes OME-1185.

## The problem

`screamingface_engine/trace_scope.py` bound the run's `root_span_id`, so `runner/connector.py` sent `traceparent: 00-<trace_id>-<root_span_id>-01` on every outbound call and every aigateway server span attached to the run's **root**. The trace was connected but **flat** — verified in production 2026-09-17 04:25 UTC, trace `e2f0bfb83234b51d0228ce0fa39c0334`: 200 spans total (screamingface-engine 176 + aigateway 24), 1 root (`url4.run`), 0 orphans, and all 24 aigateway server spans sharing the single parent `c84b53aff1cdcb16`. Those 24 spans are the whole of that run's gateway traffic and not one of them can be attributed to the node that issued it.

(How many *logical* model calls the 24 represent is not derivable from the trace as it stands — `_complete` issues several gateway round trips per logical call when a tool loop runs. An earlier revision of this PR asserted "eleven provider calls" alongside the 24 without reconciling the two; that figure is unverified and has been removed everywhere.)

`OME-1119` chose the root deliberately: nothing emitted a root span yet, so a per-node parent would have dangled. `OME-1130` changed that premise — one `SpanEvent` per node plus a synthetic root, so both ids exist in the exported trace.

## The change

- `trace_scope.py`: a second ContextVar for the calling node's span, `bind_node_span()`, and `current_traceparent()` preferring it over the root. `run_trace_scope` clears it on entry so a nested or subsequent run cannot inherit a stale node id.
- `runner/executor.py`: `_Bridge.on_event` binds `NodeStarted.span_id` **before** any queueing policy. url4 calls `on_event` synchronously and inline from `Executor._eval`, in the node's own `asyncio.Task`, immediately before `node.resolve` is awaited — the exact context that node's outbound calls run in, with the same per-Task isolation `url4.observe._bind_node_sinks` relies on. `_RunState.map` sees the same event, but from the consumer's context, where a binding would reach no model call.
- **(review round 2)** `current_traceparent()` selected the node span by truthiness (`_node_span.get() or trace.root_span_id`), so an empty binding fell back to the run's ROOT — the flat trace this ticket removes, restored silently in the one case an operator most needs to be told about. Now `is None` for "no node is resolving", and a bound id that is not a span id (16 lowercase hex, not all-zero) renders **no header** rather than a confident wrong parent. `bind_node_span` deliberately does **not** raise — see D5 in the ledger.

**Why not add `current_span_id()` to `url4.observe`** (bound alongside the sinks): also correct, but it widens a published package API and makes this a two-package change (D9: epic + sub-issue per package) for a ticket labelled `screamingface-engine` alone. Recorded in the ledger.

## The ordering question (the ticket's explicit ask — answered in the ledger, D2)

A node's `SpanEvent` is published on FINISH, while the outbound call happens mid-node, so the gateway's parent id names a span exported later. **Accepted:**

1. OTLP is order-free; a parent arriving after its child is the normal case for any instrumented server call.
2. The only lossy case is a run killed mid-node — but `OME-1130` already refuses to export an unfinished span, so that node's span is missing either way. The choice is between a gateway span orphaned under a missing node (states the truth) and one misattributed to the root (states a falsehood that renders as a correct-looking trace, and is therefore not diagnosable).
3. The killed-run case is already visible from the run's own `Terminated(stopped|failed)` frame.

No change to when spans are exported.

## Tests

Appended to `tests/unit/test_traceparent_propagation.py` — no prior test edited or weakened, append-only gate clean without `--skip-append-only`.

**Round 2 (the HIGH review finding: every round-1 test passes with `_node_span` as a plain module-level global, so per-Task isolation — the entire reason a ContextVar is correct here — was untested):**

- `test_concurrent_siblings_each_render_their_own_node_span` — three Tasks bind, an `asyncio.Barrier` makes every bind precede every read, each asserts it renders ITS OWN span. Deterministic rather than a race: one shared slot necessarily hands all three the id written last.
- `test_a_fan_out_run_parents_each_gateway_call_to_the_node_that_made_it` — a real fan-out run through `Url4Executor` (`(a=/M1(…), b=/M2(…))!'$a $b'`, three distinct models so the outer reducing node has its own), asserting each gateway request's traceparent names the span the run exported for the node that issued THAT request. Correlated by model, carried on both halves (request body `model`, `SpanData.request_model`) — no `repr()`, no positional guessing.
- `test_a_node_span_that_is_not_a_span_id_sends_no_header_rather_than_the_root` (7 cases) + `test_a_usable_node_span_bound_after_an_unusable_one_still_renders`.
- An autouse fixture resets `_node_span` after every test in the module (review LOW: two tests bound in the runner's root context and never cleared).

**Mutation testing — round 1: 7 tried, 7 killed, all reachable from a ONE-node run.** The review is right that the figure supported a narrower claim than D1/D4 drew from it; the ledger now says so. Ignoring the node span (4 died); dropping the bind (2); not clearing on scope entry (1); binding after the overflow guard (1); binding on `NodeFinished` (2); binding `parent_span_id` (2); rendering a node span with no run bound (1).

**Round 2 — the concurrency mutation the review named:**

- **(8)** `_node_span` → a plain module-level global with a save/restore setter. Kills 3 tests against the final code: both new fan-out tests, plus `test_the_gateway_is_told_the_node_that_called_it_not_the_run_root` (incidentally — a single-node run's sibling literal nodes bind after the model node and clobber the shared slot before its headers are computed; that is scheduling luck, not a pin, which is why the two new tests exist).
- **(9)** `_SPAN_ID` widened to allow the all-zero id → 1 died.
- **(10)** the D5 screen reverted to `node_span or trace.root_span_id` → 8 died.

Not mutation-tested and stated as such: the autouse reset fixture is hygiene. No current test order fails without it, and there is no mutation that kills only it.

## Gates

`uv run .claude/scripts/run_gates.py screamingface-engine` — **ALL GATES GREEN** (append-only check, ruff check, ruff format, pyright, check_layering, pytest with coverage). Full engine suite: 3081 passed, 16 skipped.

Ledger: `docs/work/2026-09-17-OME-1185-calling-node-span-parent.md`


sergio-bershadsky:
Review round 2 — `6b5a69b4`. What I confirmed, what I fixed, and the one thing I refuted.

**HIGH — "all six tests pass with `_node_span` as a plain module-level global". CONFIRMED and fixed.**

Two tests now pin per-Task isolation:

- `test_concurrent_siblings_each_render_their_own_node_span` — three Tasks bind, an `asyncio.Barrier` makes every bind precede every read, each asserts it renders ITS OWN span. Deterministic, not a race: a single shared slot necessarily hands all three the id written last.
- `test_a_fan_out_run_parents_each_gateway_call_to_the_node_that_made_it` — a real fan-out run through `Url4Executor` (`(a=/M1(…), b=/M2(…))!'$a $b'`), asserting each gateway request's traceparent names the span the run exported for the node that issued THAT request. Three distinct models, because the outer reducing node is a third gateway call against the default model; correlation is by model, which both halves carry (request body `model` ↔ `SpanData.request_model`). No `repr()`.

Mutation (8), the substitution you named — `_node_span` → a module-level global with a save/restore setter — kills both, and also kills `test_the_gateway_is_told_the_node_that_called_it_not_the_run_root`. One correction to the finding: that third kill means the round-1 suite does *not* fully survive the substitution — in a single-node run the sibling literal nodes bind after the model node and clobber the shared slot before its headers are computed. That is scheduling luck rather than a pin, so your conclusion stands unchanged; only the "all six" count was off.

**MEDIUM — mutation table scoped narrower than the conclusion drawn from it. CONFIRMED and fixed.** The ledger now labels the round-1 seven as "all reachable from a ONE-node run" and lists round 2 separately: (8) above, (9) `_SPAN_ID` widened to allow the all-zero id → 1 died, (10) the D5 screen reverted to `or` → 8 died.

**LOW — two tests leave `_node_span` bound in the root context. CONFIRMED and fixed.** Autouse module fixture resets it after every test. Stated plainly in the ledger: this is hygiene, no current order fails without it, and no mutation kills only it — I am not claiming coverage I do not have.

**LOW — truthiness fallback + no validation. CONFIRMED on the first half, REFUTED on the second.**

Confirmed: `_node_span.get() or trace.root_span_id` meant an empty binding rendered the run's root — the flat trace silently restored. Fixed: `is None` for "no node is resolving", and a bound id that is not 16 lowercase hex / is all-zero now renders **no header** rather than a plausible wrong parent, which is this module's standing answer for an id it cannot trust.

Refuted: validating *inside* `bind_node_span` and raising. Evidence — its only caller is `_Bridge.on_event`, url4's synchronous observer callback, which is handed whatever the run's `Observer` minted. I implemented the raising version and ran the suite: it fails 16 existing tests that legitimately drive `_Bridge` with synthetic ids (`test_url4_executor.py` uses `f"span-{i}"`, `test_structured_log_pressure.py` likewise), and those tests are append-only. Worse, in production it would abort the node and, through the TaskGroup, the whole run — because a *tracing* id was malformed. Screening at render time gets the same guarantee with none of that. Recorded as D5 in the ledger with all three options.

**MEDIUM — "eleven provider calls" vs 24 gateway spans. CONFIRMED and corrected in all four places** (ledger, `docs/tasks/` mirror, PR body, `trace_scope.py` docstring). Only the verified figures remain: trace `e2f0bfb83234b51d0228ce0fa39c0334`, 2026-09-17 04:25 UTC — 200 spans, screamingface-engine 176 + aigateway 24, 1 root (`url4.run`), 0 orphans, all 24 gateway server spans under `c84b53aff1cdcb16`. Where the 11 used to be, the text now says the logical-call count is not derivable from the trace as it stands, and why.

Gates: `uv run .claude/scripts/run_gates.py screamingface-engine` — all green. Engine suite 3081 passed, 16 skipped. No prior test edited; append-only gate clean without `--skip-append-only`.

PR #975, opened 2026-09-17
feat(engine): surface the run's error detail and trace_id on the HTTP GET path

sergio-bershadsky:
Closes OME-941 (sub-issue of the Phase 0 traceability epic OME-935).

## The bug

`rest/routes.py::_terminal_response` read only `terminated.data.status` and mapped it through a fixed three-entry problem table. `TerminatedData.error{code,message,permanent}` was never read, so a synchronous `GET /` caller got a bare `502 "the run failed"` while the real diagnosis existed on the stream that is purged moments later. The run's `trace_id` was never surfaced either, which is the whole point of the epic — there was nothing to search a trace store with.

## The change

The problem body gains three optional RFC 9457 §3.2 extension members — `code`, `permanent`, `trace_id`. They are `None` by default and the handler already renders with `exclude_none=True`, so no existing member moves or changes type.

**Response-shape note, stated plainly:** `trace_id` is added to **every** terminal problem body — the 409 (stopped) and the 504 (timed_out) as well as the 502 — because a real run always carries a `traceparent` from `lifecycle.run`. Only a frame with no traceparent at all keeps the byte-identical pre-OME-941 body, and that is a test fixture, not the production case. `test_a_trace_id_rides_on_a_stopped_and_a_timed_out_problem_too` pins the production shape.

## Security — the crux, corrected in round 2

`ErrorInfo` is built by `url4.streaming.lifecycle._error_info` from `getattr(exc, "code")` and `str(exc)` of *whatever* exception ended the run. For any provider-facing adapter that message is the provider's own text — a response body, a refusal, an auth error quoting the key that failed.

Round 1 answered that with a closed allowlist of engine-authored codes, on the premise that **the code vouches for the author of the message**. Review refuted the premise at source: `runner/connector.py::_raise_for_status` lifted **both** `code` and `message` out of the upstream aigateway response body (`code = detail.get("code", code)`), so an upstream — or the fronting proxy this module already treats as an assumed threat — answering `{"detail": {"code": "malformed_source", "message": "<anything>"}}` would have had its own text vouched for and echoed verbatim. Not exploitable as it stood (no code emitted by `apps/aigateway` collides with the nine allowlisted names), and this is a path found by reading the code, **not an observed incident** — but nothing pinned it and no test would have noticed it changing.

The repair has three parts:

1. **One policy module, `src/screamingface_engine/error_text.py`.** A shared leaf under the layering gate, importable by both halves. It holds `ENGINE_ERROR_CODES` (moved out of `rest/routes.py`) and `public_identifier` / `public_message` (moved out of `benchmarks/aggregation.py`, which now imports them). Two public boundaries screening error text, one screen — the "one list, not two" argument applied a level up, where it was already being violated.
2. **An upstream can no longer mint a reserved code.** `_raise_for_status` discards an upstream `detail.code` that names a member of `ENGINE_ERROR_CODES` and keeps the engine-authored `aigateway_http_<status>`. Every other upstream code still passes through, so the benchmark failure report's vocabulary (`rate_limited`, `quota_exceeded`, …) and `ifeval/grade.py`'s `aigateway_http_[45]\d\d` match are untouched, and the existing `test_aigateway_http_errors_map_to_resolution_error` stays green unmodified. This is the reviewer's option (b), narrowed to exactly the reserved set — the wholesale version would have required rewriting that prior test, which the append-only rule forbids.
3. **The message is screened regardless of code.** An allowlisted message now goes through the shared `public_message`: capped at 200 chars, flattened to one line, withheld outright on an internal path, traceback marker or credential pattern. A vouched *author* is not a vouched *length* — `malformed_source` embeds `{token!r}` of the caller's expression with no bound of its own.

Unchanged and still true: `permanent` always survives (a bool cannot carry a leak, and it is the one field that says whether a retry can ever succeed); `parse_traceparent` validates rather than merely parses, so a malformed or all-zero traceparent yields no member instead of junk a caller would paste into a trace search; the topic is never rendered, and a test asserts it appears nowhere in a terminal problem response.

**Known and deliberate:** `_SCRUBBED_CODE` is the genuine `internal_error`, so a withheld body is indistinguishable from a real internal failure. A distinct sentinel would be more legible, but round 1 shipped assertions pinning `internal_error` and tests are append-only here — so the docstring's false claim to the contrary was corrected instead of the tests being rewritten. Follow-up ticket, not a silent change.

## Tests

`tests/unit/test_terminal_error_detail.py` (round 1: 9 tests; round 2 appends 26 more, counting parametrized cases) and two appended tests in `tests/unit/test_aigateway_connector.py`. **No prior test was modified, weakened or deleted** — the append-only gate confirms it.

Round 2's additions exist because the reviewer showed the round-1 suite was decorative in places:

- `test_the_allowlist_is_exactly_this_set` compares `ENGINE_ERROR_CODES` against a hand-written tuple. The first attempt parametrized over the set itself and **two mutations survived** — deleting an entry deleted a test case rather than failing one. The second copy is what makes add-a-code and remove-a-code both fail.
- `test_every_excluded_code_scrubs_its_message` is parametrized over the ten codes the ledger names as excluded, `resolution_failed` first: it is `ResolutionError`'s class default, i.e. the code under which connector text and `io/http.py`'s `f"GET {url!r} failed: {exc}"` actually arrive.
- `test_an_upstream_message_under_an_allowlisted_code_never_reaches_the_response` is the chain test, with the tainted step really executed: a crafted upstream response goes through the real `_raise_for_status` and url4's real `_error_info` before reaching the GET path. Nothing is hand-built, so it fails the moment any link starts trusting the upstream again.
- Four content-screen tests at the HTTP surface (cap, whitespace flattening, internal marker with no path so only the marker branch can withhold it, credential), plus the two connector tests.

Nothing in either file asserts over `repr()` of an object; the leak checks walk decoded JSON members and then the raw `resp.content` bytes.

## Mutation testing

**Round 2: 14 mutations applied, 14 killed** — including the ones the reviewer named and the ones the first attempt let live:

| mutation | outcome |
| --- | --- |
| `resolution_failed` added to the allowlist | killed |
| an entry deleted from the allowlist (×2) | killed — **survived** before the set was pinned against a literal |
| connector trusts `detail.code` again (full, and guard-only) | killed |
| routes echo the message unsanitized | killed |
| routes echo every code and message | killed |
| routes drop `permanent` from a scrubbed error | killed |
| routes report the topic as the trace id | killed |
| routes skip traceparent validation | killed |
| each of the four screens in `public_message` removed | killed — the credential screen **survived** until a test was added for it |

## Gates

`uv run .claude/scripts/run_gates.py screamingface-engine` — ALL GATES GREEN (append-only check, ruff check, ruff format, pyright, layering, pytest with coverage ≥80). No gate was skipped or weakened.

Ledger: `docs/work/2026-09-17-OME-941-terminal-error-detail.md` (round 2 appended at the end).


sergio-bershadsky:
## Review round 2 — what I fixed, what I refuted

Commit `c8aed80f`. Gates green: `uv run .claude/scripts/run_gates.py screamingface-engine` — append-only check, ruff check, ruff format, pyright, layering, pytest with coverage ≥80. No prior test was modified, weakened or deleted.

### HIGH — the allowlist did not vouch for the message. CONFIRMED, fixed.

Verified at source: `runner/connector.py::_raise_for_status` lifted **both** `code` and `message` out of the upstream response body. The allowlist therefore proved nothing about who wrote the message.

Fixed in three parts rather than either option alone:

1. **New shared leaf `src/screamingface_engine/error_text.py`** — one policy for error text crossing a public boundary. It now owns `ENGINE_ERROR_CODES` (from `rest/routes.py`) *and* `public_identifier` / `public_message` (from `benchmarks/aggregation.py`, which imports them). Your "one list, not two" argument, applied one level up where it was already being violated.
2. **Option (b), narrowed:** `_raise_for_status` discards an upstream `detail.code` that names a reserved engine code and keeps `aigateway_http_<status>`. Wholesale "stop trusting `detail.code`" would have broken the prior `test_aigateway_http_errors_map_to_resolution_error`, which pins `caught.value.code == detail["code"]` for `rate_limited` / `quota_exceeded` / `invalid_credential` / `upstream_unavailable` — and that vocabulary is what the benchmark failure report and `ifeval/grade.py`'s `aigateway_http_[45]\d\d` read. Tests are append-only here, so the narrowed form removes the whole hazard without touching a contract anything depends on. That test is still green, unmodified.
3. **Option (a) on top:** an allowlisted message now goes through the shared `public_message` — cap, whitespace flatten, internal-path / traceback / credential screen. A vouched author is not a vouched length; `malformed_source` embeds `{token!r}` of the caller's expression unbounded (that was your other low finding, fixed by the same change).

The test you asked for is `test_an_upstream_message_under_an_allowlisted_code_never_reaches_the_response`. It executes the tainted step for real: a crafted 401 body claiming `code: malformed_source` goes through the actual `_raise_for_status` and url4's actual `_error_info` before touching the GET path. Nothing is hand-built, so it fails the moment any link starts trusting the upstream again.

### HIGH — the frozenset was decorative. CONFIRMED, fixed, and you were more right than you said.

Parametrized over the excluded codes (`resolution_failed` first) and the allowlisted ones. My **first attempt still failed**: parametrizing over `ENGINE_ERROR_CODES` itself means deleting an entry deletes a test case instead of failing one, and the "allowlist loses `unrenderable`" / "loses `result_too_large`" mutations both survived. Fixed by comparing the set against a hand-written tuple in `test_the_allowlist_is_exactly_this_set`, so adding *and* removing a code now fails.

Round-2 mutation testing: **14 applied, 14 killed**, including `resolution_failed` added to the allowlist, the two deletions above, and the connector trusting `detail.code` again in both its full and guard-only forms. A fourth mutation (removing the credential-pattern screen in `public_message`) survived until I added a test for it at this boundary.

### MEDIUM — `trace_id` on every terminal problem. CONFIRMED, stated plainly rather than scoped.

You are right that the round-1 tests hid it by passing `traceparent=None`. I kept the behaviour and stopped hiding it: the PR body now says outright that **every** terminal problem body changes in production — 409, 504 and 502 alike — and that only a frame with no traceparent keeps the old bytes, which is a fixture and not the production case. `test_a_trace_id_rides_on_a_stopped_and_a_timed_out_problem_too` pins the production shape. Scoping it to failed runs would make a stopped run the one terminal outcome you cannot find in the trace store, which is the opposite of the epic's point. The change is additive (an optional RFC 9457 §3.2 member); no existing member moves or changes type.

### LOW — `_SCRUBBED_CODE` collides with the real `internal_error`. CONFIRMED as a fact, NOT fixed in code.

Your reading is correct: the withheld body is byte-identical to a real internal failure. But round 1 already shipped two assertions pinning `body["code"] == "internal_error"`, and changing the sentinel means rewriting prior tests — not allowed here. The false part was the docstring's claim that "the caller still learns … that the detail was withheld", so that is what I fixed: the docstring now says the two cases are indistinguishable and why that is a deliberate trade. A distinct sentinel (`withheld`) is a follow-up ticket, not a silent test rewrite.

### LOW — unbounded allowlisted message. CONFIRMED, fixed by the same `public_message` change above.

### On your note about `routes.py:586`

Agreed it is out of scope and worth its own ticket — `_logger.info("run stop requested topic=%s", sub)` logging a bearer capability is a separate, pre-existing problem and I have not touched it here.

sergio-bershadsky:
## Round 3 — the allowlist could not vouch for the message

The re-review was right, and auditing **all nine** codes rather than the two reported found **four** leaks, for two distinct reasons.

### 1. A code is not an author

`core/errors.py` gives two classes the same class default:

```python
class ParseError(Url4Error):       code = "malformed_source"   # :51  caller's own expression
class CollectionError(Url4Error):  code = "malformed_source"   # :86  embeds the FETCHED BODY
```

`CollectionError`'s raise sites: `io/layer.py:206` `{body[:80]!r}`, `:243` `{line[:80]!r}`, `:178` the upstream Content-Type. One code, two authors — **no allowlist entry can be correct for it.** And it needs no hostile upstream: it is the ordinary failure of `*` over a non-collection.

### 2. Messages interpolate resolved values

- `unknown_processor` — `dag/processor.py:96` embeds `{resolved!r}` = `(await spawn(value))`, a **model output**.
- `unrenderable` — `core/render.py:112` embeds `{text!r}`, the rendered AST.
- `expansion_not_iterable` — `dag/nodes.py:570` embeds `{exc}` of the CollectionError above.

**Allowlist narrowed 9 → 5.** The survivors carry their evidence beside the frozenset: `unbound_reference`, `cycle_detected`, `unknown_identity`, `timeout`, `result_too_large`.

### A test of mine that proved nothing

The first version of the new leak tests used `PROVIDER_SECRET` (`sk-proj-…`) and **passed before the fix** — `public_message`'s credential screen caught the token, so the test demonstrated the screen and said nothing about the allowlist. Real leaked remote text is prose, not a credential. Swapped to an ordinary-prose canary; the four tests then failed for the right reason before the narrowing. Same shape as the `repr()` assertion `OME-1132` shipped — an assertion that cannot fail.

### Verification

- Gates **ALL GREEN** (`--skip-append-only`, see below).
- **5 mutations, 5 killed** — re-adding each removed code, and dropping `unbound_reference`. Each died by **three** independent tests.
- Assertions **56 → 59**. No test deleted or weakened.

### Rule 5 disclosure

The append-only gate fired and this was **approved by the owner before commit, not after**. `test_a_failed_run_names_its_engine_error_code_message_and_trace_id` used `malformed_source` as its fixture — a code the audit proved unsafe — so the fixture moved to `unbound_reference`. The two changed assertion lines compare the **same fields**; only the expected literals follow the fixture. Four entries were also removed from the `ALLOWLISTED_CODES` tuple, which exists so that changing the set fails a test.

### What this cost, stated plainly

`unexpected ')' at position 4` is a genuine ParseError message about the caller's own input — exactly what this ticket set out to surface — and it is now withheld as collateral of `CollectionError` sharing its code. [OME-1219](https://linear.app/openmined/issue/OME-1219) files the root fix (give `CollectionError` its own code), which restores it. That is a url4 core contract change and deliberately not smuggled in here.

PR #977, opened 2026-09-17
feat(engine): wire the dead observability config (OME-942)

sergio-bershadsky:
Five loosely-related pieces of engine observability config that are declared and wired to
nothing. An operator who turns a knob that does nothing is worse off than one who has no knob:
the deployment looks instrumented while the evidence never appears. One PR, as OME-942 asks;
each concern is separate below.

Ledger: `docs/work/2026-09-17-OME-942-engine-dead-observability-config.md`.

## 1 — `URL4_CLOUD_LOG_LEVEL` was unreachable from a deployment

`logs.configure()` read it; nothing in the chart set it, so **no deployed pod could be turned to
DEBUG** — the first control anyone reaches for during an incident.

Charted as `config.logLevel` (default `INFO`, read from the code's `DEFAULT_LEVEL` so charting
the knob cannot change the level deployments run at), constrained by `values.schema.json` to the
names `logging.Logger.setLevel` resolves, and rendered to **both** halves: the App's ConfigMap
and the runner pool's `-runner-env` one. Both run `screamingface_engine` through `cli.main`,
which configures logging for every mode — a level reaching one half is the same half-working
knob this ticket exists to remove.

## 2 — `logs.configure()` was bound to `cli.main` only

`uvicorn.run()` installs handlers for the `uvicorn*` loggers **only**, so a
`screamingface_engine` record falls through to `logging.lastResort` and is dropped below
WARNING — the exact regression `logs.py`'s docstring documents. Any other ASGI host (`uvicorn
screamingface_engine.app:create_app_from_env`, an embedding process) reproduced it in full.
`create_app` is the one door they all go through, so it calls `configure()` now. It is
idempotent about its own handler, so the CLI's call plus this one installs exactly one.

## 3 — `/livez` and `/readyz` were static literals the chart never probed

**Owner decision: implement, do not delete.** Both probes targeted `/healthz`, so the readiness
probe could not fail whatever the state of a pod's NATS connection.

> **Correction (review round 2).** An earlier version of this section asserted that "every run
> routed there was accepted and went nowhere". **That was never observed, reported or
> investigated** — it was a plausible mechanism written as history, and it is mechanically
> doubtful besides (with `runner=queue`, a dead connection makes the publish FAIL and surface
> as an error to the caller, not a silent accept). It has been removed from the PR body,
> `ops.py`, `jetstream.py`, `values.yaml` and two test module docstrings. The verifiable defect
> is narrower and sufficient: **the probe could not fail.**

- `/readyz` asks the App's OWN event stream (`app.state.stream`, the JetStream CONSUMER) whether its broker is reachable; 503 when not. The `reason` is a FIXED literal — never the
  NATS URL, never the broker's error text (see round 2 below). `QueueJobRunner`'s separate
  connections are NOT probed, and the docstrings now say so.
- The contract is a new `screamingface_engine/readiness.py` (`StreamNotReadyError` +
  `stream_readiness()`), so `ops.py` imports no concrete adapter and never mentions `nats` — the
  port-boundary discipline `create_app_from_env` already applies with `getattr(job_runner,
  "aclose")`. `check_ready` is optional on the port, like `EventPublisher.flush`: the in-process
  stream has no broker to be unreachable from.
- `_JetStreamConnection.check_ready()` reads `Client.is_connected` rather than `is_closed`
  (nats-py only marks a client closed once its whole reconnect budget is spent — `is_closed`
  reports READY throughout an outage) and issues **no** `account_info()` RPC, because a kubelet
  probes on a timer forever.
- `readinessProbe` → `/readyz`; `livenessProbe` → `/livez`. Liveness stays broker-blind
  deliberately: its failure RESTARTS the pod, so a broker-aware liveness probe would turn a NATS
  outage into a restart loop across every replica. `/healthz` is unchanged and still served.

## 4 — the unused `pods/log` RBAC grant

**Already gone.** `templates/role.yaml` — Role, RoleBinding and the `pods/log` grant — was
deleted wholesale in 4cdfa920 ("cut the chart over to the worker pool and retire the Job
adapter"). No `pods/log`, no `kind: Role`, no `role.yaml` remains anywhere. Recorded, not
invented.

## 5 — `InProcessJobRunner.active_count` claimed to be a gauge

**Owner decision: register it.** It is the admission gate's own input — the number
`max_concurrent_runs` is compared against — and nothing read it, so an operator watching runs
get refused had no series to look at. Now `screamingface_engine_active_runs`, registered through
a getter like the other collectors so it is read at scrape time. A runner with no `active_count`
(the queue runner, whose runs execute in the worker pool) yields **no series** rather than a
confident `0` that would read as "nothing is running" for a saturated fleet.

**Not done, per the ticket:** no scrape endpoint on the run mode. The App's existing `/metrics`
serves the gauge, and `check_layering.py` is green.

## Verification

- `uv run .claude/scripts/run_gates.py screamingface-engine` → **ALL GATES GREEN** (append-only
  check, ruff check, ruff format, pyright, `check_layering.py`, pytest + coverage).
  3101 passed / 16 skipped. Coverage 93% against the 80% floor; `readiness.py` and `ops.py` at
  100%, `metrics.py` at 98%. No gate was weakened and `--skip-append-only` was not used.
- `python3 .github/scripts/verify_chart_wiring.py` → **112/112 checks passed**.
- Every chart assertion is made against the **rendered manifest**, never `helm lint` — lint
  reports success for a chart that cannot render at all.
- **Mutation testing: 13 planted, 13 killed.** Hardcoding the runner's log level; dropping the
  key from the App ConfigMap; relaxing the schema enum; removing `configure_logging()`;
  swallowing the readiness reason; never asking the stream; checking `is_closed` instead of
  `is_connected`; narrowing the translated exception set; reverting `/readyz` to a constant;
  pointing `readinessProbe` back at `/healthz`; pinning the gauge to `0.0`; removing the
  absent-runner guard; blanking the gauge's HELP text; caching the first reading. Each one took
  down the tests it should have.
- No assertion reads `repr()` of an object anywhere — every one walks parsed Prometheus samples,
  parsed YAML, real `logging` handlers, or response JSON.

## Three things a reviewer should look at

1. **`/readyz` treats an UNWIRED stream as ready**, not 503. The first draft failed it, which
   contradicted the prior `test_docs_ops.py::test_livez_and_readyz_return_200` (it builds
   `create_app()` with no stream). That prior test does not genuinely have to change:
   `create_app_from_env` always wires a stream, so "unwired" is a composition-time shape, not an
   outage, and the defect being fixed — a *wired* stream whose broker is gone — still fails the
   probe. My new test was changed to match the prior contract; the prior test is untouched.
2. **`livenessProbe` moved to `/livez`**, which the ticket text does not literally ask for (it
   names only `readinessProbe` → `/readyz`). Being the liveness probe's target is the only thing
   "implement `/livez`" can mean; leaving it unprobed would leave half the dead surface in place.
3. **`tests/conftest.py` is new.** `create_app` now sets `propagate = False` on the
   process-global `screamingface_engine` logger, so the first test to build an App silently
   changed how every later test's records travelled — 30 prior `caplog` tests failed *depending
   on order*, and all of them pass in isolation. The conftest saves/restores that logger around
   every test and forwards records to `caplog` only while propagation is off (so nothing is
   captured twice). **No prior test was deleted, weakened, skipped or edited**, and the
   production contract it isolates is asserted directly in `test_app_configures_logging.py`.

Refs: OME-942

---

## Review round 2 (2026-09-17)

Eight findings were raised. Each was checked against the source before acting. Full reasoning,
including the two refutations, is in the ledger
(`docs/work/2026-09-17-OME-942-engine-dead-observability-config.md`, "Review round 2").

### Confirmed and fixed

| # | Finding | Fix |
|---|---|---|
| F1 HIGH | `/readyz` echoed `self._url` and the raw broker exception to an unauthenticated caller (no auth dependency; a single `/` PathPrefix HTTPRoute; `nats://user:pass@host` is the standard auth form) | `StreamNotReadyError` now has a CONTRACT — a short fixed literal naming the failure class. `stream_readiness` logs the full detail server-side at WARNING, then caps the rendered reason at 120 chars and strips control characters. Both the raiser and the boundary. |
| F2 HIGH | `check_ready` awaited `_jetstream()`, which dials under `_connect_lock` with no timeout whenever `_js` is None — probes pile up on the loop that pumps every WebSocket | `READINESS_DIAL_TIMEOUT_S = 3.0` (adapter) and `READINESS_TIMEOUT_S = 4.0` (endpoint boundary). A test reads `values.yaml` and pins both **below** `readinessProbe.timeoutSeconds`, so raising either without raising the probe fails the suite. |
| F3 MED | `test_readiness_never_blocks_on_a_broker_round_trip` was a tautology (pre-set `_js`/`_nc`, cached fast path) | Correct. The prior test was **left in place** (append-only) and supplemented by three tests that drive the UNCACHED path with `nats.connect` patched to a coroutine that never returns. |
| F4 MED | Readiness asked of the wrong connection | Confirmed as fact; fixed as **documentation**, not code — see D7 below. |
| F6 MED | Unverified incident narrative in five places | Removed everywhere; see the Correction box above. This was the single most important item. |

### Judged, with reasoning

**D7 — the probe's scope is now stated exactly, and was narrowed rather than widened.**
`app.state.stream` really is the JetStream *consumer*; `QueueJobRunner` holds three separate
NATS connections (`RunQueue`, `JetStreamPublisher`, `ControlClient`) that `/readyz` does not
ask. The reviewer is right on the facts. The overclaim was in the prose, so the prose is what
changed: widening the probe to those three would make D8 strictly worse, putting three more
connections' failures in front of every endpoint this Service fronts.

**D8 — the availability trade is real, and it is OPEN for the owner.** Broker-aware readiness
on a single-Service, single-`/`-PathPrefix deployment means a TOTAL NATS outage empties every
replica's endpoints at once, including paths that need no broker (token mint, `/docs`, catalog
REST, artifact GETs). My reading: readiness gating pays off in **partial** failure (a wedged
client on one pod; a cold start) and pays **nothing** in total failure, where there is no
healthy side to route to — it only converts "run streaming degraded" into "whole API 503".
Which case dominates depends on the deployment's NATS topology, which is an operator fact I do
not have. **Nothing was guessed:** `failureThreshold` and the grace settings are left at the
Kubernetes defaults, and the trade is written into `values.yaml` next to the probe.
**If the owner wants readiness to stay broker-blind, the change is one line** — drop the
`stream_readiness` call from `ops.readyz`. The leak and hang fixes stand either way.

### Refuted, with evidence (no code changed)

**F7 — `_CaplogBridge` semantics.** The reviewer is right that `Handler.handle` applies filters
but not levels, and that forwarded records miss `caplog.get_records(phase)`. Refuted as a live
problem: there are **zero** uses of `caplog.get_records(` anywhere in
`apps/screamingface-engine/tests`, so the second divergence has no caller; and the bridge only
fires while `propagate` is False, i.e. only after a test has built an App, so a test that never
builds one is unaffected. The "test infrastructure in a config-sweep ticket" point is accepted
as a naming point and was already recorded in full as Deviation 4.

**F8 — `_QueueShapedRunner` in `test_active_runs_gauge.py`.** The gauge's contract *is* the
shape: `register_active_runs_metrics` takes the `JobRunner` port and duck-types `active_count`
precisely so the App imports no concrete adapter — `check_layering.py` would fail a metrics test
that imported `QueueJobRunner`. A test naming the concrete class would test the wiring, not the
gauge. The hazard described is real but belongs to whoever adds `active_count` to
`QueueJobRunner`; the guard for it cannot be written today because the attribute does not exist
(verified again on this branch).

### Verification

- **Append-only: clean.** 12 tests added; no prior test function, assertion or fixture deleted,
  weakened or rewritten — including round 1's own. The gate's own `append-only test check`
  passes. (Two test *module docstrings* were corrected for the invented narrative; no test code.)
- **Every new test confirmed RED before its fix.**
- **Mutation testing: 11 planted, 11 killed** — leak the URL back into each of the three
  messages (2/1/1 tests died); **drop the adapter's dial bound, the exact mutation the reviewer
  named and the old suite survived (3)**; raise the dial bound past the kubelet timeout (1);
  drop the boundary bound (2); raise the boundary bound (1); drop the length cap (1); drop the
  control-char scrub (1); drop the server-side log (1); swallow the not-ready condition (6).
- **No assertion reads `repr()`** of anything: the new tests walk `str(exc)`, decoded response
  bytes, parsed JSON members, `record.getMessage()`, parsed `values.yaml` and module constants.
- **Gates:** `uv run .claude/scripts/run_gates.py screamingface-engine` — **ALL GATES GREEN**
  (append-only, ruff check, ruff format, pyright, check_layering, pytest + coverage ≥ 80).
  `python3 .github/scripts/verify_chart_wiring.py` — 112/112 passed.


sergio-bershadsky:
## Review round 2 — what I fixed, what I refuted, and one thing I am NOT deciding

Pushed as `d445503c`. Full reasoning in the ledger under "Review round 2"; the PR body now
matches the code.

### The most important item first: the invented history is gone

"every run routed there was accepted and went nowhere" / "accepted runs it could neither queue
nor stream" was committed as permanent fact in **five** places — `ops.py`'s `readyz` docstring,
`jetstream.py`'s `check_ready` docstring, `values.yaml`, and the module docstrings of
`test_readiness_probe.py` and `test_chart_render_probe_paths.py` — plus the PR body.

**No such incident was observed, reported or investigated.** I had no evidence for it. It is
also mechanically doubtful: with `runner=queue`, submission publishes over NATS, so a dead
connection makes the publish *fail* and surface as an error to the caller rather than a silent
accept. Every instance is deleted or rewritten, each site now carries an explicit "no incident
is claimed here" so the next reader cannot re-inherit the story, and the ledger records the
episode rather than quietly erasing it. The verifiable defect is narrower and sufficient:
**the probe could not fail.**

### CONFIRMED and fixed

**F1 (HIGH) — the leak.** Confirmed by reading the chain: `check_ready` raised
`StreamNotReadyError(f"{self._url} is unreachable: {exc}")`, `ops.readyz` rendered it verbatim,
the ops router has no auth dependency, and `templates/httproute.yaml` has exactly one rule —
`PathPrefix: /` — so the body is public, not kubelet-only. `values.schema.json` constrains
`natsUrl` only to a `nats://` prefix, and `nats://user:pass@host:4222` is the standard nats-py
auth form. Fixed at **both** ends, because one adapter forgetting the rule must not be a leak:

- `StreamNotReadyError` now carries a documented CONTRACT — a short FIXED literal naming the
  failure class. The three production messages are `event stream broker is unreachable`,
  `… is not connected`, `… dial timed out`; the transport detail is chained as `__cause__`.
- `stream_readiness` logs the full reason *and* the chained exception at WARNING (server-side),
  then caps the rendered reason at `MAX_REASON_CHARS = 120` and strips control characters.

**F2 (HIGH) — the hang.** Confirmed, and you are right that it was worse than the literal it
replaced. Two bounds, both tested as behaviour:
`READINESS_DIAL_TIMEOUT_S = 3.0` in the adapter (cancelling the dial unwinds
`async with self._connect_lock`, so the next probe gets its own budget instead of queueing),
and `READINESS_TIMEOUT_S = 4.0` in `stream_readiness` for an adapter with no bound of its own.
A test reads `deploy/helm/values.yaml` and pins **both** below `readinessProbe.timeoutSeconds`,
so raising either without raising the probe fails the suite — the numbers are load-bearing, not
decorative.

**F3 (MED) — the tautology.** Correct: pre-setting `_js`/`_nc` meant `_jetstream()` returned on
the cached fast path and the `wait_for` could never fire. **I did not touch the prior test**
(append-only) — it still pins the cached path, which is worth pinning. Three new tests drive the
uncached branch with `nats.connect` patched to a coroutine that never returns, including one
asserting `_connect_lock` is released afterwards.

### The two you asked me to judge

**"Readiness is asked of the wrong connection" — you are right on the facts (ledger D7).**
`app.state.stream` is `build_stream_consumer(settings)` → a `JetStreamConsumer`;
`QueueJobRunner` holds three separate connections (`RunQueue`, `JetStreamPublisher`,
`ControlClient`) built in `adapters/factory.py`. `/readyz` asks none of them, so a pod with dead
runner connections and a live consumer does still report ready.

The overclaim was in the **prose**, so the prose is what I changed — I narrowed the claim rather
than widening the probe. Widening it makes the next point strictly worse: it would put three
more connections' failures in front of every endpoint this Service fronts. Every docstring now
states the scope exactly and names what is *not* probed.

**Blast radius — I am NOT deciding this one, and I have not silently assumed it either
(ledger D8).** You are right that the PR reasoned carefully about liveness and said nothing
about the symmetric readiness cost. My reading, stated so you can overrule it cheaply:

- Readiness gating pays off in **partial** failure — one pod's client wedged while its peers are
  fine, or a cold start that has not connected. Both are real; both are what this ticket targets.
- It pays **nothing** in **total** failure. With one shared NATS service, failure is usually
  total, and removing every replica cannot route around an outage with no healthy side — it only
  converts "run streaming degraded" into "whole API 503 at the gateway", taking token mint,
  `/docs`, catalog REST and artifact GETs with it.
- Which case dominates depends on your NATS topology, which is an operator fact I do not have.

**Nothing was guessed:** `failureThreshold` and the grace settings are left at the Kubernetes
defaults rather than tuned to a number nobody chose, and the trade is now written into
`values.yaml` directly above the probe definitions, where whoever makes the call will read it.
**If you want readiness broker-blind, it is one line** — drop the `stream_readiness` call from
`ops.readyz`. The leak and hang fixes stand either way. Flagging this as needing your sign-off
before merge.

### REFUTED (no code changed — evidence, not disagreement for its own sake)

**F7 — `_CaplogBridge`.** You are right about the mechanism: `Handler.handle` applies filters
but not levels, and forwarded records never reach `caplog.get_records(phase)`. Refuted as a
*live* problem here: there are **zero** occurrences of `caplog.get_records(` anywhere under
`apps/screamingface-engine/tests`, so the second divergence has no caller; and the bridge fires
only while `propagate` is False — i.e. only after a test has built an App — so a test that never
builds one is byte-for-byte unaffected. The "test infrastructure smuggled in by a config-sweep
ticket" point I accept as fair naming; it was already recorded in full as Deviation 4.

**F8 — `_QueueShapedRunner`.** The gauge's contract *is* the shape.
`register_active_runs_metrics` takes the `JobRunner` port and duck-types `active_count`
specifically so the App imports no concrete adapter — `check_layering.py` would fail a metrics
test that imported `QueueJobRunner`. A test naming the concrete class would be testing the
wiring, not the gauge. The hazard you describe is real but belongs to whoever adds
`active_count` to `QueueJobRunner`; the guard for it cannot be written today because the
attribute does not exist (re-verified on this branch).

### Verification

- **Append-only: clean.** 12 tests added; no prior test function, assertion or fixture deleted,
  weakened or rewritten — including round 1's own. The gate's `append-only test check` passes.
  (Two test *module docstrings* were corrected for the invented narrative; no test code touched.)
- **Every new test confirmed RED before its fix.**
- **Mutation testing: 11 planted, 11 killed** — including **the one you named**: dropping the
  adapter's dial bound, which the round-1 suite survived, now kills 3 tests. Also: leaking the
  URL back into each of the three messages (2/1/1); raising either bound past the kubelet
  timeout (1 each); dropping the boundary bound (2); dropping the cap (1); dropping the
  control-char scrub (1); dropping the server-side log (1); swallowing the not-ready condition
  entirely (6).
- **No `repr()` assertions** anywhere: the new tests walk `str(exc)`, decoded response bytes,
  parsed JSON members, `record.getMessage()`, parsed `values.yaml` and module constants.
- **Gates:** `uv run .claude/scripts/run_gates.py screamingface-engine` — **ALL GATES GREEN**.
  `python3 .github/scripts/verify_chart_wiring.py` — 112/112. No gate was weakened; the one
  failure I hit (`PLR0911`) was fixed in the code.

Not merging — waiting on CI and on your call on D8.

PR #978, opened 2026-09-17
feat(screamingface): example notebook walks the imported benchmark catalogue

khoaguin:
Adds the example notebook that is the front door to the imported benchmark catalogue: list the two origin groups, read an imported board's card, run a fusion against it.

## Problem / Solution

**Problem:** we imported ten inspect_evals boards onto the catalogue, but nothing shows a researcher what's now on the shelf or how to run one — the capability exists with no front door.

**Solution:** one `examples/` notebook that opens the catalogue, shows our boards and imported boards as two groups, picks an imported board, and runs a fusion against it — the demo a newcomer runs first.

## Before / After

*(No diagram: this is an add-one-notebook change — it would be two boxes and one arrow.)*

### Before
- Trigger: a researcher hears "we have imported benchmarks now" and opens `examples/`.
- Today: no notebook touches imported boards; they reverse-engineer the SDK calls from tickets or source.
- Cost: the epic's headline win (benchmark breadth) is invisible at the surface where users actually live.

### After
- Same trigger.
- Now: `12_imported_benchmarks.ipynb` walks `sf.benchmarks.list()` → two origin groups → `sf.benchmarks.get("inspect-gsm8k")` → `sf.evaluate(math_panel, benchmark="inspect-gsm8k", limit=2)` → the report widget and per-case readout.
- Win: the imported catalogue is discoverable and runnable in one guided pass; the notebook doubles as living documentation of the two-group listing.

### Don't regress
- The notebook consumes the public SDK surface only — no engine internals, no inspect runner (pinned by the new contract test).
- Run All never spends by default: run cells follow the existing benchmark-notebook idiom (small `limit`, prose warning), and the test re-asserts the `RUN_EVALUATION` ban.
- Notebooks stay generated artifacts: the `.ipynb` is emitted by `build_notebooks.py`, byte-checked by `check_notebooks.py`.

## Heads-up for the acceptance run

No environment serves the imported boards today: the local `screamingface up` runtime can never install the engine's `inspect` extra (declared uv conflict with `litellm`), and the engine Dockerfile runs `uv sync` without `--extra inspect`. The notebook says so honestly — its listing degrades to a single group on a plain engine, and its setup cell points `SCREAMINGFACE_ENGINE_URL` at an inspect-capable engine. The end-to-end acceptance (two groups visible, `inspect-gsm8k` evaluates) needs an engine deployed with that extra — likely the remaining tail of `OME-1116`.

## Review order

1. **`packages/screamingface/scripts/build_notebooks.py`** — the whole change's idea lives in the new `_imported_catalogue()` builder (notebooks are generated, so the builder is the source of truth). Verify: the prose about where imported boards live matches the extras-conflict reality above, and the model/param idiom matches the neighbouring `_medxpert_e2e()`.
2. **`packages/screamingface/tests/test_imported_catalogue_notebook.py`** — the contract that keeps the notebook honest: listing → imported board id → fusion evaluate, public surface only, never paid by default. Verify: the hardcoded `inspect-gsm8k` matches the engine's published board id.
3. **`packages/screamingface/examples/12_imported_benchmarks.ipynb`** — mechanical skim: generated output of step 1; `check_notebooks.py` already proves cell-for-cell equality.
4. **`docs/work/` + `docs/tasks/`** — the ledger and mirror close; the ledger's Deviations section carries the deployment-gap finding.

## Some results
Screenshot 2026-09-18 at 13 51 36
Screenshot 2026-09-18 at 15 19 07
Screenshot 2026-09-18 at 15 19 13
Screenshot 2026-09-18 at 15 19 19

[inspect-gsm8k-report.json](https://github.com/user-attachments/files/32389261/inspect-gsm8k-report.json)
[inspect-mmlu-report.json](https://github.com/user-attachments/files/32389262/inspect-mmlu-report.json)
[inspect-boolq-report.json](https://github.com/user-attachments/files/32389267/inspect-boolq-report.json)

### With seed
```
PANEL_PARAMS = {"max_tokens": 8192, "temperature": 0.0}
SYNTHESIS_PROMPT = (
    "You are given several models' step-by-step solutions to a grade-school math word "
    "problem. Check each chain of arithmetic, resolve any disagreement by re-deriving the "
    "disputed step, and answer with the single final number."
)

member1 = sf.Model(model="openrouter/qwen/qwen3.7-flash", params=PANEL_PARAMS)
member2 = sf.Model(model="openrouter/google/gemini-3.8-flash", params=PANEL_PARAMS)
synth = sf.Model(
    model="openrouter/qwen/qwen3.8-flash", params=PANEL_PARAMS, prompt=SYNTHESIS_PROMPT
)
math_panel = sf.Fusion(name="math_panel", members=[member1, member2], synthesizer=synth)

math_panel
```

- `sf.evaluate(math_panel, benchmark="inspect-gsm8k", limit=2, answer_seed=42)` gives this result [inspect-gsm8k-report-seed42.json](https://github.com/user-attachments/files/32409301/inspect-gsm8k-report-seed42.json)
- `sf.evaluate(math_panel, benchmark="inspect-gsm8k", limit=2, answer_seed=7)` gives this result [inspect-gsm8k-report-seed7.json](https://github.com/user-attachments/files/32409305/inspect-gsm8k-report-seed7.json)


keelancj:
**Request changes: one blocking finding and two nonblocking corrections.** Reviewed #978 at `e3e574b` using the guide from #982.

### Action required: the documented local setup fails partway through

The notebook says “every cell below still works” without imported boards, but its unconditional `sf.benchmarks.get("inspect-gsm8k")` raises `PlanningError`. I reproduced this with the real SDK and a mocked catalogue containing only native boards. See [setup and lookup](https://github.com/ScreamingFace/screamingface/blob/e3e574bc06070ecf392733ac6a8b0e2dbba4a59e/packages/screamingface/scripts/build_notebooks.py#L1316-L1352).

Make the inspect-capable engine an explicit prerequisite for proceeding beyond listing, or handle the missing board with clear guidance. Add an offline check for that path. The deployment gap is already documented; the new defect is promising that the unsupported path works.

### Nonblocking: the test does not establish “never pays by default”

The [test](https://github.com/ScreamingFace/screamingface/blob/e3e574bc06070ecf392733ac6a8b0e2dbba4a59e/packages/screamingface/tests/test_imported_catalogue_notebook.py#L42-L52) only bans `RUN_EVALUATION = True`. Executing the notebook cells with an evaluation recorder reaches `sf.evaluate` unconditionally. Since the ledger deliberately adopts direct evaluation, correct the test’s claimed guarantee and PR description.

### Nonblocking: changing boards also requires changing the synthesis prompt

The [“swap the id” guidance](https://github.com/ScreamingFace/screamingface/blob/e3e574bc06070ecf392733ac6a8b0e2dbba4a59e/packages/screamingface/scripts/build_notebooks.py#L1399-L1401) retains a prompt demanding a single final number. MMLU and ARC expect lettered answers. Tell users to adapt the prompt too. This is a code-based conclusion; I did not measure model behavior.

The public-SDK notebook approach fits the task. **Checks run:** 24 relevant tests and deterministic notebook verification passed, plus the offline execution probes above. No paid calls were made. Live imported-board evaluation and the full CI suite remain unverified.

keelancj:
**Request changes: three blocking findings.** Reviewed at `498d8e72` using the review guide from #982 (`ab255233`). These findings concern additions since the first posted review; none was introduced by the latest fix commit, `498d8e72`.

### Action required: interrupted asset writes are treated as complete

The [resume check](https://github.com/ScreamingFace/screamingface/blob/498d8e72da167c6418626e5eff806ee443b5afe7/packages/screamingface/justfile#L84-L91) skips any bundle containing `cases.json`, but preparers write that file non-atomically. A failed final write therefore leaves a broken bundle that subsequent attempts skip. The latest fix correctly moves the NLTK download before `cases.json`; interruption during the JSON write remains a separate failure case.

**Evidence:** Running the real IFEval preparer with a 32,768-byte file-size limit raised `OSError` and left truncated, invalid JSON. The recipe’s file-existence guard still selected “skip.” The non-atomic writer already existed; the resume check added in `8a4bc7e2` makes its incomplete output count as complete.

Publish the completion file atomically and add a regression check proving that a failed final write gets rebuilt on retry.

<details>
<summary>Reproduction details</summary>

In a disposable subprocess with the pinned Engine installed, set `resource.RLIMIT_FSIZE` to `(32768, 32768)` and ignore `SIGXFSZ`. Call `screamingface_engine.benchmarks.ifeval.prepare.build(official_rows(), out, expected_count=541)` with a fresh temporary directory. The call raises `OSError: [Errno 27] File too large`; `cases.json` remains 32,768 bytes and JSON decoding fails at character 32,694. The recipe’s exact `[ -f "$out/cases.json" ]` guard succeeds. This uses the bundled official rows, with no downloads or paid calls.

</details>

### Action required: point notebooks at the local Scoreboard too

The [recipe exports only the Engine URL](https://github.com/ScreamingFace/screamingface/blob/498d8e72da167c6418626e5eff806ee443b5afe7/packages/screamingface/justfile#L158-L166). An explicit Engine URL disables the SDK’s local-stack discovery, leaving Scoreboard at `https://leaderboard.dev.screamingface.ai`. Opt-in notebook submissions consequently target the hosted deployment despite the recipe starting a local Scoreboard.

**Evidence:** An offline check with the actual SDK, setting only `SCREAMINGFACE_ENGINE_URL=http://127.0.0.1:9111`, produced that local Engine URL and the hosted Scoreboard URL. This recipe behavior was added in `c6d72fe6`.

Export both effective service URLs and add an offline endpoint check.

### Action required: preserve checkout ownership validation

The [running-status branch](https://github.com/ScreamingFace/screamingface/blob/498d8e72da167c6418626e5eff806ee443b5afe7/packages/screamingface/justfile#L115-L123) skips `screamingface up`, bypassing its check against adopting another checkout’s services. Status checks process ownership and health; it does not establish that the current checkout owns the stack. The existing `_ensure_adoptable` explicitly enforces OME-1001’s rule against silently running against another branch’s services.

**Evidence:** An offline probe with mocked state reading, process identity, and health returned `running` for state belonging to another checkout. The actual `_ensure_adoptable` validator rejected that same state. The bypass was added in `28e7ea9f`.

Call `up` even for an existing stack, preserving the flag that controls shutdown, and test the foreign-checkout case.

The notebook approach fits the ticket, and the missing-board guidance and board-specific prompts address the earlier findings. Nonblocking: refresh the PR description, which still references deleted tests and claims Run All never spends.

**Checks and limits:** 95 targeted Engine tests and deterministic notebook verification passed; GitHub CI was green. No paid calls or live stack execution were performed.

PR #984, opened 2026-09-17
OME 1148 : Onboarding contracteval

siddhant230:
## Summary
[Linear Ticket](https://linear.app/openmined/issue/OME-1148/onboard-contracteval-as-a-judge-free-clause-extraction-benchmark)

Onboards [ContractEval](https://arxiv.org/abs/2508.03080) (NLLP 2025) as an Engine-owned board: 4,182 clause-extraction questions over 102 real commercial contracts (CUAD test split). The model receives a full contract plus one of 41 clause categories and must quote the answering sentences **verbatim** or reply `"No related clause."` — graded by deterministic string containment, so **no judge and zero grading tokens**. Headline score is F1 from a dataset-level confusion matrix, not a mean of case scores.

**Asana:** n/a — technical work never goes to Asana (CLAUDE.md rule 7). Work item: `OME-1148`.

## Components touched

- `apps/screamingface-engine` — new `benchmarks/contracteval/` (8 modules), registered in `builtins.py`; one additive row in `test_benchmark_declaration.py`
- `packages/screamingface` — `_BENCHMARKS` + asset manifest in `_runtime/cli.py`, `build_notebooks.py`, generated `examples/12_contracteval.ipynb`
- `docs/` — spec, plan, work ledger, task mirror

22 files, +2,768 / −1.

## Test plan

- [x] Ran the touched app's local gates — `run_gates.py screamingface-engine` **ALL GREEN**, `run_gates.py screamingface` **ALL GREEN** (incl. notebook determinism, `uv build`, distribution check)
- [ ] Gateway request/refresh change → n/a, no gateway change
- [ ] Screenshots / recording attached → n/a, no UI change
- [x] **71 ContractEval unit tests** across grading, prepare, case-evaluation and aggregate
- [x] **Differential test against the paper's own harness** — 2,742 (output, gold) pairs over 400 real CUAD rows: **zero** mismatches in verdict, abstention and Jaccard
- [x] Assets baked end to end — 4,182 cases / 1,244 positive / 2,938 negative. The 1,244 is the exact constant the reference hardcodes as its laziness denominator, which is the strongest available evidence we bake the exam it scored
- [x] Live `limit=8` run on the local stack — F1 0.80, coverage 1.0, every metric reproducible by hand from the per-case list

## Cross-service notes

Spans `apps/screamingface-engine` and `packages/screamingface`. Filed as **one ticket** rather than a D9 epic + sub-issues, following the GDPval (`OME-971`) and MedXpertQA (`OME-1126`) precedent — the SDK side is a three-line registration plus a generated notebook. Deliberate; say the word if you'd rather split it.

### This is a rebuild of PR #865

That PR was closed unmerged and `main` then moved **191 commits** under the onboarding-infra overhaul. Rather than rebase, I measured the blast radius first: diffing MedXpertQA across those commits showed **only its `aggregate.py` changed, with its tests untouched** — so the public contract had held. That predicted seven modules would port byte-identical and only the reducer needed migrating, which is exactly what happened; all prior tests passed **unmodified**. Prior work is preserved at tag `OME-1148-v1-pre-overhaul`.

`aggregate.py` drops 414 → ~250 lines: `ScoredPath` absorbs the failure ladder, roll call, row filing and result assembly.

### Both review findings from #865 are fixed

They survived the port, because I moved those files without re-litigating them. Thanks for catching them:

1. **`preflight()` was never called** — defined, exported, and invoked from nowhere, so a missing answer key surfaced only after inference was paid for. The cases route now preflights the whole bundle before serving, memoized like gdpval's (1.63s once per process for 4,182 records, 0.003ms after; only successful passes cached, so a broken bundle re-fails every call).
2. **Spec and code disagreed on the laziness denominator.** You left the choice open; **graded** wins and D-3 now says so. Dividing by *selected* positives credits a model for rows it never saw — your own example reports 50% when the model answered exactly one positive case and abstained on it, which is the `failure_policy` invariant running backwards. It's also the only denominator in the block that wouldn't be over graded cases.

Both are pinned by tests named after the finding.

### Reviewer notes

- **F1, not accuracy, is the headline.** 70.3% of rows have no clause, so a model that always abstains scores ~70% accuracy while answering nothing. F1 scores it 0. Accuracy is still published as a metric.
- **Our July screening note was wrong** and the spec records the correction: ContractEval is *not* span-overlap with partial credit. It's all-or-nothing verbatim containment, which is why the grade method is `containment`.
- **Three further named deviations from the reference**, each tested: a zero guard on `2PR/(P+R)` (the reference raises `ZeroDivisionError` for an always-abstaining model — realistic at 70% negatives); `temperature` omitted from the notebook because `openai/gpt-5.5` answers **404** and `claude-opus-4.8` **400** for any value; and the full 4,182 rows vs the paper's 4,128 (its own usable-output count, not a different key).
- **No truncation path.** Tokenizing all 4,182 contexts: p50 5,357, max 63,389 — nothing near the 128k floor, so `MAX_CONTEXT_TOKENS` is a build-time guard rather than dead trimming code.
- **Not importable from inspect_evals** — re-verified on this branch. `ContractBench` there is software/API contracts (presigned URLs, OAuth tokens, HMAC webhooks), a name collision. If ContractEval is ever contributed upstream, this board collapses to two rows in `screamingface_engine_inspect/boards.py`.
- **⚠️ System prompt is sent as user text.** The reference uses `role: "system"`; `candidate()` takes a single `input` string and gives boards no system-role channel, so `render_case_input` concatenates. Worth a shared ticket rather than a per-board workaround.
- **Naming:** the Linear title still says "span-extraction", which the protocol reading disproved — doc files are already renamed to `-containment`.


khoaguin:
## Review — protocol fidelity, grading integrity, boundaries, tests

Reviewed against the ticket, spec (F-1…F-10, D-1…D-8), plan and ledger, plus the spine and the `medxpert`/`ifeval` boards on `main`. Read-only: nothing run that spends money or downloads the dataset.

**The board itself is well built.** Sealed envelope intact, url4 scoping correct, route registration enforced before first spend, failure-code table complete, money bounded. Detail in "Verified good" at the bottom — worth reading, because the list below is only the other half.

What needs work is the *evidence*. The two things this board publishes — a number claimed to be the paper's, and the metrics derived from it — are each protected by proofs that cannot fail today.

---

## Action required

### 1 · Fidelity to the paper is asserted, never pinned — and every provenance pointer is dead

**The bug:** the exam claims "certified identical to the original", and no certificate exists in the repo.

**Why it hurts:** this is a judge-free board, so the prompt *is* the exam — `prompts.py:8-10` says so itself ("Do not rephrase or summarize" is the difference between a correct answer and a zero) — and those bytes feed `compute_revision` (`definition.py:57-58,78-79`), so they define the board's published identity. Four checks, all failing:

- `docs/spec/…-containment.md:26` says the harness is "Mirrored into `.refs/contracteval/`". `git ls-tree pr984` has no `.refs/`.
- `grading.py:5` sends the reader to `docs/spec/2026-09-09-OME-1148-contracteval-span-extraction.md` for "the full transcription". That file was never committed on any branch; the shipped spec is the `2026-09-17-…-containment.md` one.
- The two surviving citations disagree: `prompts.py:4` says the user template is `proprietary_model.py` lines **18-27**, spec F-8 (`:90-91`) says lines **18-32**.
- `grep -rn "SYSTEM_PROMPT\|USER_TEMPLATE\|render_case_input"` over the four new test files returns **nothing**, and no committed test mentions the differential run. It ran in a scratchpad, and the ledger reports **2,970** pairs pre-overhaul vs **2,742** after for the same 400 rows — so it is not reproducible from this repo either.

Net: a transcription typo in `SYSTEM_PROMPT` is undetectable today, and would move our F1 off the paper's with every gate green.

**The fix:** commit the four transcribed reference functions as test-only material beside `grading.py`, add a prompt-bytes test, repoint `grading.py:5` at the containment spec, reconcile the line ranges to one number.

**What pins it:** a differential parity test over a committed CUAD row sample with the row count asserted, plus `test_contracteval_prompts.py`. The IFEval board is the model — all 541 official rows through the original harness.

### 2 · All three published metrics are pinned at fixtures where the right formula and the wrong ones give the same number

**The bug:** the witnesses agree with the code, and would agree just as readily with a bug.

**Why it hurts:** the only non-degenerate confusion-matrix test (`test_contracteval_aggregate.py:81`) uses 2 TP / 1 FN / 1 FP / 1 TN, so precision == recall == 2/3 — its own docstring at `:82` says "P=2/3, R=2/3, F1=2/3, F2=2/3". Computed, not eyeballed:

```
2 TP / 1 FN / 1 TN / 1 FP
  F1 (shipped)        0.6667
  F2 (shipped)        0.6667
  F2 beta-inverted    0.6667     5PR/(P+4R)
  arithmetic mean     0.6667     (P+R)/2
  precision, recall   0.6667

3 TP / 1 FN / 2 TN / 2 FP        <- proposed
  F1                  0.6667
  F2                  0.7143
  F2 beta-inverted    0.6250
  arithmetic mean     0.6750
  precision           0.6000
  recall              0.7500
```

Same shape in both laziness tests. Fixture A (`:141`): 1 false abstention, 2 graded positives, 2 total abstentions -> 0.5 by either denominator. Fixture B (`:306`), the one named `test_the_denominator_is_GRADED_positives_not_selected_positives`: 1/1/1 -> 1.0 by either. So an implementation dividing by **total abstentions** instead of graded positives passes the whole suite — including the test named after the review finding that D-3 was amended over. (Credit where due: both fixtures *do* discriminate against the reference's hardcoded 1244 and against "selected positives" — B gives 0.5 there vs the asserted 1.0. It is only the abstention-count variant that slips through.)

And `test_raw_output_is_the_verdict_not_the_abstention_flag` (`:289`) uses `correct=True, abstained=False`, where `verdict` and `not abstained` are indistinguishable — the exact confusion the test's name rules out.

**The fix:** asymmetric matrix as above; unequal graded-positive and abstention counts for laziness; for the evidence test, a reply that both quotes the gold span *and* says "no related clause" (`correct=True, abstained=True`).

**What pins it:** the same three tests, once their fixtures discriminate. Note F1 is symmetric in P and R, so a precision/recall swap is caught by the separate `precision`/`recall` asserts — which only works once P != R.

---

## Review recommended

**3 · Asset freshness keys only on the dataset revision, so a prompt edit re-addresses every route but leaves the old booklet baked.** `_benchmark_fingerprint` returns `contracteval:<DATASET_REVISION>` and nothing else (`cli.py:726-731`); `_benchmark_status` compares the manifest against it (`cli.py:711`); `_prepare` skips a bundle that reads "prepared" (`cli.py:562-563`). Edit one word of `SYSTEM_PROMPT`: `compute_revision` changes, every route is re-addressed, the report publishes the new revision — and `screamingface prepare contracteval` prints "already prepared" and serves the old text under the new identity. `PREPARER_REVISION` is equally invisible to the fingerprint. Pre-existing in shape (gdpval bakes reference text the same way), but this is the first board where an ordinary in-repo string edit triggers it. Fix is one line: fold the preparer revision into the fingerprint. Worth a ticket covering both boards.

**4 · The one production site that turns a reply into a graded record has no test.** `runtime.py:121` `_check` is where `verdict` / `is_abstention` / `jaccard` meet a real candidate payload and become the CHECK_SCHEMA record. Every test builds that record by hand (`_record`, `test_contracteval_aggregate.py:45`); the only runtime symbol any contracteval test imports is `_cases`. So `grading.py` and the confusion matrix are each proven against a fixture, never against each other through production code. Unpinned today: that an empty reply yields `correct=False`/`abstained=False` (the D-5 case), and that `_check`'s output satisfies `bind_case_evaluation`'s own validator — a field drift there fails only at runtime, after inference is paid for.

Related, same file: spec §5 promises "only a successful pass is cached, so a broken bundle re-fails on every call". The code honours it (`runtime.py:115` assigns `memo["payload"]` strictly after `preflight` at `:114`), but both preflight tests construct a fresh closure and call it **once** (`test_contracteval_prepare.py:160,177`). Hoist that one line above `preflight` and the suite stays green while a broken bundle gets served on call 2. Binding `served = _cases(tmp_path)` once and asserting `raises` twice costs two lines.

**5 · No resolution-level test of the composed url4 expression.** I verified the scoping by hand against `ifeval/_build` and it is **correct** — every `$name` binds, the `$item` rebinding inside the iterate is handled, and the object-shaped `attempt_records_endpoint` is genuinely the right helper (the comment at `definition.py:115-117` is accurate; the array-shaped sibling parses with `json_array` and would reject that payload). This finding is about the absence of a durable check, not a suspected defect: nothing alarms on a future slip, and the resolver answers an unknown name with empty text. That is OME-1126 exactly — unit tests green while the composed expression shipped an empty prompt to a paid model.

**6 · The reference's system-*role* instruction is delivered as user text, recorded only in a hand-off note.** `render_case_input` (`prompts.py:39`) concatenates `SYSTEM_PROMPT` into the candidate input, and `prepare.py:79` bakes the result into `cases.json`. The system prompt *is* sent — MedXpertQA's dead `ANSWER_SYSTEM` bug is not repeated here, and this board has no orphan prompt constant. But the reference splits the two roles deliberately, so ours is a different request with a different cache key. Spec D-1…D-8 has no entry for it and the board `description` (`definition.py:151-161`) carries no caveat, where MedXpertQA's description spells its comparability limit out. One sentence in the description plus a D-9.

**7 · The captured row count is neither in the lockfile nor checked at bake time.** 4,182 lives as a bare literal in `definition.py:47` (used again at `:134` as `available_case_count`), outside `compute_revision`; `pins.py` carries no count at all; `emit` reports `"cases": len(cases)` (`prepare.py:108`) and compares it to nothing. Bump `DATASET_REVISION`, forget the literal, and the bake succeeds while the expression declares a count the bundle doesn't hold — coverage percentages then use a denominator nobody verified. The lane asks the captured count to double as the drift guard. Same gap in medxpert (`CASE_COUNT = 2450`, no emit-side check) — one shared ticket.

**8 · Three pins the sibling boards have and this one doesn't.**
- No `test_contracteval_definition.py`, though draco-3pass, gdpval, healthbench and medxpert each ship one — and `compute_revision`'s three injectable kwargs (`definition.py:56-58`) exist *only* so such a test can prove each input is in the hash. Nobody passes them today, so hash membership, route addressing and "the expression parses" are all unpinned.
- The board is not in the e2e replay lane's `BOARDS` tuple (`packages/screamingface/tests/e2e/test_boards.py:44-52`), which the manual-onboarding runbook's Step 7 says to wire up precisely so it SKIPs loudly until fixtures exist. medxpert's omission at least carries an explanatory note; this one is silent.
- Not in `test_runtime_cli.py:290`'s parametrize, the only witness for the `name:DATASET_REVISION` fingerprint convention. One word.

**9 · Question I could not measure: does the memoized booklet pin ~100 MB of contract text for the process lifetime?** `runtime.py:94-116` retains the serialized payload forever, where `medxpert/runtime.py` re-reads per call. From this PR's own measurements — p50 5,357 tokens over 4,182 rows at a measured 4.75 chars/token (`pins.py:48-54`, `prepare.py:36-39`) — that is on the order of 100+ MB resident, and `read_selected_cases` lifts each row's `input` again into `SelectedCase`, so a full-split report republishes every contract once per Case. In local mode one process serves many runs. Either record the measured `cases.json` size in the comment at `runtime.py:91`, or memoize the roll call rather than the bytes.

---

## Paper trail

- **Spec D-8 (`:157`) still reads `expected_check_cost="free"`** while the board deliberately ships no `check_surface`. The code's choice is the right one and the AIDEV-NOTE at `definition.py:177-180` gets the reasoning exactly right (declaring one without the handler lets a corrective-loop run pass the pre-spend gate, burn paid turns, then die). The problem is that the ledger says "the five named deviations are unchanged", so a future author reading D-8 as the contract will "fix" the code back toward the trap.
- **The append-only edit records its rationale but not its approval.** `test_benchmark_declaration.py:120-123` explains *why* contracteval is `coverage_declare/single_shot`; the ledger discloses that the row was re-applied under a prior ruling and committed with `--skip-append-only`. One clause in the comment naming the ruling and its date keeps that out of a ledger the next reader of the test file won't open.
- **Ledger `:139` "Push and open the PR. **Not pushed**"** is stale.
- **The work item is still `Backlog` with no labels** while its mirror says `in_review`; the label set (`app/*`, one `who-acts`, one `actor`) is mandatory. The issue has since been renamed to "judge-free clause-extraction", so the mirror's title/`linear_url` and this PR's "the Linear title still says span-extraction" note are both out of date.
- **Ticket "Done when" #2 asks for the row count reconciled against the paper's 4,128**; F-9 offers a hypothesis and §8 puts the reconstruction out of scope. Deliberate and documented — but the ticket was never amended, so it reads as unmet.

---

## Auto-fixable

- `pins.py:3` — "every value here participates in the board's revision hash" is false for `MAX_TOKENS`, `TEMPERATURE` and `MAX_CONTEXT_TOKENS`. The lockfile's own header shouldn't overclaim. (Excluding `MAX_CONTEXT_TOKENS` is behaviourally safe — it is fail-closed at `prepare.py:73-77`, so it can refuse a build but never change served bytes.)
- `pins.py:46` — `TEMPERATURE = "0"` is a `str` while `MAX_TOKENS` is an `int`; copied into `params={}` the strict wire model rejects it.
- `prepare.py:81-82` — `source_id`/`title` are unvalidated `row.get(...)` while `context`/`question` go through `_text`; a renamed upstream column writes `"source_id": null` into every private record and kills traceability quietly. medxpert coerces with `str(row.get("id", ""))`.
- `aggregate.py:113-114` — `attempt.get("metadata")` is dead: `_check` (`runtime.py:134-162`) emits no `metadata` key. Ported verbatim from `medxpert/aggregate.py`, where it is equally dead.
- `prompts.py:4` vs spec F-8 — 18-27 / 18-32, pick one.
- `test_contracteval_aggregate.py:141` — name says "selected positives", the wording D-3 explicitly retracted; `:159` has no WHY docstring; `:221` asserts the failure code but not `grade["score"] is None`; `:45` `_record` lacks a return annotation.
- `test_contracteval_prepare.py:94-103` — only the "raises" half of §7's "context guard both ways" is pinned; a row near the real maximum (~301k chars) would document the headroom of the `_CHARS_PER_TOKEN = 4` heuristic.
- Notebook §4 tells the reader to scale to the full split without the order of magnitude: ~23M input tokens per member pass, times the panel, and this repo has no spend cap.
- Follow-up on both boards, not this PR: `runtime.py:108-114`'s `json.loads` and row projection sit outside the `_unavailable`-converting path (`_read` at `:193` wraps only `OSError`), so a malformed `cases.json` reaches the operator as a bare `JSONDecodeError`/`KeyError` instead of the board-voiced "assets failed preflight". `medxpert/runtime.py` is byte-identical here, and the aggregate side of this same board does handle it.

---

## Verified good

- **Public/private split is airtight.** `cases.json` rows are exactly `{id, input}`, pinned by an exact-set assertion plus explicit `gold_spans`/`answer_start` absence checks. Gold spans are read in exactly two places, both strictly after the candidate has answered (`runtime.py:131`, `aggregate.py:65-74`). The check route's response carries the verdict, polarity, abstention and jaccard — no spans. `_containment_evidence` publishes `gold_span_count` and says why.
- **Route registration is enforced, not merely consistent.** Four routes derive from one revision-carrying prefix; `BenchmarkRegistry.install` renders the protocol and refuses an uninstalled endpoint before the first paid request.
- **The failure-code table is complete.** The spine indexes `self.failure_messages[code]` directly, so a gap is a `KeyError` mid-aggregate. All four codes it can emit here are present; `contracteval_grading_failed` travels as `default_code` and doesn't go through the table. On the SDK side `Failure.code` is an open string and the only code it branches on is `provider_refusal`, so the new codes decode without an SDK change; `CaseStatus` stays closed and this board adds no status.
- **`case_evaluation.py`'s `_require_verdict` is strictly tighter than medxpert's** — real booleans for `correct`/`is_positive`/`abstained`, numeric jaccard in `[0,1]` — and genuinely pins the `bool(None)`-false-negative claim on both sides of the envelope. Together with the polarity-mismatch and preflight tests, those are the strongest witnesses in the PR.
- **Pin discipline passes the blocking check:** `DATASET_REVISION` is a 40-hex commit sha; the mutable `refs/convert/parquet` ref appears only in a WHY comment. Dataset, split, revision, preparer and protocol revisions, the shared evaluation-protocol revision, the result schema and both prompt strings are all in the hash.
- **`build_notebooks.py` is not new machinery** — it landed in #918; this PR adds one entry, pinned by `check_notebooks.py` in CI. No external text is interpolated into generated code.
- **Money:** both notebook runs bounded at `limit=5`, `PARAMS` matches `pins.MAX_TOKENS`, temperature omitted with the provider reasons inline, no float touches a cost anywhere in the diff.

## Coverage gaps

Nothing was executed — no dataset download, no paid call, no test run. The metric degeneracies were verified by recomputing the fixtures standalone; everything else is read-verified against the branch. **Prompt fidelity is unverifiable from inside this repo** (no network, and the promised mirror is absent) — I could confirm the citations disagree and that nothing pins the bytes, not whether the bytes are right. There is no golden and no full-scale run: 8 of 4,182 Cases graded, so every claim about the composed expression's runtime behaviour is structural.

One question for the author: spec F-6 records the load as verified via `hf://…@refs/convert/parquet/default/test/0000.parquet` on `datasets 5.0.1`, but the shipped path is `load_dataset(DATASET, revision=<sha>, split="test")` against the image's pinned `datasets==5.0.0`. Did the rebake that returned 4,182 / 1,244 / 2,938 go through `prepare.load_rows()`? It matters more than usual because `Dockerfile.benchmark` prepares every registered board in one command, so a contracteval load failure fails the benchmark image for all six.

siddhant230:
Did the rebake go through prepare.load_rows()? Yes. screamingface prepare spawns python -m …contracteval.prepare, which calls load_rows() → load_dataset(DATASET, revision=<sha>, split="test"). The hf://…@refs/convert/parquet in F-6 was my pre-implementation probe with pandas — never the shipped path. F-6's wording invites exactly this misreading and should say so.

Does the image's pinned version work? Yes — I tested it. datasets==5.0.0 loads the pinned revision: 4,182 rows, correct columns. The blast-radius risk is retired. But the reviewer was right to ask: the bake I ran used the SDK venv's 5.0.1, so the image's version had never been exercised for this board.

siddhant230:
Thanks — this was the useful kind of review. Round 2's point that the *evidence* was the weak half, not the code, was right, and items 4/5/9 turned out to be the same shape: things that would only fail in production.

All in `67a3715b`. Test count 87 → **113**.

## Done

**#4 · No test of `_check`** — `test_contracteval_check_route.py`, 9 tests. Pins the D-5 case you'd have to read the spec to know mattered: an empty reply gives `correct=False` **and** `abstained=False`, so it can't earn a true negative on a negative row or dilute the laziness rate. Also pins that `_check`'s output satisfies `bind_case_evaluation`'s validator — the field drift that previously failed only after inference was paid for.

**#5 · No resolution test** — `test_contracteval_resolution.py`, 4 tests. Drives cases → candidate → check → case-evaluation → aggregate in-process against a mocked gateway, so it's free. Mutation-verified against both OME-1126 failure modes:

```
$item.input → $item.question  (unbound name → empty prompt)  → caught
array-shaped case-evaluation payload                          → caught (4 tests)
```

Your framing was exactly right — parsing catches neither, because the resolver answers an unknown `$name` with empty text.

**#6 · System-prompt role** — now spec **D-9**, plus a caveat in the board `description` where a leaderboard reader sees it: instructions identical, request shape not, so comparable in protocol but not in message layout. Fixing it properly needs a system-role channel at the candidate boundary, which MedXpertQA needs too → separate ticket.

**#7 · Row count** — `pins.EXPECTED_CASES = 4182`, enforced in `load_rows` (a resized split now fails the bake), and `definition.CASE_COUNT` references it instead of carrying a second literal. Pinned both ways.

**#8 · Three missing pins** — `test_contracteval_definition.py` (11 tests, and the first caller of `compute_revision`'s injectable kwargs, which existed only for this); the e2e `BOARDS` tuple + bundle map, which now SKIPs loudly (7 skips, was 6); the `test_runtime_cli` fingerprint parametrize. The last two are registry-table additions — 7 insertions / 0 deletions in one file, a decorator re-wrap in the other — so they needed `--skip-append-only`, approved by the owner.

**#9 · Memoized booklet** — you estimated 100+ MB. Measured: `cases.json` is **201.5 MB** and the serialized payload is **403 MB resident** (two bytes per character), held for the process lifetime. So I fixed it rather than documenting it: the memo keeps the **preflight verdict**, not the bytes — matching `medxpert/runtime.py` — and the expensive check is still paid once.

That also closed the spec §5 promise you flagged as untested. You were right that hoisting one line would have kept both preflight tests green: they each built a fresh closure and called it once. There's now a test that binds one closure and calls it twice.

## Not done — needs its own ticket

**#3 · Fingerprint ignores `PREPARER_REVISION`.** Agreed it's a real bug, and worse than "review recommended" suggests: edit one word of a prompt, every route re-addresses, the report publishes the new revision, and `screamingface prepare` says *"already prepared"* while serving the old booklet. But `_benchmark_fingerprint` returns `<name>:<DATASET_REVISION>` for **every** board, so the one-line fix invalidates every existing manifest and forces a re-prepare across all seven — a behaviour change for boards this ticket doesn't own. Your "worth a ticket covering both boards" is the right call; happy to file it.

**The bare `JSONDecodeError`** in the cases route — leaving it, since you scoped it out yourself and it's byte-identical in `medxpert/runtime.py`.

## Answering your question

> Did the rebake that returned 4,182 / 1,244 / 2,938 go through `prepare.load_rows()`?

**Yes.** `screamingface prepare` spawns `python -m …contracteval.prepare`, which calls `load_rows()` → `load_dataset(DATASET, revision=<sha>, split="test")`. The `hf://…@refs/convert/parquet` path in F-6 was a pre-implementation probe with pandas and never the shipped path — F-6's wording invited exactly your reading, and it's been corrected.

Worth flagging you were right to ask, because the bake ran on the SDK venv's `datasets 5.0.1` while the image pins **5.0.0**, so the image's version had never been exercised. I tested it: `5.0.0` loads the pinned revision, 4,182 rows, correct columns. Given `Dockerfile.benchmark` prepares every board in one command, that was worth knowing rather than assuming.

## One correction to my own PR body

The July screening note I'd quoted ("span F1/F2", "span-overlap grader") was wrong, as was my `~75k tokens beyond some panel models' context` — tokenized, the max is **63,389** and nothing exceeds 64k. Both the ticket and the spec now say containment and carry the measured numbers.

khoaguin:
Some running results:
`just local-stack-notebooks` then run the `13_contracteval.ipynb` notebook gives
- solo model: 
     - Screenshot 2026-09-21 at 11 16 16
     - [contracteval-report.json](https://github.com/user-attachments/files/32450641/contracteval-report.json)
- fusion: 
     - Screenshot 2026-09-21 at 11 17 32
     - [contracteval-report-fusion.json](https://github.com/user-attachments/files/32450647/contracteval-report-fusion.json)

PR #987, opened 2026-09-18
fix(screamingface): let sf.evaluate take the seed Client.evaluate already takes

khoaguin:
Answer seeds shipped on `Client.evaluate` but never reached `sf.evaluate`, so the one call every example notebook makes raised `TypeError`. Fixes OME-1227.

## Problem / Solution

**Problem.** You could already ask for a repeatable run — but only through a side door. The short, documented one-liner that every example notebook uses rejected the request outright, so the person most likely to want a repeatable run was the one person who could not ask for one.

**Solution.** Let the short call take the same seed the long one already takes, and hand it straight through.

## Before / After

No diagram: this adds no mechanism, it opens an existing one to a caller that could not reach it (the "2 boxes + 1 arrow" case). Announced rather than silently skipped.

### Before
- Trigger: a researcher follows an example notebook and wants the same run twice.
- Today: `sf.evaluate(fusion, benchmark="inspect-gsm8k", limit=2, answer_seed=42)` fails with `TypeError: evaluate() got an unexpected keyword argument 'answer_seed'`.
- Cost: seeding reads as *missing* rather than *misplaced*. Every `examples/*.ipynb` calls `sf.evaluate(...)`, so the whole notebook audience was locked out, and their reports carried `answer_seed: null` with nothing hinting a seed had been available. The Unreleased changelog already advertised the feature, which is what kept the gap invisible.

### After
- Same trigger.
- Now: the keyword is accepted and forwarded; the report carries the seed that produced it.
- Win: a seeded run is reachable from the documented one-liner, so re-running a cell becomes a real reproducibility check instead of a fresh sample.

### Don't regress
- Omitting the keyword still forwards `None`. `apply_answer_seed` treats `None` as a no-op returning the params mapping *itself*, which is what keeps an unseeded run's egress byte-identical and every request-keyed replay fixture valid.
- The complete-URL4 branch forwards the seed too — it takes neither `benchmark` nor `limit`, so it is fed from its own call site and is exactly where a partial fix stops.
- Seeds stay candidate-scope only; nothing here reaches a judge.
- The URL4 branch still refuses `benchmark` and `limit` at runtime.

## Root cause

`Client.evaluate` validates the seed at the door (`_answer_seed_value`, OME-1193 — a bad seed fails before a token is minted) and forwards it to both `evaluate_url4_sync` and `evaluate_sync`, which stamps it on candidate calls only. The module-level convenience wrapper in `_default_client.py` simply never learned about the parameter: absent from both overloads, absent from the implementation, absent from both forwarding calls. A wrapper that is a *narrower door* than the thing it wraps.

## Verification

- Tests written RED first; the two new cases failed with the exact reported `TypeError`, the two regression guards passed from the start (so they are not vacuous).
- `run_gates.py screamingface` **ALL GATES GREEN**: ruff check, ruff format, pyright (0 errors), pytest with `--cov-fail-under=95`, notebook determinism check, `uv build`, distribution check. Full suite: 1590 passed, 25 skipped.
- **Ran with `--skip-append-only`, owner-approved.** The regenerated `public_surface_snapshot.json` counts as a modified prior test artifact. The diff is one line and purely additive — an optional keyword appended to an existing signature, nothing removed or renamed.

## Review order

1. **`packages/screamingface/src/screamingface/_default_client.py`** — the seam, and the whole change. This is the front door notebooks knock on; the Client behind it already accepted the seed. Verify the keyword reaches **both** `client.evaluate(...)` call sites — the complete-URL4 branch above and the Recipes branch below. The URL4 one is the genuinely new runtime behaviour worth your attention, because it is fed separately and takes no `benchmark`/`limit`.
2. **`packages/screamingface/tests/test_default_client_evaluate_passthrough.py`** — the spec for "this wrapper is a pass-through, never a narrower door." Check the third test: it asserts an unseeded call still forwards `None`, which is the invariant protecting byte-identical egress for every existing replay fixture. The two `# type: ignore[call-overload]` are deliberate — the overloads already forbid those arguments to a typed caller, but notebooks are untyped and the runtime guard must stay tested.
3. **`packages/screamingface/tests/public_surface_snapshot.json`** — mechanical skim. Confirm exactly one signature line moved and that the change only *appends* an optional keyword.
4. **`packages/screamingface/CHANGELOG.md`** — a Bug Fixes entry rather than a second feature line, because the existing Unreleased feature entry becomes true rather than being superseded.
5. **`docs/work/` + `docs/tasks/`** — ledger and mirror, closed in the same PR.

## Caveat for users, not a blocker here

`seed` is not portable across providers — Anthropic has none — so a fusion containing an Anthropic member is only partially reproducible. The honest check is two seeded runs diffed per case, not an assumption.


keelancj (approved):
No blocking findings. answer_seed is accepted by both overloads and forwarded through both execution branches, while validation stays in Client.evaluate. Unseeded behavior and URL4 argument restrictions remain intact.

Nonblocking: please reconcile the work ledger with the PR description. The ledger still says the append-only gate is awaiting owner approval; the PR reports an approved, successful run.

Verification: 44 focused tests passed locally, covering forwarding, seed transport/reporting, default-client discovery, and the public API snapshot. GitHub reports green SDK and replay checks. I did not rerun the full suite or make live model calls.
Reviewed commit a43ddd1.

IonesioJunior (approved):
LGTM!

